PR #6, opened 2023-04-16
Add bc to the torizoncore-builder image

### LINKED ISSUE #5: Feature request: Add bc binary to torizoncore-builder
Hi,

The `torizoncore-builder` image gives a really nice dev experience for everything around the TorizonCore image management (including the cross compilation and toolchain handling). Thanks a lot for that!

Recently we stumbled upon a missing binary in the `torizoncore-builder` image.

**Feature Request:**

Add the `bc` tool (https://packages.debian.org/bullseye/bc) to the torizoncore-builder image

**Reason:**

Because we need to include a linux kernel module which is in-tree but not enabled as module in TorizonCore 5.7, we're compiling/including it as an external kernel module with our own Makefile (as described in https://developer.toradex.com/torizon/os-customization/use-cases/building-external-kernel-modules-with-torizon/). Before using any of those in-tree module sources, certain make targets should be called within the linux kernel source directory, like `prepare`. Unfortunately, `prepare` requires the `bc` binary and so the module build fails (when adding `bc` manually to the image, everything goes well).

I can create a PR if it helps (and this request makes sense).

With our setup and `bc` included in `torizoncore-builder` it seems like one can add even an in-tree kernel module easily if it is not enabled by Torizon.

Best regards,

Lukas

lzaugg:
Add the bc tool (https://packages.debian.org/bullseye/bc) to the torizoncore-builder image to support the `prepare` target of the linux kernel source.

### DIFF
--- a/torizoncore-builder.Dockerfile
+++ b/torizoncore-builder.Dockerfile
@@ -60,7 +60,7 @@ FROM common-base AS tcbuilder-base
 RUN apt-get -q -y update && apt-get -q -y --no-install-recommends install \
     python3 python3-pip python3-setuptools python3-wheel python3-gi \
     file curl gzip xz-utils lz4 lzop zstd cpio jq acl libmpc-dev \
-    device-tree-compiler cpp  bzip2 flex bison kmod libgmp3-dev \
+    device-tree-compiler cpp  bzip2 flex bison kmod libgmp3-dev bc \
     && apt-get -q -y --no-install-recommends install python3-paramiko \
     python3-dnspython python3-ifaddr python3-git avahi-daemon && \
     rm -rf /var/lib/apt/lists/*


Nathan-ma:
@lzaugg , Thank you for the PR. I will update you soon.

Nathan-ma:
@lzaugg , I'm sorry to ask, but could you provide me an example of usage? I've read the issue and tried to reproduce it on my end, but I couldn't. I still think your change is valid, but I would appreciate you could send me a little example.

Also, I appreciate you gpg-signed your commit, but if you don't mind, could you also sign it off with the `-s` option? You can quickly amend and force-push this commit.

Thank you, and sorry for the trouble.

lzaugg:
@Nathan-ma, thanks for the feedback.

>  I'm sorry to ask, but could you provide me an example of usage?

No worries. The same understanding helps everyone.

We need to include an in-tree kernel module out-of-tree (namely vub300), because the module is not activated in TorizonCore 5.7 LTS. 

We‘re compiling and including the module by using the torizoncore-builder with `torizoncore-builder kernel build_module ./kernel-modules/vub300` whereas `kernel-modules/vub300` is a directory containing a symlink (to the in-tree module source) and the Makefile:

```bash
> ls -l ./kernel-modules/vub300 
total 16
-rw-r--r--  1 lzaugg  staff  889 Apr 17 17:25 Makefile
lrwxr-xr-x  1 lzaugg  staff   45 Apr 17 17:25 vub300.c -> ../../linux-toradex/drivers/mmc/host/vub300.c
```

But to prepare the linux kernel sources beforehand, a `make script prepare` within the linux kernel source directory (in our case in `./linux-toradex`) is needed to have the include files preprocessed and in it‘s correct place (for the arm64 architecture). 

Eventually the `prepare` target for the linux kernel is using `bc`, which is not available in the torizoncore-builder image and will fail therefore. And that's the reason for raising the issue and this PR. Having the `bc` tool included in the torizoncore-builder image makes my life easier (and potentially others having a similar issue), because there's no need to bring the bc tool somehow into the image at build time or create our own builder image.

Hope that makes sense.
 
> Also, I appreciate you gpg-signed your commit, but if you don't mind, could you also sign it off with the `-s` option?

Yes, will do.

Nathan-ma:
Yes, it does. Thank you very much for the explanation.

You can consider this PR approved after adding the sign-off. I will cherry-pick this commit and add it to our pipeline. 

The PR will be automatically closed once the PR is merged in our Pipeline.

Thank you, Mate!

lzaugg:
@Nathan-ma Ready for cherry-pick then.

Thanks.

Nathan-ma:
I already did! 
The changes are now under an internal review and this PR will be automatically closed once merged internally!

PR #11, opened 2024-02-07
README: Document how to test without building

matthijskooijman:


lucas-akira:
Hi @matthijskooijman ,

It makes sense to document this and I was able to follow the added instructions with no problems.

Can you add the `Signed-off-by: <name and email>` to your commit message with the `git commit -s` option? You can do this by amending then force-pushing the commit.

After that I can approve then continue with the merge process.

matthijskooijman:
> Can you add the Signed-off-by: <name and email> to your commit message with the git commit -s option? You can do this by amending then force-pushing the commit.

Done, also for my other PRs.

lucas-akira (dismissed):

lucas-akira:
Thank you for updating the commits.

Currently this GitHub repo is a mirror from our internal repo, so the merge process is different compared to directly merging here.

I'll merge your PR commits locally, then create an equivalent PR internally where it will be reviewed by other members of the team. Your commits won't be altered and the hashes should be the same as they are here.

Once it's approved and merged there GitHub should detect your commits and mark this PR as merged automatically. If any changes are needed on the commits, I'll reply here.

matthijskooijman:
> I'll merge your PR commits locally, then create an equivalent PR internally where it will be reviewed by other members of the team. Your commits won't be altered and the hashes should be the same as they are here.

Sure. If there is something I can do now, or in future contributions to make things easier, let me know (but I'm assuming this internal repo is not public anywhere).

lucas-akira:
Right, our internal repo is not public. What you can eventually do is sync your other open PRs with the `bullseye` branch once something changes there, preferably via rebasing + force-push.

We'll let you know which PR to update and when to do it, so you don't need to update all of them at the same time.

leograba:
We're also in the process of making development more in the open to make these contributions easier. One of the most impactful examples is https://github.com/torizon/meta-toradex-torizon where development is already being done on GitHub.

Sometime soon this will also happen to TorizonCore Builder and then this merge hassle will be avoided.

I can't guarantee any deadlines, yet this is one of our focuses this quarter.

matthijskooijman:
> We're also in the process of making development more in the open to make these contributions easier.

Good to hear!

lucas-akira:
Hi @matthijskooijman ,

The rest of the team isn't very confident on merging this PR, as adding the instruction comes with the risk that unaware users can try to run the container with an incompatible version of the code, while users that know Docker probably already know how to bind-mount directories to achieve the same results.

Rebuilding the container should be very fast as well if the only thing changed is the code.

Currently our Readme is more geared towards internal development, and as such, it isn't exactly adequate for a full open-source development yet. Because of this I appreciate the need to add more info to potential external contributors, as I can see your intention to do so with this PR.

As said previously, we're in the process to make our development more open, and working on a more suitable Readme is part of this.

I'm inclined to close this PR, but before that let us know if you have other considerations on this matter.

lucas-akira:
Closing PR as there's no activity for the last couple of days. Feel free to reply here if you have any thoughts to share.

matthijskooijman:
I was still planning to respond, but it slipped my mind :-)

> Rebuilding the container should be very fast as well if the only thing changed is the code.

This is true, but the turnaround time of testing a change with the volume mounting is even faster - taking only one step instead of two (rebuilding and running), so I much prefer that approach.

> as adding the instruction comes with the risk that unaware users can try to run the container with an incompatible version of the code,

That is a genuine concern, but that could be easily fixed by adding a warning about this, and/or recommending that users do a clean docker build first and then use the volume trick for subsequent changes.

> while users that know Docker probably already know how to bind-mount directories to achieve the same results.

Sure, but this approach would also be useful for users that do not know docker very well but are interested in developing the builder tool. And users that are familiar with docker can probably derive the right mounts, but that 1) takes some effort and 2) they might not realize it as an option and 3) they might not want to invest time into a method that might or might not work (so pointing out that it does work helps them). In particular, I would like to have this documented for myself and my colleagues. I could put this in our internal docs, but putting it in the README here would make it easier to find and allow others to benefit.



I've pushed one more commit to this PR that adds a note about mismatched image versions to the README, which might alleviate some of your concerns.

While writing the note, I did realize that my README suggestion now documents how to mount the volume using `tcb-env-setup.sh`, but that command normally pulls a new image from remote, replacing the locally built version. There is an `-a local` option to use a local version, but that only considers numerical tags and uses the latest one, which does not work with the `local` tag that is recommended for local builds by the README. I can imagine changing the `tcb-env.sh` script to allow `-a local -t local` to force using the local image tagged `local`, which then allows using the `tcb-env-setup.sh` with proper arguments in all parts of the README (probably keeping the current custom `alias` documented as a secondary option). If that sounds reasonable, I can prepare a PR for that as well.

lucas-akira:
Hi @matthijskooijman ,

I'll bring your points to the rest of the team in our next sync.

lucas-akira:
Hey @matthijskooijman ,

>I can imagine changing the tcb-env.sh script to allow -a local -t local to force using the local image tagged local, which then allows using the tcb-env-setup.sh with proper arguments in all parts of the README (probably keeping the current custom alias documented as a secondary option). If that sounds reasonable, I can prepare a PR for that as well.

At least to me this sounds reasonable, yes. Feel free to prepare the PR and the we'll take a more detailed look at it.

-----

> That is a genuine concern, but that could be easily fixed by adding a warning about this, and/or recommending that users do a clean docker build first and then use the volume trick for subsequent changes.

Even with a warning the team still isn't very confident on adding the section. They feel overall that the procedure is too specific of a use-case to be documented on the README file and endorsed by us.

Of course, we believe the instructions can be valid in some situations as you pointed out, but documenting it here is the main concern on our side.

In any case, we really appreciate your PRs, your merged contributions and your efforts to improve TCB.

PR #12, opened 2024-02-07
Improve failure output

### LINKED ISSUE #10: Errors unpacking image are not shown
As part of the build process, an existing tarball is unpacked. If this fails, because the file is corrupt, does not exist, or is not otherwise available (in this case the file is outside of the project root, so not available in the docker container without additional docker options), an exception is shown but without stdout/stderr output from the tar/cat command that would help diagnose the issue:

```
Traceback (most recent call last):
  File "/builder/torizoncore-builder", line 221, in <module>
    mainargs.func(mainargs)
  File "/builder/tcbuilder/cli/build.py", line 479, in do_build
    build(args.config_fname, args.storage_directory,
  File "/builder/tcbuilder/cli/build.py", line 446, in build
    handle_input_section(config["input"], storage_dir=storage_dir)
  File "/builder/tcbuilder/cli/build.py", line 98, in handle_input_section
    handle_easy_installer_input(props["easy-installer"], **kwargs)
  File "/builder/tcbuilder/cli/build.py", line 119, in handle_easy_installer_input
    images_cli.images_unpack(
  File "/builder/tcbuilder/cli/images.py", line 139, in images_unpack
    images.import_local_image(image_dir, dir_list[0], dir_list[1], dir_list[2])
  File "/builder/tcbuilder/backend/images.py", line 282, in import_local_image
    subprocess.check_output(tarcmd, shell=True, stderr=subprocess.STDOUT)
  File "/usr/lib/python3.9/subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command 'cat /home/matthijs/docs/MKIT/3Devo/FM23/yocto-workdir/torizon/build-torizon/deploy/images/verdin-imx8mp/torizon-core-docker-verdin-imx8mp-Tezi_6.5.0-devel-20231127143231+build.0.tar | cat | tar -xf - -C /storage/tezi.tmp' returned non-zero exit status 2.
```

The relevant python code is here: https://github.com/toradex/torizoncore-builder/blob/5ac80ac541c19d5faf834ac9271fc5553f25e0ad/tcbuilder/backend/images.py#L282

This captures stdout, redirects stderr to stdout but then discards the output. I would suggest either not capturing output at all (just let everything print to the terminal by default), or alternatively use a non-check method so output and exist status can be captured separately and then print the output only if the command failed.

I also wonder why the commandline itself is so complicated with two cat calls and one tar in a pipeline that could, AFAICS, be just a single tar command. The current command is also "vulnerable" to shell injection from the filename (which is not so much a security problem, but will break if the filename to the tarball to unpack contains spaces or special characters). This would also be easy to fix if no shell was required to execute the command (but just a single tar command whose arguments could be specified as a list instead of a single string).

matthijskooijman:
This makes two changes:
 1. Show output of failing commands (such as tar). This allows diagnosing problems such as invalid or missing tarballs or any other command failure. This fixes #10.
 2. Stop printing a useless backtrace on failing module builds.

See the commit message for details.

### DIFF
--- a/tcbuilder/backend/kernel.py
+++ b/tcbuilder/backend/kernel.py
@@ -7,7 +7,6 @@ import os
 import re
 import shutil
 import logging
-import traceback
 import urllib.request
 
 from tcbuilder.backend import ostree
@@ -90,9 +89,8 @@ def build_module(src_dir, linux_src, src_mod_dir,
             CROSS_COMPILE={c_c} ARCH={arch} make -C {src_dir}""",
                        shell=True, stderr=subprocess.STDOUT, check=True)
         print()
-    except:
-        logging.error(traceback.format_exc())
-        raise TorizonCoreBuilderError("Error building kernel module(s)!")
+    except subprocess.CalledProcessError as exc:
+        raise TorizonCoreBuilderError(f"Error building kernel module(s): {exc}") from exc
     finally:
         set_output_ownership(src_dir)
 
--- a/torizoncore-builder.py
+++ b/torizoncore-builder.py
@@ -21,6 +21,7 @@ if sys.version_info < MIN_PYTHON:
 import argparse
 import logging
 import os
+import subprocess
 import traceback
 
 from tcbuilder.cli import (bundle, build, combine, deploy, dt, dto, images, isolate,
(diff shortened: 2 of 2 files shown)


lucas-akira:
I agree that showing failed command output is useful, though I'm not sure on having the traceback only appearing on debug log level, because I'd argue it's also useful to have it shown by default.

With the traceback present it's easy to see where the error occurred in the code. It would be more difficult to find the error source with just the command output, and I think most of the bug reports we receive are with the default log level.

lucas-akira:
Hey @matthijskooijman ,

After a brief internal discussion we concluded that there's no problem in having the traceback available only on debug log level, given that's mostly useful for developers and not for users.

So we have approved and merged the PR, along with some additional commits to our integration tests to take the PR changes into account.

Thank you for your contribution!

matthijskooijman:
Awesome, thanks!

> After a brief internal discussion we concluded that there's no problem in having the traceback available only on debug log level, given that's mostly useful for developers and not for users.

That was also my thought, and it is easy to instruct users to run with debug level if a backtrace is needed.

PR #14, opened 2024-02-07
Make subprocess invocations more robust - do not break on spaces and special characters

matthijskooijman:
In a lot of places, subprocesses were spawned using a shell, passing a single string to be parsed by the shell. This is fragile, since the shell can interpret all kinds of special characters, including spaces. In practice, spaces or shell special characters in a lot of filenames would break the build.

This:
 1. Removes subprocess calls entirely where possible (using `shutil.rmtree` instead).
 2. Stops using the shell in commands that are just a single command and do not need a shell at all.
 3. Uses the `tar --use-decompress-program` to prevent the need for piping when calling tar, removing the need for a shell.
 4. Uses `shlex.quote()` to properly quote arguments in the remaining calls that need a pipeline and thus need a shell (or [more complex Python code](https://docs.python.org/3/library/subprocess.html#replacing-shell-pipeline)).

It might have made sense to add some unit or integration tests for this, but I could not quickly figure out how to run these tests at all. It seems there's only a gitlab CI config that runs them, but that looks quite complex to run manually, so I did not try.

I ran pylint on the code, but there are already a ton of pylint warnings in the `bullseye` branch. I made a diff and tried to fix every newly introduced warning, but I might have missed one ore more.

### DIFF
--- a/tcbuilder/backend/common.py
+++ b/tcbuilder/backend/common.py
@@ -163,6 +163,17 @@ def get_unpack_command(filename):
     return cmd
 
 
+def get_tar_compress_program_options(filename):
+    """ Get array with options to pass to tar to decompress given file format. """
+    cmd = get_unpack_command(filename)
+    # Tar adds a -d option to the command passed, but cat does not
+    # accept this, so omit the --use-compress-program entirely in this
+    # case.
+    if cmd == "cat":
+        return []
+    return ["--use-compress-program", cmd]
+
+
 def get_all_local_ip_addresses():
     """
     Get all local IP addresses on this host except the ones assigned to the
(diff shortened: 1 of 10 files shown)


rborn-tx:
@matthijskooijman These are very nice code improvements. Thank you!

@leograba From my side, these changes are approved.

lucas-akira:
Everything looks good to me as well. @matthijskooijman can you sync your PR to be up-to-date with `bullseye`?

matthijskooijman:
> Everything looks good to me as well. @matthijskooijman can you sync your PR to be up-to-date with `bullseye`?

Done. I haven't tested the code after rebase, but I did previously test all my PRs combined, so I'm confident it should still work.

matthijskooijman:
\o/

lucas-akira:
Code has been merged. Thank you for your contribution!

PR #15, opened 2024-03-20
Make image unpacking more robust

matthijskooijman:
This makes a number of changes:

 1. Allow any supported tarball type, instead of just tar, tar.gz and tgz before.
 2. Actually try to unpack only files, not directories named like a tarball (and copy directory contents only when passed a directory only, not an unsupported file type).
 3. Give better error messages when the passed image does not exist or has an unsupported type.
 4. If the archive contains a single directory, discard the directory and use its contents instead. Previously this would be done only for tarballs with specific names, now it is done for all tarballs and zip files as well.

    This has the side effect of supporting tarballs whose name does not match the contained directory, which is particularly useful to support generically named symlinks to tarballs. This is the primary motivation for this commit.
 5. Use the tempfile module to generate a directory to extract to, which guarantees the directory name is unique and it will be cleaned up afterwards.
 6. After unpacking a tarball or zip, move its contents into the final place, instead of copying and then removing the unpacked files. This should save some time.

### DIFF
--- a/tcbuilder/backend/images.py
+++ b/tcbuilder/backend/images.py
@@ -13,6 +13,7 @@ import shlex
 import shutil
 import subprocess
 import sys
+import tempfile
 import urllib.request
 
 from zipfile import ZipFile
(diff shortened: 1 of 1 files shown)


matthijskooijman:
This PR is a more generic replacement for #13.

I've tested it with a directory (e.g. `images unpack`), with a tarball containing a single directory (a locally generated yocto build) and a tarball without a top-level dir (a torizoncore 5 image from econ that was unpacked well enough to have its image.json processed and then failed for a missing rootfs, but that is after the modified code and already happened before this change). I have *not* tested with a zipfile, since I'm unsure where to find a zip-compressed image to test with.

lucas-akira:
Hey @matthijskooijman ,

Thanks for the request. As we're currently preparing for a new release of TCB in the next couple of days, we may take a bit longer to start reviewing it.

You can see that I recently added some commits into `bullseye` that conflicts with some parts of your PR: we added an initial support for non-TEZI* images, in particular .wic image files, which is the format used for some community-maintained versions of Torizon OS, called [Common Torizon](https://github.com/commontorizon/meta-common-torizon).

(*TEZI image being an OS image that is installed to a device using [Toradex Easy Installer](https://developer.toradex.com/easy-installer/toradex-easy-installer/))


Unpacking a .wic image works a bit differently compared to TEZI images: it doesn't need `tezi_dir`, and TCB unpacks the rootfs contents directly to `src_sysroot_dir` by calling `unpack_local_wic_image()`. You can take a look at `tcbuilder/backend/images.py` in the up-to-date `bullseye` branch for more details.

It shouldn't be very difficult, but you'll need to rebase the PR and adapt the code to also take WIC images into account when unpacking.

Also, don't forget to add the sign-off line at the end of your commit messages.

lucas-akira:
Hi @matthijskooijman ,

As it's been some time, I took the liberty to, on a separate branch, cherry-pick your commit and rebase/adapt it to the current `bullseye` code.

Given that your changes introduce a useful feature on TCB, I'd like to proceed with the merge process with this rebase, with your permission.

As the rebase altered the commit hash, I would not be able to close this GitHub PR as 'merged'. However, the rebased commit will still have you credited as the author.

Let me know your thoughts.

matthijskooijman:
Thanks for picking this up! Totally ok if you merge this manually and then close the PR with a commit. Alternatively, you could just force push to this PR, I think I checked the box to allow that. But whatever works for you :-)

lucas-akira:
Thanks! I've tried pushing to the PR, but it gave me a permission error. Can you double-check if you have the "Allow edits from maintainers" option enabled?

matthijskooijman:
> Thanks! I've tried pushing to the PR, but it gave me a permission error. Can you double-check if you have the "Allow edits from maintainers" option enabled?

Ah, seems this is only possible for PR from forks on personal accounts, but I've created this PR from an organization account, so that won't work: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

lucas-akira:
I see. I'll proceed without updating the commit here then. I'll let you know once the rebased commit gets merged.

lucas-akira:
Hi @matthijskooijman ,

The rebased commit got merged on the default branch here: https://github.com/toradex/torizoncore-builder/commit/65dedbccb95c3aea495473bc73ffcc9d350bfebb

With this I'm closing this PR. Thank you very much for your contribution!

PR #17, opened 2024-03-24
build: Add support for JSON configuration files

microhobby:
Adding support also to JSON configuration files. This way users and other programs can easily read and write the configuration file. JSON is a widely used format and has support in many programming languages and tools sometimes even natively, not the case of YAML.

### DIFF
--- a/tcbuilder/backend/build.py
+++ b/tcbuilder/backend/build.py
@@ -2,6 +2,7 @@
 Backend handling for build subcommand
 """
 
+import json
 import os
 import copy
 import logging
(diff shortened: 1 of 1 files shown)


tcbuilder/backend/build.py
```diff
@@ -242,18 +281,11 @@ def parse_config_file(config_path, schema_path=DEFAULT_SCHEMA_FILE, substs=None)
...
-            error_exc = ParseError(getattr(ex, "problem", "parsing error"))
-            error_exc.set_source(file=config_path)
-            if hasattr(ex, "problem_mark"):
-                mark = getattr(ex, "problem_mark")
-                error_exc.set_source(line=mark.line, column=mark.column)
-            raise error_exc
+    if config_path.endswith(".json"):
+        config = _load_config_file_json(config_path)
+    elif config_path.endswith(".yaml") or config_path.endswith(".yml"):
+        config = _load_config_file_yaml(config_path)
```
matthijskooijman, line 288: This should probably provide a clean error message (and abort) if no supported extension was found. Now I think it will raise an exception for `config` being undefined.

tcbuilder/backend/build.py
```diff
@@ -228,6 +229,44 @@ def make_download_fname(fname):
...
+    with open(config_path) as file:
+        try:
+            config = json.load(file)
+            return config
+
+        except json.JSONDecodeError as ex:
+            error_exc = ParseError(getattr(ex, "msg", "parsing error"))
+            error_exc.set_source(file=config_path)
+            if hasattr(ex, "doc"):
+                error_exc.set_source(line=ex.doc)
```
matthijskooijman, line 267: Why does this use `ex.doc` as the line number? Seems that `JSONDecodeError` has `lineno` (and also `colno`) attributes: https://docs.python.org/3/library/json.html#json.JSONDecodeError

matthijskooijman (commented):
Just a passerby here, but I noticed this PR and it looks interesting. I took a quick look, overall looks good, but I left some small comments inline. This probably also needs some documentation/example and tests, but docs probably live elsewhere and how to run (and write) tests is not really documented AFAICS, so that might be up to Toradex to add.

PR #19, opened 2024-06-11
Dynamic loglevel

neophob:
Hi

### DIFF
--- a/tcbuilder/backend/common.py
+++ b/tcbuilder/backend/common.py
@@ -29,7 +29,7 @@ from tcbuilder.errors import (FileContentMissing, OperationFailureError,
                               LicenceAcceptanceError)
 
 log = logging.getLogger("torizon." + __name__)
-logging.getLogger().setLevel(level=os.getenv('LOGLEVEL', 'WARNING').upper())
+NO_PROGRESS_BAR = os.getenv('NO_PROGRESS', 'false')
 
 DOCKER_BUNDLE_FILENAME = "docker-storage.tar.xz"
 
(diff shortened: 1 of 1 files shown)


lucas-akira:
Hi @neophob ,

Thank you for the PR. Having a way to disable progress bars for automated tests is a valid feature to have on TorizonCore Builder.

I took a quick look at the PR, and apparently the changes introduced in the first commit ([aee37fe](https://github.com/toradex/torizoncore-builder/pull/19/commits/aee37fe72834631fe084b51de5d99fde72092309)) are superseded by the second one ([17390d9](https://github.com/toradex/torizoncore-builder/pull/19/commits/17390d90e145fa7355fc67e50250f2c06a1580ac)). If that's the case, you can rebase the branch and remove the first commit, as it is unnecessary.

And I think it may be better to implement this feature as an argparse argument instead of using an environment variable, as it would be easier to see this option with `--help` instead of only relying on documentation that mentions the env var. 

e.g. an argument like `--no-progress-bars` to be used like `torizoncore-builder --no-progress-bars build`

I'll need to discuss this point with the rest of the team, so no need to change this right now.

Also, can you add the `Signed-off-by: <name and email>` at the end of your commit messages with `git commit -s`?

neophob:
> And I think it may be better to implement this feature as an argparse argument instead of using an environment variable, as it would be easier to see this option with --help instead of only relying on documentation that mentions the env var.

sure, that might work as well (and might be a bit more consistent), my patch was a bit more a quick win.

let me know what you decided - if you agree to go with this PR I will cleanup this and change the commit message

lucas-akira:
Hi @neophob ,

After discussing with the rest of the team we see no problem in disabling the progress bars with an environment variable, so we can continue with the PR.

We plan on eventually implement a switch argument via argparse to do the same as well, as an extra option.

Some suggestions from the team about the environment variable:

- Add a prefix to indicate that the variable is specific to TorizonCore Builder e.g. `TCB_HIDE_PROGRESS`
- Use numeric string values to indicate true/false like "1" and "0" e.g. `TCB_HIDE_PROGRESS=1`

Let me know if you have any questions.

neophob:
updated https://github.com/toradex/torizoncore-builder/pull/21

PR #21, opened 2024-06-24
chore: hide progress via env

neophob:
updated PR, recreated from https://github.com/toradex/torizoncore-builder/pull/19/files

ping @lucas-akira 

---
Hi

### DIFF
--- a/tcbuilder/backend/tcbuild.schema.yaml
+++ b/tcbuilder/backend/tcbuild.schema.yaml
@@ -235,7 +235,7 @@ properties:
   output:
     type: object
     description: "target artifact configuration"
-    properties:
+    definitions:
       ostree:
         type: object
         description: "OSTree-related configuration"
@@ -263,6 +263,8 @@ properties:
     oneOf:
       - required: [raw-image]
         properties:
+          ostree:
+            $ref: '#/properties/output/definitions/ostree'
           raw-image:
             type: object
             description: "configuration items specific to WIC/raw image output"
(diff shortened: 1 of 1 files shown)


lucas-akira:
Hi @neophob ,

Thank you for applying the suggestions.

With the new changes I think the variable check should be:

```
if NO_PROGRESS_BAR:
```

instead of 

```
if NO_PROGRESS_BAR == 'true':
```

Given that `NO_PROGRESS_BAR` now receives a boolean value.

In addition, could you rebase your fork to be up-to-date with `bullseye`? I ask this because right now we merge PRs on an internal GitLab repo (currently this one is a mirror), and so PRs created here will only be considered 'merged' by GitHub if the commit hashes remain the same i.e. if we don't rebase it.

We're in the process of migrating the TorizonCore Builder development publicly to this GitHub repository, so this rebase request from our side shouldn't be necessary in the future.

neophob:
@lucas-akira sure, done!

lucas-akira:
Hi @neophob ,

I see that you incorporated the changes from the last two `bullseye` commits into the one in this PR. However, this would still require me to rebase the PR commit, changing its hash. I could create a merge commit, but I'd prefer [to keep the commit history linear if possible](https://git-scm.com/book/en/v2/Git-Branching-Rebasing).

I need your fork to first have the last two `bullseye` commits in it i.e. up-to-date with `bullseye`, then apply your PR commit (with only your changes) on top of that.

That way the merge process would be just a simple fast-forward merge.

PR #22, opened 2024-06-24
chore: option to hide progress bar

neophob:
@lucas-akira sorry, don't ask how I messed up that PR 🙈

--
Hi

### DIFF
--- a/tcbuilder/backend/common.py
+++ b/tcbuilder/backend/common.py
@@ -31,6 +31,7 @@ from tcbuilder.errors import (FileContentMissing, OperationFailureError,
 log = logging.getLogger("torizon." + __name__)
 
 DOCKER_BUNDLE_FILENAME = "docker-storage.tar.xz"
+NO_PROGRESS_BAR = os.getenv('TCB_HIDE_PROGRESS', '0') == '1'
 
 # Mapping from architecture to a Docker platform.
 ARCH_TO_DOCKER_PLAT = {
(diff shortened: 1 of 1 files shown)


lucas-akira:
Hi @neophob ,

No problem! I'll take the PR for a last review with the team, but it shouldn't take long. Once that's done the PR should be merged here automatically.

I'll let you know here if any updates happen.

rborn-tx:
@neophob Thanks for your contribution. Before we can merge this I'd like to ask for a couple of changes:

- Rather can creating variable `NO_PROGRESS_BAR` and initializing it from the environment variable, create a function in `common.py` say `should_show_progress()` returning a boolean based on the current value of the environment variable `TCB_HIDE_PROGRESS` and on the return value of `sys.stdout.isatty()` (if is not a tty then return False).
- Rather that testing the variable `NO_PROGRESS_BAR` inside `progress()` (i.e. on every call to that function) simply do not call that function; a quick search on the code shows that the function is called in two places:
    - https://github.com/neophob/torizoncore-builder/blob/4ca3ad645d8b63943e53069225c3cb17fb14be26/tcbuilder/backend/build.py#L184 where the code could become:
      ```python
      if should_show_progress():
          print(f"Fetching URL '{url}' into '{in_fname}'")
          ...
      ```
      (before the condition was `if sys.stdout.isatty()`).
    - https://github.com/neophob/torizoncore-builder/blob/4ca3ad645d8b63943e53069225c3cb17fb14be26/tcbuilder/backend/kernel.py#L191 where the code could become:
      ```python
      reporthook=progress if should_show_progress() else None
      urllib.request.urlretrieve(url, filename=tarball, reporthook=reporthook)
      ```

- The above changes only stop the progress bars from showing when downloading files via `urlretrieve()`; other progress information output by the program will continue to show up. I assume you don't care about the other cases which is okay but please make sure the commit message clearly states which cases it's covering (i.e. add a message body with that information).
- Please remove the "chore:" string from the commit message title.

PR #24, opened 2024-08-13
images unpack: Add deprecation note when using TorizonCore 5 images

lucas-akira:
Related-to: TCB-466

### DIFF
--- a/tcbuilder/backend/images.py
+++ b/tcbuilder/backend/images.py
@@ -42,6 +42,11 @@ VERSION_TO_YOCTO_MAP = {
     "kirkstone": "kirkstone-6.x.y"
 }
 
+LAST_DEPRECATED_IMAGE_MAJOR = 5
+LAST_DEPRECATED_IMAGE_NAME = "TorizonCore"
+LAST_DEPRECATED_IMAGE_VERSION = "5.7.2"
+LAST_TCB_VERSION_SUPPORTING_DEPRECATED = "3.10.0"
+
 def serve(images_directory):
     """
     Serve TorizonCore TEZI images via HTTP so they can be installed directly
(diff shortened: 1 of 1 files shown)


tcbuilder/backend/images.py
```diff
@@ -383,6 +387,18 @@ def import_local_image(image_dir_or_file, tezi_dir, src_sysroot_dir, src_ostree_
 
     log.info(f"  Commit checksum: {csum}".format(csum))
     log.info(f"  TorizonCore Version: {metadata['version']}")
+
+    image_major = int(metadata['oe.tdx-major'])
+
+    if image_major <= LAST_DEPRECATED_IMAGE_VERSION:
+        if image_major == LAST_DEPRECATED_IMAGE_VERSION:
+            log.warning("Warning: The last officially supported version of TCB for "
+                        f"{LAST_DEPRECATED_IMAGE_NAME} {LAST_DEPRECATED_IMAGE_VERSION} "
```
jsrc27: My only feedback here is maybe we can make this message more clear that this applies to the entire 5.X series? If we just say:

> The last officially supported version of TCB for TorizonCore 5 images...

I can see some people misunderstanding this to mean version 5.0, instead of all of 5.X.Y versioned images.

lucas-akira: If the message instead said something along the lines of:

> The last officially supported version of TCB for the TorizonCore 5 series (up to and including 5.7.2) is 3.10.0. Newer versions of TCB may work with them, but we don't guarantee support for this use case. Proceed at your own risk.

Would that be clearer?

jsrc27: Yeah that sounds pretty clear to me.

lucas-akira: OK, done!

PR #25, opened 2024-08-15
registryops: Fix --login-to arg behavior when using default registry

lucas-akira:
Until now when using the --login-to arg to login to a private container registry, if the specified registry is equal to the default one (in this case registry-1.docker.io), then the login credentials end up not being applied, resulting in a 'not authorized' error message.

Related-to: TCB-477


jsrc27 (approved):
LGTM

PR #28, opened 2024-08-29
.gitlab-ci.yml: Run 'platform' tests sequentially instead of in parallel

lucas-akira:
It has been observed that when running the 'platform push' tests on both test-torizoncore5 and test-torizoncore6 jobs at roughly the same time, specifically when pushing a generic package to the Cloud servers, one of them would succeed and the other one would fail, seemingly at random.

The cause of this behavior is suspected to be an update conflict when the Cloud metadata repository changes between 'uptane-sign targets pull' and 'uptane-sign targets push', which can occur when multiple clients are trying to push packages to the same Cloud account in a short period of time.

By running the respective Torizon 5 and 6 tests sequentially rather than in parallel, the conflict is avoided.

Related-to: TCB-479

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -56,6 +56,7 @@ services:
 stages:
   - pre-build
   - build
+  - test-platform
   - test
   - build-multiarch
   - deploy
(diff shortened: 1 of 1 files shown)


jsrc27 (approved):
LGTM

rborn-tx:
LGTM as well.

rborn-tx (approved):

PR #30, opened 2024-09-13
bundle: Remove dependency on compose V1 library

lucas-akira:
I did some tests by running 'bundle' with compose files with V2 specific features like `cgroup` and `create_host_path`, and things are working as expected.

Related-to: TCB-458

### DIFF
--- a/tcbuilder/backend/bundle.py
+++ b/tcbuilder/backend/bundle.py
@@ -12,16 +12,14 @@ import time
 
 from datetime import datetime
 
-import compose.config
-import compose.config.environment
-import compose.config.serialize
 import docker
 import docker.errors
 import docker.types
+import yaml
 
 from tcbuilder.errors import (InvalidArgumentError, OperationFailureError,
                               InvalidStorageDriverError)
-from tcbuilder.backend.common import get_own_network
+from tcbuilder.backend.common import get_own_network, validate_compose_file
 from tcbuilder.backend.registryops import RegistryOperations
 
 log = logging.getLogger("torizon." + __name__)
@@ -516,9 +514,15 @@ def download_containers_by_compose_file(
     """
     # Open Docker Compose file
     if not os.path.isabs(compose_file):
-        base_dir = os.path.dirname(os.path.abspath(compose_file))
-    else:
-        base_dir = os.path.dirname(compose_file)
+        compose_path = os.path.abspath(compose_file)
+
+    if not os.path.exists(compose_path):
+        raise InvalidArgumentError(f"Error: File does not exist: {compose_file}. Aborting.")
+
+    if not os.path.isfile(compose_path):
+        raise InvalidArgumentError(f"Error: Not a file: {compose_file}. Aborting.")
+
+    log.info("NOTE: TCB no longer expands environment variables present in the compose file.")
 
     if show_progress:
         _term = os.environ.get('TERM')
@@ -527,16 +531,11 @@ def download_containers_by_compose_file(
         elif not (_term.startswith('xterm') or _term.startswith('rxvt')):
             show_progress = False
 
-    try:
-        environ = compose.config.environment.Environment.from_env_file(base_dir)
-        details = compose.config.find(
-            base_dir, [os.path.basename(compose_file)], environ, None)
-        config = compose.config.load(details)
+    with open(compose_path, encoding='utf-8') as file:
+        compose_file_data = yaml.load(file, Loader=yaml.FullLoader)
 
-    except compose.config.errors.ConfigurationError as exc:
-        raise OperationFailureError(
-            "Error: "
-            f"Could not load the Docker compose file '{compose_file}'.") from exc
+    # Basic compose file validation e.g. if it has 'services' section, images are specified, etc.
+    validate_compose_file(compose_file_data)
 
     if use_host_docker:
         log.debug("Using DockerManager")
@@ -561,9 +560,9 @@ def download_containers_by_compose_file(
             login_to_registries(dind_client, logins)
 
         # Now we can fetch the containers...
-        for service in config.services:
-            image_name = service['image']
-            log.info(f"Fetching container image {image_name}")
+        for svc_name, svc_spec in compose_file_data['services'].items():
+            image_name = svc_spec.get('image')
+            log.info(f"Fetching container image {image_name} in service {svc_name}")
             if not ":" in image_name:
                 image_name += ":latest"
 
@@ -577,14 +576,11 @@ def download_containers_by_compose_file(
                 # Use high-level API (no progress info).
                 image = dind_client.images.pull(image_name, platform=platform)
 
-            service['image'] = image.attrs['RepoDigests'][0]
+            svc_spec['image'] = image.attrs['RepoDigests'][0]
 
         log.info("Saving Docker Compose file")
         with open(os.path.join(manager.output_dir, "docker-compose.yml"), "w") as file:
-            # Serialization need to escape dollar sign and requires no
-            # environment variables interpolation.
-            file.write(compose.config.serialize.serialize_config(
-                config, escape_dollar=False).replace("@@MACHINE@@", "$MACHINE", 1))
+            file.write(yaml.dump(compose_file_data, Dumper=yaml.Dumper))
 
         log.info("Exporting storage")
         manager.save_tar(output_filename)
--- a/tcbuilder/backend/bundle.py
+++ b/tcbuilder/backend/bundle.py
@@ -485,11 +485,39 @@ def login_to_registries(client, logins):
         client.login(username, password, registry=registry)
 
 
+def check_double_dollar_sign(string, config_path):
+    if "$$" in string:
+        log.info(f"Replacing '$$' with '$' in compose config value found in: {config_path}.")
+        return string.replace("$$", "$")
+    return string
+
+
+# Implementation based on 'recursive_interpolate()' from Compose V1
+# https://github.com/docker/compose/blob/v1/compose/config/interpolation.py#L71
+def recursive_yaml_value_check(obj, config_path):
+    log.debug(f"recursive_yaml_value_check: Call with obj {type(obj)}. "
+              f"Checking config values in '{config_path}'")
+
+    if isinstance(obj, str):
+        obj = check_double_dollar_sign(obj, config_path)
+
+    if isinstance(obj, dict):
+        return {
+            key: recursive_yaml_value_check(value, f"{config_path}/{key}")
+            for key, value in obj.items()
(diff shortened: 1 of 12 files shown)


jsrc27:
Just to understand the implications here.

So now, the only kind of "validation" done on the compose file is whether it's proper yaml or not. We no longer check if the fields and options in the file follow the compose spec or not, correct?

TCB no longer expands environment variables. So that means users don't have to work around this limitation anymore?: https://developer.toradex.com/torizon/os-customization/use-cases/pre-provisioning-docker-containers-onto-a-torizoncore-image/#environment-variables

As for users who actually were using environment variables to be expanded during `bundle`, this is kind of a big change no?

jsrc27:
With this change, do we use the compose python library anywhere now? If not, could we safely remove it from the container image now?

lucas-akira:
> So now, the only kind of "validation" done on the compose file is whether it's proper yaml or not. We no longer check if the fields and options in the file follow the compose spec or not, correct?

That's basically it, although TCB will also check if the YAML file has a `services` section listing each service, each one with an `image` string field in it to properly get all container image names to download.

Apart from the above no other compose specification will be verified.

> TCB no longer expands environment variables. So that means users don't have to work around this limitation anymore?: https://developer.toradex.com/torizon/os-customization/use-cases/pre-provisioning-docker-containers-onto-a-torizoncore-image/#environment-variables

As things are right now, that's correct. We would need to remove/update this section of the developer site.

> As for users who actually were using environment variables to be expanded during bundle, this is kind of a big change no?

Right, this change will affect people using the workaround linked above. I'm not really sure if TCB should be expanding environment variables in a compose file in the first place, so I ended up not including any logic to process these.

The fact that it was expanding env vars before looks like more of an unintended side effect of using the compose library to parse through the compose YAML file rather than a deliberate decision.

Is there a valid use-case for TCB to interpret env vars during a bundle command?

I'm open to implement the logic to process env vars (with one or two preceding dollar signs) if the impact of not doing so is large enough.

> With this change, do we use the compose python library anywhere now? If not, could we safely remove it from the container image now?

Good catch, I forgot to remove the compose library from `requirements_debian.txt`. I'll amend the last commit.

rborn-tx:
> The fact that it was expanding env vars before looks like more of an unintended side effect of using the compose library to parse through the compose YAML file rather than a deliberate decision.

I agree. This probably explains why we documented a workaround (which was probably also a bad idea).

Now, the situation due to the workaround is that any users utilizing the env var expansion feature would have replaced `$` => `$$` in their compose files. So, if we don't want to introduce a breaking change from the perspective of those users, we'd need to do the replacement `$$` => `$`. In this way, things would work for both users applying or not applying the workaround. AFAICT that would be the only replacement we would need to do ourselves since any environment variable expansion should only be happening on the device.

Apart from the above, I agree with this implementation and I don't think TCB should itself be trying to validate the compose file since this would be a maintenance burden. For example, in theory the schema would need to be updated in TCB whenever we updated the `docker-compose` tool in Torizon OS since the schema used in the validation is always the latest one known by `docker-compose` independent of the version field value (https://docs.docker.com/reference/compose-file/version-and-name/#version-top-level-element-obsolete).

tcbuilder/backend/bundle.py
```diff
@@ -577,14 +576,11 @@ def download_containers_by_compose_file(
...
 
-            service['image'] = image.attrs['RepoDigests'][0]
+            svc_spec['image'] = image.attrs['RepoDigests'][0]
 
         log.info("Saving Docker Compose file")
         with open(os.path.join(manager.output_dir, "docker-compose.yml"), "w") as file:
-            # Serialization need to escape dollar sign and requires no
-            # environment variables interpolation.
-            file.write(compose.config.serialize.serialize_config(
-                config, escape_dollar=False).replace("@@MACHINE@@", "$MACHINE", 1))
```
rborn-tx, line 587: I imagine this was done as a hack to deal with the variables expansion before the "workaround" of using a double `$` was known. So once we merge this changes I imagine some internal pipeline will break.

This seems to be used in the bullseye branch of project `torizon-core-docker-evaluation` at least.

rborn-tx: Checking the commit history:

```
bundle: replace @@MACHINE@@ with $MACHINE when parsing docker-compose file

Replace @@MACHINE@@ with $MACHINE with $MACHINE when bundle command is parsing docker-compose file.

Related-to: TOR-1458
Signed-off-by: Dejan Jesenko <dejan.jesenko@toradex.com>
```

So it's probably worth checking ticket TOR-1458.

lucas-akira: Looking at TOR-1458 (now TCB-203), this workaround was made in order to avoid `${MACHINE}` from being expanded in the `command` field of the `portainer` service of the docker compose YAML used in the Portainer version of `torizon-core-docker-evaluation`.

lucas-akira:
> So, if we don't want to introduce a breaking change from the perspective of those users, we'd need to do the replacement $$ => $. In this way, things would work for both users applying or not applying the workaround. AFAICT that would be the only replacement we would need to do ourselves since any environment variable expansion should only be happening on the device.

I agree with only doing this replacement instead of implementing environment variable expansion. I would also add a log message to the user whenever TCB replaces `$$` by `$`.

jsrc27:
> Is there a valid use-case for TCB to interpret env vars during a bundle command?

Honestly I don't know for sure. It's not something we explicitly advertise as a "feature", I don't think. But I can't really guarantee that no one was using this "feature". Personally I agree and don't think it makes much sense either, but it is a difference.

lucas-akira:
I've removed the `compose` V1 library from the container image and replaced it with `jsonschema`, a compose dependency that is directly used in the `build` command.

I've also added the double dollar sign substitution during the `bundle` command. Input compose YAML is not modified.

Lastly I added an additional commit which replaces all `yaml.load()` and `yaml.dump()` calls with their respective safe counterparts. It passed the Gitlab test pipeline and some manual tests, so I don't think this change will break things.

jsrc27 (approved):
Latest revision looks okay to me.

rborn-tx:
Perhaps it's worth adding a testcase for the double dollar replacement?

tcbuilder/backend/bundle.py
```diff
@@ -487,6 +485,13 @@ def login_to_registries(client, logins):
         client.login(username, password, registry=registry)
 
 
+def check_double_dollar_signs(string):
+    if "$$" in string:
```
rborn-tx, line 489: Checking the documentation on [Interpolation](https://docs.docker.com/reference/compose-file/interpolation/) I notice that the feature works in a way that differs from what this stretch of code assumes (that the substitutions are performed before parsing the file as yaml (i.e. in the raw text)). In fact, the docs says "Interpolation applies only to YAML values, not to keys".

Another complication I found is what is said [here](https://docs.docker.com/compose/environment-variables/variable-interpolation/#interpolation-syntax): "Interpolation is applied for unquoted and double-quoted values.", i.e. apparently not to single quoted strings.

In light of the above, the question now becomes: do we want to handle the extra complexity to ensure backwards compatibility? If the answer is yes, then we'd need to parse the yaml and then perform the replacement `$$` => `$` only on unquoted and double quoted string **values**. Opinions?

lucas-akira: Parsing the YAML contents to replace `$$` with `$` in string _values_  should be doable without much effort.

However, the main challenge that I see with this is to only select unquoted and double quoted strings, especially after parsing the YAML file, as all of those would be converted to Python strings, with no easy way to tell whether they were originally unquoted, single or double quoted in the YAML file.

While this unquoted/double quoted requirement for interpolation is specified in the Compose V2 documentation, I couldn't find it on the legacy Compose V1 documentation: https://github.com/docker/compose/blob/v1/docs/Compose%20file%20reference%20(legacy)/version-3.md#variable-substitution

rborn-tx: That's a good point.

Well, in the end what matters for us is if the methods we used before did the substitution or not. Performing a quick test it seems it does the substitution on all value strings (but not the keys):

E.g. entering this to `tcb bundle`:

```
services:
  hello-world:
    image: hello-world:latest
    environment:
      $var: $HOME
      HOME1: $HOME
      HOME2: "$HOME"
      HOME3: '$HOME'
      HOME4:
        >-
          $HOME
      HOME5: |
          $HOME
      HOME6: |
          \$HOME
```

gives this output:

```
services:
  hello-world:
    environment:
      $var: /root
      HOME1: /root
      HOME2: /root
      HOME3: /root
      HOME4: /root
      HOME5: '/root

        '
      HOME6: '\/root

        '
    image: hello-world@sha256:91fb4b041da273d5a3273b6d587d62d518300a6ad268b28628f74997b93171b2
version: '3.9'
```

So, it should be enough to make the substitutions on all string values...

lucas-akira: That's great to know, thanks for the tests!

Aside from this, the Compose V2 docs also mention the fact that a user may want to keep the double-dollar sign if they don't want environment variables being processed by Compose:
```
web:
  build: .
  command: "$$VAR_NOT_INTERPOLATED_BY_COMPOSE"
```

To cover this use case I can also add a switch to the bundle command (and its respective option on tcbuild.yaml) to disable the replacement, something like `--keep-double-dollar-sign`.

rborn-tx: Once we get rid of the workaround documentation from our docs, then any users using the `$$` in their compose would likely be following the Compose documentation and them their expectation would be that TCB doesn't itself touch the file.

So, yes I think it makes sense.

tcbuilder/backend/tcbuild.schema.yaml
```diff
@@ -365,6 +365,9 @@ properties:
                         type: string
                         description: "CA certificate path of alternative container registry"
                         tdxMeta: "file;cer,crt"
+                      keep-double-dollar-sign:
+                        type: boolean
+                        description: "Keep double dollar sign intead of replacing it with a single one when parsing string values of the compose file"
```
rborn-tx: typo: "intead" -> "instead"

lucas-akira: Fixed. Thanks for the correction!

rborn-tx (approved):
Latest version looks good to me!

lucas-akira:
Great! I'll add the testcases for the double dollar sign replacement. After that I think we're done.

rborn-tx:
> Great! I'll add the testcases for the double dollar sign replacement. After that I think we're done.

Looking forward to it.

rborn-tx (approved):
New test looks good to me as well.

PR #31, opened 2024-10-03
build: Add support to fetch Torizon OS 7 images via toradex-feed option

lucas-akira:

Related-to: TCB-484

### DIFF
--- a/tcbuilder/backend/build.py
+++ b/tcbuilder/backend/build.py
@@ -44,7 +44,8 @@ MAJOR_TO_YOCTO_MAP = {
     6: "kirkstone-6.x.y",
     7: "scarthgap-7.x.y"
 }
-DEFAULT_IMAGE_VARIANT = "torizon-core-docker"
+DEFAULT_LEGACY_IMAGE_VARIANT = "torizon-core-docker"
+DEFAULT_IMAGE_VARIANT = "torizon-docker"
 
 # Assigment regex pre-compiled.
 ASSGN_REGEX = re.compile(r"^([a-zA-Z_][a-zA-Z_0-9]*)=(.*)$")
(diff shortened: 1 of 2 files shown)


lucas-akira:
I did some quick tests and things appear to be working fine without breaking things with 6.X and 5.X images.

tcbuilder/backend/build.py
```diff
@@ -317,6 +318,12 @@ def make_feed_url(feed_props):
         raise InvalidDataError(
             f"Don't know how to handle a major version of {version_major}")
 
+    if params["variant"] is None:
+        if MAJOR_TO_YOCTO_MAP[version_major] in ("dunfell-5.x.y", "kirkstone-6.x.y"):
+            params["variant"] = DEFAULT_LEGACY_IMAGE_VARIANT
+        else:
+            params["variant"] = DEFAULT_IMAGE_VARIANT
```
jsrc27, line 325: Instead of an `else` here should we explicitly check if the `version_major` is even scarthgap based or not?

rborn-tx: The `if version_major` above should prevent any other major from being entered in which case I think the condition here is okay considering that we probably don't want to change the name of the variant any time soon. When another major of TOS is available we'd add it to `MAJOR_TO_YOCTO_MAP` but the code here wouldn't need to be changed.

rborn-tx (approved):
LGTM

tcbuilder/backend/build.py
lucas-akira: > The if version_major above should prevent any other major from being entered in which case I think the condition here is okay considering that we probably don't want to change the name of the variant any time soon. When another major of TOS is available we'd add it to MAJOR_TO_YOCTO_MAP but the code here wouldn't need to be changed.

Exactly, that's why I added the variant check after the `if version_major` check. And in `MAJOR_TO_YOCTO_MAP` 5 is the lowest version defined, so the `else` statement will only be reached with major version 7 (and above when they're eventually defined in `MAJOR_TO_YOCTO_MAP`).

jsrc27 (approved):

PR #32, opened 2024-10-16
Add debugpy support & implement bundle property to raw-image yaml & schema fixes & vs code pylint support

microhobby:
This series:

- Adds fixes to the schema tdxMeta to integrate with IDE v2
- Add support for debugpy if the developer set TCB_DEBUGPY env var
- Implement the bundle property into raw-image property

Open to suggestions from maintainers.

### DIFF
--- a/tcbuilder/backend/bundle.py
+++ b/tcbuilder/backend/bundle.py
@@ -33,8 +33,10 @@ def get_compression_command(output_file):
 
     Returns:
         (str, str): output_file without compression ending, compression command
+        (str, None): if output_file doesn't end with any of the compression formats below
     """
     command = None
+    output_file_tar = output_file
     if output_file.endswith(".xz"):
         output_file_tar = output_file[:-3]
         command = ["xz", "-3", "-z", output_file_tar]
@@ -50,6 +52,8 @@ def get_compression_command(output_file):
     elif output_file.endswith(".zst"):
         output_file_tar = output_file[:-4]
         command = ["zstd", "--rm", output_file_tar, "-o", output_file]
+    elif not output_file.endswith(".tar"):
+        output_file_tar = f"{output_file}.tar"
 
     return (output_file_tar, command)
 
@@ -87,7 +91,7 @@ class DockerManager:
         """Create an instance of the Docker client"""
         return docker.from_env()
 
-    def save_tar(self, output_file, compress=True):
+    def save_tar(self, output_file):
         """Create compressed tar archive of the Docker images"""
 
         output_file_tar, compression_command = get_compression_command(output_file)
@@ -101,7 +105,7 @@ class DockerManager:
         if os.path.exists(output_filepath):
             os.remove(output_filepath)
 
-        if compress:
+        if compression_command is not None:
             subprocess.run(compression_command, cwd=self.output_dir, check=True)
         else:
             log.debug(f"Not compressing {output_file_tar}")
@@ -322,7 +326,7 @@ class DindManager(DockerManager):
         dind_client = docker.DockerClient(base_url=self.docker_host, tls=tls_config)
         return dind_client
 
-    def save_tar(self, output_file, compress=True):
+    def save_tar(self, output_file):
         """Create compressed tar archive of the Docker images"""
 
         log.info(f"Storing container bundle into \"{self.bundle_dir}\"")
@@ -379,7 +383,7 @@ class DindManager(DockerManager):
             raise OperationFailureError(
                 f"Could not create output tarball in '{output_file_tar}'.")
 
-        if compress:
+        if compression_command is not None:
             log.debug(f"compression_command: {compression_command}")
 
             output_filepath = os.path.join(self.bundle_dir, output_file)
@@ -525,7 +529,7 @@ def recursive_yaml_value_check(obj, config_path):
 def download_containers_by_compose_file(
         output_dir, compose_file, host_workdir, output_filename,
         keep_double_dollar_sign=False, platform=None, dind_params=None,
-        use_host_docker=False, show_progress=True, compress=True):
+        use_host_docker=False, show_progress=True):
     """
     Creates a container bundle using Docker (either Host Docker or Docker in Docker)
 
@@ -625,7 +629,7 @@ def download_containers_by_compose_file(
             file.write(yaml.safe_dump(compose_file_data))
 
         log.info("Exporting storage")
-        manager.save_tar(output_filename, compress)
+        manager.save_tar(output_filename)
 
     except docker.errors.APIError as exc:
         raise OperationFailureError(
--- a/tcbuilder/backend/bundle.py
+++ b/tcbuilder/backend/bundle.py
@@ -87,7 +87,7 @@ class DockerManager:
         """Create an instance of the Docker client"""
         return docker.from_env()
 
-    def save_tar(self, output_file):
+    def save_tar(self, output_file, compress=True):
         """Create compressed tar archive of the Docker images"""
 
         output_file_tar, compression_command = get_compression_command(output_file)
@@ -100,7 +100,11 @@ class DockerManager:
         output_filepath = os.path.join(self.output_dir, output_file)
         if os.path.exists(output_filepath):
             os.remove(output_filepath)
-        subprocess.run(compression_command, cwd=self.output_dir, check=True)
+
+        if compress:
+            subprocess.run(compression_command, cwd=self.output_dir, check=True)
+        else:
+            log.debug(f"Not compressing {output_file_tar}")
 
     def add_cacerts(self, cacerts):
         assert cacerts is None, "`cacerts` should be used with DindManager"
@@ -318,7 +322,7 @@ class DindManager(DockerManager):
         dind_client = docker.DockerClient(base_url=self.docker_host, tls=tls_config)
         return dind_client
 
-    def save_tar(self, output_file):
+    def save_tar(self, output_file, compress=True):
         """Create compressed tar archive of the Docker images"""
 
         log.info(f"Storing container bundle into \"{self.bundle_dir}\"")
@@ -375,12 +379,16 @@ class DindManager(DockerManager):
             raise OperationFailureError(
                 f"Could not create output tarball in '{output_file_tar}'.")
 
-        log.debug(f"compression_command: {compression_command}")
+        if compress:
+            log.debug(f"compression_command: {compression_command}")
 
-        output_filepath = os.path.join(self.bundle_dir, output_file)
-        if os.path.exists(output_filepath):
-            os.remove(output_filepath)
-        subprocess.run(compression_command, cwd=self.bundle_dir, check=True)
+            output_filepath = os.path.join(self.bundle_dir, output_file)
+            if os.path.exists(output_filepath):
+                os.remove(output_filepath)
+            subprocess.run(compression_command, cwd=self.bundle_dir, check=True)
+        else:
+            log.debug(f"Not compressing {output_file_tar}")
+
 
     def add_cacerts(self, cacerts):
         """Add the required certificate files for a secure registry
@@ -517,7 +525,7 @@ def recursive_yaml_value_check(obj, config_path):
 def download_containers_by_compose_file(
         output_dir, compose_file, host_workdir, output_filename,
         keep_double_dollar_sign=False, platform=None, dind_params=None,
-        use_host_docker=False, show_progress=True):
+        use_host_docker=False, show_progress=True, compress=True):
     """
     Creates a container bundle using Docker (either Host Docker or Docker in Docker)
 
@@ -617,7 +625,7 @@ def download_containers_by_compose_file(
             file.write(yaml.safe_dump(compose_file_data))
 
         log.info("Exporting storage")
-        manager.save_tar(output_filename)
+        manager.save_tar(output_filename, compress)
 
     except docker.errors.APIError as exc:
         raise OperationFailureError(
--- a/tcbuilder/backend/combine.py
+++ b/tcbuilder/backend/combine.py
@@ -4,12 +4,13 @@ import logging
 import re
 import datetime
 
+import fnmatch
 import guestfs
 
 from tezi.image import ImageConfig
 from tcbuilder.backend.common import \
     (set_output_ownership, check_licence_acceptance,
-     run_with_loading_animation, DOCKER_BUNDLE_FILENAME, DOCKER_BUNDLE_FILENAME_UNCOMPRESSED)
+     run_with_loading_animation, DOCKER_BUNDLE_TARNAME)
 from tcbuilder.errors import InvalidStateError, InvalidDataError, TorizonCoreBuilderError
 
 log = logging.getLogger("torizon." + __name__)
@@ -17,15 +18,9 @@ log = logging.getLogger("torizon." + __name__)
 TARGET_NAME_FILENAME = "target_name"
 DOCKER_FILES_TO_ADD = [
     "docker-compose.yml:/ostree/deploy/torizon/var/sota/storage/docker-compose/",
-    DOCKER_BUNDLE_FILENAME + ":/ostree/deploy/torizon/var/lib/docker/:true",
     TARGET_NAME_FILENAME + ":/ostree/deploy/torizon/var/sota/storage/docker-compose/"
 ]
-
-DOCKER_FILES_TO_ADD_TO_RAW = [
-    "docker-compose.yml:/ostree/deploy/torizon/var/sota/storage/docker-compose/",
-    DOCKER_BUNDLE_FILENAME_UNCOMPRESSED + ":/ostree/deploy/torizon/var/lib/docker/:true",
-    TARGET_NAME_FILENAME + ":/ostree/deploy/torizon/var/sota/storage/docker-compose/"
-]
+DOCKER_BUNDLE_DESTINATION = "/ostree/deploy/torizon/var/lib/docker/:true"
 
 TEZI_PROPS = [
     "name",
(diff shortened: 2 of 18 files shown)


microhobby:
Probably I need to add some integration with Pylint also to vs code 😅
/me remember

tcbuilder/backend/deploy.py
```diff
@@ -354,6 +354,60 @@ def write_rootfs_to_raw_image(base_raw_img, output_raw_img, base_rootfs_partitio
                 args=(f"{dst_sysroot_dir}/{content}", "/"),
                 loading_msg=f"  Copying /{content}...")
 
+        # handle the combine of docker-storage bundle
+        bundle_dir = "bundle.tmp"
+        if os.path.exists(bundle_dir):
```
jsrc27: Here we're just relying on the existence of a directory with a pre-determined name. Seems kind of fragile to me. Especially when we try to add support for raw images in the standalone `bundle` command, we may have to re-think or revise this.

Looking at what was done for the Tezi equivalent here. Internally to the build command we do `bundle` -> `combine`. But, looking at the `combine` command, it would need to be revised to add support for raw images if we want to have a similar workflow.

@lucas-akira Did you have any ideas/plans for raw-images and the `bundle`/`combine` workflow when you were adding initial support for raw-images?

jsrc27:
Seems like pylint is complaining. Could you fix what it's complaining about or disable the check for the section pylint is complaining about if needed.

microhobby:
> Seems like pylint is complaining. Could you fix what it's complaining about or disable the check for the section pylint is complaining about if needed.

Yeah, I will handle this.

tcbuilder/backend/deploy.py
microhobby: Well, the idea is to create the path anyway if it does not exist, and if it exists the code cleans it and recreates it. It follows the same way, or kind of, that the combine works. But instead of removing the folder at the end, it leaves it there. If you guys though that is better remove it I can do the same.

lucas-akira: > Looking at what was done for the Tezi equivalent here. Internally to the build command we do bundle -> combine. But, looking at the combine command, it would need to be revised to add support for raw images if we want to have a similar workflow.
>
> Did you have any ideas/plans for raw-images and the bundle/combine workflow when you were adding initial support for raw-images?

Actually the `combine` command already supports raw images. For an eventual `build` command support, the workflow that I was planning would be to internally call `bundle` followed by `combine`, just like it is right now with TEZI images.

That way we would avoid having part of the `bundle` command code doing essentially the same thing that `combine` already does.

jsrc27: > Actually the combine command already supports raw images.

Oh okay so in that case this can just use the same workflow  as Tezi then?

microhobby: @jsrc27 @lucas-akira I updated the MR for the bundle bd12fd197118f0dde1c77a068e03a3083f1930d4 I'm following the same flow from the CLI command now, but with one exception. The compression of the .tar file does not make sense. For TEZI it makes sense, but not for raw images. So, I added some flags to tell the flow not to compress and decompress the file, which decreased the build processing time.

Take a look, thanks!

lucas-akira: > The compression of the .tar file does not make sense. For TEZI it makes sense, but not for raw images. So, I added some flags to tell the flow not to compress and decompress the file, which decreased the build processing time.

Right, given that during a `build` command the data of the bundled container images are tarred then soon after untarred into the raw image rootfs, I agree that it doesn't make sense to add the compression step to `docker-storage.tar` in this situation.

That being said, I think that the compression step should still be added by default when running the standalone `bundle` command, whether the container images are made for TEZI or raw images. I say this because in this case the resulting bundle directory will be saved on disk, and it could be stored to be reused for more than one OS image. And so it would be good if the contents are compressed to save disk space.

If doing the above then the `combine` command should be modified to be able to auto-detect if the Docker bundle tar file is compressed or not, and act accordingly.

tcbuilder/cli/build.py
```diff
@@ -438,24 +444,85 @@ def handle_bundle_output(image_dir, storage_dir, bundle_props, tezi_props):
                 "use_host_docker": False,
                 "output_filename": common.DOCKER_BUNDLE_FILENAME,
```
lucas-akira: With this line unaltered the `output_filename` parameter of `download_containers_by_compose_file()` will be `common.DOCKER_BUNDLE_FILENAME` (= `docker-storage.tar.xz`) even when using a raw image, since this code line is now part of the new `handle_bundle_common()` function.

tcbuilder/backend/bundle.py
```diff
@@ -100,7 +100,11 @@ def save_tar(self, output_file):
         output_filepath = os.path.join(self.output_dir, output_file)
         if os.path.exists(output_filepath):
             os.remove(output_filepath)
-        subprocess.run(compression_command, cwd=self.output_dir, check=True)
+
+        if compress:
+            subprocess.run(compression_command, cwd=self.output_dir, check=True)
```
lucas-akira: If the `output_file` argument from `save_tar()` doesn't end with one of the file extensions specified in `get_compression_command()`, then `compression_command` will be `None`, breaking `subprocess.run()`.

This will happen if `output_file` ends with `.tar`.

jsrc27:
Gave the latest revision another look over. Looks a lot better now that the raw image flow is more in-line with the tezi image flow.

tcbuilder/cli/build.py
```diff
@@ -384,21 +394,17 @@ def handle_easy_installer_output(props, storage_dir, union_params):
...
-        # Do a combine "in place" to avoid creating another directory.
-        combine_params = {
-            "image_dir": image_dir,
-            "bundle_dir": bundle_props["dir"],
-            "output_directory": None,
-            "tezi_props": translate_tezi_props(tezi_props),
-            "force": True
-        }
-        comb_be.combine_tezi_image(**combine_params)
+        return bundle_props["dir"]
```
jsrc27: Since this common function handles the case where the user just provides a pre-made bundle directory, shouldn't the schema for raw-images also support this then? With the current changes the schema would only support a provided compose file and not bundle directory for raw images.

microhobby: yeah, this is exactly the idea, for now, only to support the `bundle` `compose-file`. This could be a `TODO` or if you guys think that should be added now, let me know.

jsrc27: Okay well, I don't have a super strong opinion on whether it should be done now or later. I just wanted to check if it was an intended decision on your end. 

Out of curiosity is there still some extra work or something needed still to add support for the bundle directory for raw images in the `build` command?

lucas-akira: > Out of curiosity is there still some extra work or something needed still to add support for the bundle directory for raw images in the build command?

I don't think there is any extra work other than to add the fields in `tcbuild.schema.yaml` and add the relevant calls in `cli/build.py`. I can take care of this.

lucas-akira:
As internally discussed with @microhobby , I'll take care of the PR from now on. I plan to keep the current commits and just add the necessary fixes/suggestions.

lucas-akira (commented):
Changes look good overall, with only a few points to be aware of.

lucas-akira:
~~Apparently I cannot commit changes to this Pull Request branch, so I've asked @microhobby to add here all changes that I've made on my local branch.~~

Edit: Just got direct access to the forked repo and I can push to it now.

jsrc27 (approved):
Gave the latest revision a look through and I don't think I have any further comments, seems okay from my end.

jsrc27:
@rborn-tx Did you want to take a look at this as well?

rborn-tx:
> Did you want to take a look at this as well?

@jsrc27 Yes! I hope to be able to do it today.

torizoncore-builder.py
```diff
@@ -30,6 +30,14 @@
...
 # pylint: enable=wrong-import-position
 
+# if the TCB_DEBUGPY environment variable is set
+# means that a developer is trying to debug the code from outside
+# so, we wait for a debugger to attach to the process
+if os.getenv("TCB_DEBUGPY"):
+    import debugpy
+    debugpy.listen(("0.0.0.0", 5678))
+    debugpy.wait_for_client()
+
```
rborn-tx, line 40: The goal here is to allow TCB to be debugged, right? Any reason why we want this? Who is the user of such a feature?

tcbuilder/backend/bundle.py
```diff
@@ -87,7 +87,7 @@ def get_client(self):
         """Create an instance of the Docker client"""
         return docker.from_env()
 
-    def save_tar(self, output_file):
+    def save_tar(self, output_file, compress=True):
```
rborn-tx: For my understanding: what's the purpose of not compressing the image?

tcbuilder/backend/combine.py
```diff
@@ -178,11 +184,15 @@ def combine_single_tezi_image(bundle_dir, files_to_add, output_dir, tezi_props):
...
+def check_combine_files(bundle_dir, is_for_raw_image=False):
 
     files_to_add = []
     if bundle_dir is not None:
-        files_to_add = DOCKER_FILES_TO_ADD
+        if not is_for_raw_image:
+            files_to_add = DOCKER_FILES_TO_ADD
+        else:
+            files_to_add = DOCKER_FILES_TO_ADD_TO_RAW
+
```
rborn-tx: Related to a previous question: what's the relationship between the image being "raw" and the compression being active or not? Is this some kind of optimization?

tcbuilder/backend/combine.py
```diff
@@ -21,6 +21,12 @@
     TARGET_NAME_FILENAME + ":/ostree/deploy/torizon/var/sota/storage/docker-compose/"
 ]
 
+DOCKER_FILES_TO_ADD_TO_RAW = [
+    "docker-compose.yml:/ostree/deploy/torizon/var/sota/storage/docker-compose/",
+    DOCKER_BUNDLE_FILENAME_UNCOMPRESSED + ":/ostree/deploy/torizon/var/lib/docker/:true",
+    TARGET_NAME_FILENAME + ":/ostree/deploy/torizon/var/sota/storage/docker-compose/"
+]
+
```
rborn-tx: I'd prefer to build variable `DOCKER_FILES_TO_ADD` dynamically (i.e. make it a local variable to the relevant function (`check_combine_files()`) and fill in its contents) based on whether compression is active or not instead of duplicating the whole variable.

tcbuilder/cli/build.py
```diff
@@ -348,6 +348,16 @@ def handle_raw_image_output(props, storage_dir, union_params, default_base_raw_i
 
     deploy_cli.deploy_raw_image(**deploy_raw_image_params)
 
+    handle_raw_image_bundle_output(
+        output_raw_img,
+        output_raw_img,
+        storage_dir,
+        props.get("bundle", {}),
+        props
+    )
```
rborn-tx: nit pick: how about indenting the same way as in `handle_easy_installer_output`.

tcbuilder/cli/build.py
```diff
@@ -438,24 +444,85 @@ def handle_bundle_output(image_dir, storage_dir, bundle_props, tezi_props):
...
+            shutil.rmtree(bundle_dir)
+
+
+def handle_raw_image_bundle_output(
+        image_dir,
+        raw_image_path,
+        storage_dir,
+        bundle_props,
+        raw_props
+    ):
```
rborn-tx: nit pick: odd indentation (why not follow the same pattern as `handle_bundle_output()`?)

rborn-tx (commented):
I'm halfway through the commits but I've added some comments and questions. I hope to go over the remaining commits soon...

tcbuilder/backend/combine.py
```diff
@@ -184,16 +188,13 @@ def combine_single_tezi_image(bundle_dir, files_to_add, output_dir, tezi_props):
...
+def check_combine_files(bundle_dir):
 
     files_to_add = []
     if bundle_dir is not None:
-        if not is_for_raw_image:
-            files_to_add = DOCKER_FILES_TO_ADD
-        else:
-            files_to_add = DOCKER_FILES_TO_ADD_TO_RAW
-
+        files_to_add = DOCKER_FILES_TO_ADD
```
rborn-tx: @lucas-akira I see here you're already tackling the issue I raised before about having two versions of this variable. ~So, please ignore that previous comment.~ Ive already resolved that question.

However, here I'd suggest copying the global variable rather than just referencing it because now we're making changes to it which wasn't true before. Changing global variables makes the code harder to reuse. For example this function would return a different result if called twice.

tcbuilder/cli/bundle.py
```diff
@@ -51,7 +51,7 @@ def bundle(bundle_dir, compose_file, force=False, keep_double_dollar_sign=False,
 
     bundle_be.download_containers_by_compose_file(
         bundle_dir, compose_file, host_workdir,
-        output_filename=common.DOCKER_BUNDLE_FILENAME,
+        output_filename=f"{common.DOCKER_BUNDLE_TARNAME}.xz",
```
rborn-tx, line 54: Doesn't this setting make command "bundle" incompatible with the raw image? Or is this expected?

tcbuilder/cli/build.py
```diff
@@ -398,12 +398,26 @@ def handle_easy_installer_output(props, storage_dir, union_params):
 def handle_bundle_common(
         storage_dir,
         bundle_props,
-        is_compressed_tar=True
+        compress_tar=True
```
rborn-tx: Reminder: follow normal indentation.

rborn-tx (commented):
Finished 1st pass. Looks good overall. It would be advisable to have tests for the new feature(s), right?

tcbuilder/backend/combine.py
rborn-tx: Resolving this since it was solved an a later commit in this PR.

torizoncore-builder.py
microhobby: Developers. This should not have any impact if the `TCB_DEBUGPY` is not set, and adds a great value. I'm not a TCB expert and I was able to make this contribution in "record" time, let's say, only because I was able to attach to debug and follow the flow to better understand what was happening.

tcbuilder/backend/bundle.py
microhobby: For raw images, the file will not be shared, all flow happens locally. Docker data is downloaded -> compressed -> decompressed -> injected into rootfs. Not compressing the file means less processing time, both for compressing and decompressing, speeding up the process: docker data downloaded -> injected into rootfs.

tcbuilder/cli/build.py
microhobby: this one has two more arguments, I think that is better to read. But, this is a decision from the team, if you have a strict code style better to follow up.

tcbuilder/backend/bundle.py
lucas-akira: To give more context: @microhobby originally removed the `docker-storage.tar` compression for raw images during the `build` command (if passing a docker-compose file) and I decided to keep it.

The reasoning that I see is this: When using a compose file to bundle container images to a raw OS image with the `build` command, TCB generates a temporary bundle directory with the tarred container images (`docker-storage.tar` file) during the bundling step itself. Immediately after that, the container images are unpacked inside the raw image rootfs during the combine step.

If the tarfile is compressed, the unpack step takes considerably longer compared to an uncompressed file. Given that the main reason to compress the file is to save on disk space, it doesn't seem particularly useful to add compression to a temporary file that would be deleted at the end of the command either way.

For TEZI images the compression is still relevant as the docker-storage tarfile itself will be copied to the output image directory.

That being said, I do think that compression on `docker-storage.tar` for raw images should be done if the bundle directory is not temporary i.e. kept on disk, like when using the commands individually (`bundle` then `combine`) or if passing an already existing bundle directory instead of a docker-compose file during the `build` command.

tcbuilder/backend/combine.py
lucas-akira: Oops, you're right. I didn't notice I ended up changing the global variable. Creating a shallow copy of the list like below should be enough in this case, right?

```
files_to_add = DOCKER_FILES_TO_ADD.copy()
```

rborn-tx: Yes, the data is expected to be 1-level deep always.

torizoncore-builder.py
rborn-tx: Ok, seems sensible to me.

tcbuilder/backend/bundle.py
rborn-tx: Agree.

tcbuilder/cli/bundle.py
lucas-akira: It shouldn't, because the `combine` backend code detects if the tarfile to be unpacked to the raw image rootfs has certain types of compression applied to it, and acts accordingly. XZ compression is supported and during my tests a .xz tarfile works without issues when combining container images with raw images.

tcbuilder/cli/build.py
rborn-tx: It's just a nit pick. Since it still fits into 100 columns we could follow the original indentation.

tcbuilder/cli/bundle.py
rborn-tx: Ok then.

tcbuilder/cli/build.py
lucas-akira: You mean indenting like this?
```
    handle_raw_image_bundle_output(
        output_raw_img, output_raw_img, storage_dir, props.get("bundle", {}), props)
```

lucas-akira: Personally I'm fine with both styles, but I'll change the indentation as @rborn-tx suggested, just for the sake of consistency with the rest of the code.

tcbuilder/backend/combine.py
lucas-akira: Alright, should be fixed now. Thanks!

tcbuilder/cli/build.py
lucas-akira: OK, done.

tcbuilder/cli/build.py
lucas-akira: Should be solved.

tcbuilder/cli/build.py
lucas-akira: Should be solved now.

lucas-akira:
> Finished 1st pass. Looks good overall. It would be advisable to have tests for the new feature(s), right?

Right, I'll add the tests then.

tcbuilder/backend/combine.py
lucas-akira: I guess this question has already been addressed in the previous conversation, right?

rborn-tx: Yes!

rborn-tx (approved):

lucas-akira (approved):

PR #33, opened 2024-11-06
tests/integration: Fix raw image test when running on a busybox host

lucas-akira:
One raw image test for the combine command uses the 'truncate' command to create a temporary file. The current command syntax is only valid for the GNU coreutils version of truncate, and is invalid for the busybox one.

As our GitLab test pipeline runs in container 'docker:latest', which is an environment with busybox, this particular test started failing in it due to the invalid syntax.

This commit changes the syntax so that it works for both the GNU and the busybox version of truncate.


jsrc27 (approved):
LGTM

PR #34, opened 2024-11-07
Fix 'images download' test when device has Torizon OS 7 installed

lucas-akira:
Before checking if the downloaded image tarfile has the expected ownership, the test assumes the filename starts with 'torizon-core', which is no longer the case with Torizon OS 7 images, which dropped the 'core' part of the name. This caused the test to fail as it cannot find the image with this assumption.

To solve this the logic to get the tarfile name was changed as suggested by the TODO comment: the test now obtains the filename from the program output.


jsrc27 (approved):
LGTM

PR #35, opened 2024-11-21
backend/bundle: Bump DIND container image version to 25.0.3

lucas-akira:
Bump Docker-in-Docker container image used in the bundle command to match the Docker version included on Torizon OS 7 images.

Related-to: TCB-486

### DIFF
--- a/tcbuilder/backend/bundle.py
+++ b/tcbuilder/backend/bundle.py
@@ -128,7 +128,7 @@ class DindManager(DockerManager):
     Installer tool.
     """
 
-    DIND_CONTAINER_IMAGE = "docker:19.03.8-dind"
+    DIND_CONTAINER_IMAGE = "docker:25.0.3-dind"
     DIND_VOLUME_NAME = "dind-volume"
     DIND_CONTAINER_NAME = "tcb-fetch-dind"
     TAR_CONTAINER_NAME = "tcb-build-tar"


jsrc27:
Just checking, this updated version of the Docker Daemon in the dind container now has the CLI options for proxy configuration correct? Wanted to make sure the Docker documentation was actually accurate on this detail.

lucas-akira:
Just did a quick check here, and the proxy options are listed in the help section:

```
$ docker run -it --privileged docker:25.0.3-dind /bin/sh
/ # dockerd --version
Docker version 25.0.3, build f417435
/ # dockerd --help

Usage:  dockerd [OPTIONS]

A self-sufficient runtime for containers.

Options:
[...]
  -H, --host list                               Daemon socket(s) to connect to
      --host-gateway-ip ip                      IP address that the special 'host-gateway' string in --add-host resolves
                                                to. Defaults to the IP address of the default bridge
      --http-proxy string                       HTTP proxy URL to use for outgoing traffic
      --https-proxy string                      HTTPS proxy URL to use for outgoing traffic
[...]
      --no-proxy string                         Comma-separated list of hosts or IP addresses for which the proxy is skipped
```

These options are not listed with the previous DinD image (19.03.8).

So now it should be possible to use a proxy server with this version of the DinD container.

rborn-tx:
We should probably consider a way to passing environment variables to DinD. If we had that I imagine we'd be able to set the proxy even with the older DinD (with the usual `HTTP_PROXY`, `HTTPS_PROXY` variables), right?

jsrc27 (approved):
LGTM

lucas-akira:
>We should probably consider a way to passing environment variables to DinD. If we had that I imagine we'd be able to set the proxy even with the older DinD (with the usual HTTP_PROXY, HTTPS_PROXY variables), right?

Yeah, we could create an option to pass environment variables to DinD on the `bundle` command like `--dind-env` or something like that.

I guess we can investigate and implement this in another PR, as it will require changes to more parts of the code, like implementing the CLI argument then its respective tcbuild.yaml field for the `build` command as well.

jsrc27:
Have we decided then to merge this as is, or are we going to do the environment variable stuff here as well?

lucas-akira:
On my side we can merge now, and I'll create a separate PR to pass environment variables to DinD.

rborn-tx:
@jsrc27 

> Have we decided then to merge this as is, or are we going to do the environment variable stuff here as well?

I think we can do the environment stuff as a separate improvement as @lucas-akira said.

rborn-tx (approved):
LGTM

rborn-tx:
I'll go ahead and merge this since it's been approved by Jeremias.

PR #36, opened 2024-11-25
tests/integration: Implementation of AVAL on TCB integration tests

lucas-p-bernardes:
The goal of the implementation is to run all TCB tests (including those that access modules) daily or weekly, using Gitlab CI `schedule` and the AVAL infrastructure.

The variables of AVAL were set in my personal fork at `Settings -> CI/CD -> Variables`.

By default, a schedule run all the tests on one ARM32 and on one ARM64 devices, however it is possible to set `$AVAL_TEST_ALL`  variable as `True` to run the tests on all devices.

Related-to: TCCP-864

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,3 +1,6 @@
+include:
+  - "/tests/integration/aval/aval-tests.yml"
+
 # TODO: GitLab's documentation recommends specifying a version:
 #       See https://docs.gitlab.com/ee/ci/docker/using_docker_build.html#use-docker-in-docker
 #       Also in the services attribute.
--- a/tests/integration/README
+++ b/tests/integration/README
@@ -117,30 +117,30 @@ The command below will execute all test cases, including those that
 requires device access (the image used in the tests will be
 automatically selected based on the device):
 
-$ TCB_DEVICE=192.168.1.210 ./run.sh
+$ TCB_DEVICE=192.168.1.210 TCB_MACHINE=apalis-imx6 DEVICE_PASSWORD=<password> ./run.sh
 
 The command below will execute the test cases defined in
 testcases/dto.bats, using an Apalis iMX6 image, and a test report
 will be available in workdir/reports/:
 
-$ TCB_REPORT=1 TCB_TESTCASE=dto ./run.sh
+$ TCB_REPORT=1 TCB_TESTCASE=dto TCB_MACHINE=apalis-imx6 ./run.sh
 
 The command below will execute the test cases defined in both
 testcases/dt.bats and testcases/dto.bats:
 
-$ TCB_TESTCASE="dt dto" ./run.sh
+$ TCB_TESTCASE="dt dto" TCB_MACHINE=apalis-imx6 ./run.sh
 
 The command below will execute all test cases, including those that
 require device access, using a local torizoncore-builder container
 tagged with 'v1':
 
-$ TCB_CUSTOM_IMAGE=torizoncore-builder:v1 TCB_DEVICE=192.168.1.210 ./run.sh
+$ TCB_CUSTOM_IMAGE=torizoncore-builder:v1 TCB_DEVICE=192.168.1.210 TCB_MACHINE=apalis-imx6 DEVICE_PASSWORD=<password> ./run.sh
 
 The command below will execute all the tests:
-$ TCB_TAGS="all" ./run.sh
+$ TCB_TAGS="all" TCB_MACHINE=apalis-imx6 ./run.sh
 
 The command below will execute untagged tests:
-$ TCB_TAGS="" ./run.sh
+$ TCB_TAGS="" TCB_MACHINE=apalis-imx6 ./run.sh
 
 The command below will execute tests tagged as 'requires-device':
-$ TCB_TAGS="requires-device" TCB_DEVICE=192.168.1.210 ./run.sh
+$ TCB_TAGS="requires-device" TCB_DEVICE=192.168.1.210 TCB_MACHINE=apalis-imx6 DEVICE_PASSWORD=<password> ./run.sh
--- a/tests/integration/aval/aval-tests.yml
+++ b/tests/integration/aval/aval-tests.yml
@@ -1,0 +1,125 @@
+# Build the AVAL container with docker that will run the tests on the host PC
+build-aval-docker:
+  image: docker:dind
+  stage: build
+  rules:
+    - if: $CI_PIPELINE_SOURCE == "schedule"
+  variables:
+    IMAGE_NAME_AVAL: aval-docker
+    DOCKERFILE_NAME: tests/integration/aval/aval-docker.Dockerfile
+  before_script:
+    - docker login -u "${CI_REGISTRY_USER}" -p "${CI_REGISTRY_PASSWORD}" "${CI_REGISTRY}"
+    - docker login -u "${CI_DOCKER_HUB_PULL_USER}" -p "${CI_DOCKER_HUB_PULL_PASSWORD}"
+  script:
+    - ${B} docker build -f "${DOCKERFILE_NAME}" -t "${CI_REGISTRY_IMAGE}/${IMAGE_NAME_AVAL}:main" .
+    - ${B} docker push "${CI_REGISTRY_IMAGE}/${IMAGE_NAME_AVAL}:main"
+
+.aval-template:
+  stage: test
+  needs:
+    - build-aval-docker
+    - build-amd64
+  variables:
+    IMAGE_NAME: torizoncore-builder-amd64
+    IMAGE_NAME_AVAL: aval-docker
+  rules:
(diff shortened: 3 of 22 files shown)


lucas-p-bernardes:
@jsrc27 @rborn-tx @lucas-akira 
I think I don't have permission to set you as reviewers, please do this for me :-)

tests/integration/aval/aval-tests.yml
```diff
@@ -36,6 +36,18 @@ build-aval-docker:
     - cd tests/integration && python /aval/main.py --delegation-config $DELEGATION_CONFIG
       --run-before-on-host "./run_all.sh --device --machine $TCB_MACHINE --report
       --tcb-custom-image ${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}"
+      # workaround for cloud credentials bug when using TCB deploy
```
leonheldattoradex: The comment doesn't explain the bug properly - it's not related to the cloud credentials. I suggest something like "to circumvent the mismatch between Aktualizr's database and ostree when an external entity manages ostree without Aktualizr, install a different version than the one Aktualizr thinks it's currently installed (the one before tcb deploy), which externally forces both Aktualizr and ostree to sync. In particular, if we're running a nightly, install a release build and vice-versa." or something.

tests/integration/aval/aval-tests.yml
```diff
@@ -0,0 +1,138 @@
...
+build-aval-docker:
+  image: docker:dind
+  stage: build
+  rules:
+    - if: $CI_PIPELINE_SOURCE == "schedule"
+  variables:
+    IMAGE_NAME_AVAL: aval-docker
+    DOCKERFILE_NAME: tests/integration/aval/aval-docker.Dockerfile
+  before_script:
+    - export DATE=$(date +%Y%m%d)
```
lucas-akira: `DATE` is not being used in this job.

tests/integration/run_all.sh
```diff
@@ -0,0 +1,161 @@
...
+    echo "  -r, --report               Generate test report"
+    echo "  --tcb-tags TAGS            Specify TCB tags"
+    echo "  --tcb-custom-image IMAGE   Use custom TorizonCore Builder image"
+    echo "  --                         Pass the rest of the arguments to test scripts"
+    echo
+    echo "Examples:"
+    echo "  $0 -d                      Run tests, using device connection if required"
+    echo "  $0 -t dto                  Run the 'dto' test case"
+    echo "  $0 --tcb-tags requires-device -d  Run tests with 'requires-device' tag, using device connection"
+    exit 1
```
lucas-akira: Program would be executed with no errors in this case, so the exit code here should be `0`, right?

tests/integration/run_all.sh
```diff
@@ -0,0 +1,161 @@
...
+
+echo "Running setup..."
+source ./setup.sh
+
+if [ $? -ne 0 ]; then
+    echo "Error running setup.sh"
+    exit 1
+fi
+
+if [ ! -e workdir/images/.images_downloaded ] && [ ! -e workdir/images/.raw_images_downloaded ]; then
```
lucas-akira: Shouldn't it be a logical OR (`||`) here instead of an AND? 

Right now the script wouldn't run `get_tezi_images.sh` if `.raw_images_downloaded` exists beforehand for example, and it would proceed without any TEZI images downloaded.

tests/integration/aval/aval-tests.yml
lucas-p-bernardes: Done! Thanks!

tests/integration/run_all.sh
lucas-p-bernardes: Done! Thanks!

tests/integration/run_all.sh
lucas-p-bernardes: Done! Thanks!

tests/integration/aval/aval-tests.yml
lucas-p-bernardes: Done! Thanks!

.gitlab-ci.yml
```diff
@@ -1,8 +1,10 @@
 # TODO: GitLab's documentation recommends specifying a version:
```
leonheldattoradex: This TODO should be moved down to line 7.

lucas-p-bernardes: Done, thanks Leon!

lucas-p-bernardes:
Just to clarify here, I'm running only the tests that use device with this AVAL implementation. This way causes less overload on infrastructure.

tests/integration/aval/aval-tests.yml
```diff
@@ -0,0 +1,140 @@
...
+        elif [ "$current_release_type" == "nightly" ]; then
+          echo "release"
+        fi
+      }
+    - TARGET_BUILD_TYPE=$(toggle_release_type "$TARGET_BUILD_TYPE") python /aval/main.py --delegation-config $DELEGATION_CONFIG
+      "echo 'Updating to a different TOS version than the TCB tests version, to circumvent the mismatch between Aktualizr'\''s database and ostree when an external entity manages ostree without Aktualizr'"
+    - (! grep "^not ok" workdir/reports/*)
+
+# $AVAL_TEST_ALL can be defined on a Gitlab schedule to run tests on all devices
+aval-apalis-imx6q-scarthgap-release:
```
leonheldattoradex, line 56: Do we *need* to test all these boards? It could be best to simply choose the two most powerful 32/64-bits SoMS instead.

lucas-p-bernardes: By default, the tests will only run on the Apalis iMX6 and Verdin iMX8MP. If, for some reason, it is necessary to run the tests for all devices, the user should simply set AVAL_TEST_ALL=True.

I created jobs for all devices in case of some specific TCB tests need it.

rborn-tx: @lucasbernardestoradex Tricky question: what happens if you set `AVAL_TEST_ALL=False`?

lucas-p-bernardes: @rborn-tx Good point, the logic was "if AVAL_TEST_ALL is defined, run on all boards".
Actually, this was not what I wanted, so I changed to "if AVAL_TEST_ALL is defined as TRUE, run on all boards".

tests/integration/README
```diff
@@ -117,13 +117,13 @@ The command below will execute all test cases, including those that
...
 
-$ TCB_DEVICE=192.168.1.210 ./run.sh
+$ TCB_DEVICE=192.168.1.210 DEVICE_PASSWORD=<password> ./run.sh
 
 The command below will execute the test cases defined in
 testcases/dto.bats, using an Apalis iMX6 image, and a test report
 will be available in workdir/reports/:
 
-$ TCB_REPORT=1 TCB_TESTCASE=dto ./run.sh
+$ TCB_REPORT=1 TCB_TESTCASE=dto TCB_MACHINE=apalis-imx6 DEVICE_PASSWORD=<password> ./run.sh
```
rborn-tx: Shouldn't a `DEVICE_PASSWORD` be needed only if `TCB_DEVICE` was also specified?

tests/integration/run.sh
```diff
@@ -47,7 +47,6 @@ export SAMPLES_DIR=samples
 # device address and credentials
 export DEVICE_ADDR=$TCB_DEVICE
 export DEVICE_USER="torizon"
-export DEVICE_PASS="1"
```
rborn-tx, line 50: Now that DEVICE_PASS was renamed to DEVICE_PASSWORD and the latter must be passed whenever TCB_DEVICE is set (right?) maybe we should have a check here to see if the user is passing both pieces of information.

rborn-tx (commented):
@lucasbernardestoradex Great job!

I raised just a couple of questions that could lead to some minor improvements.

tests/integration/README
lucas-p-bernardes: Yeah, you are correct! I missed some points on README.
Fixing this!

lucas-p-bernardes: Done!

tests/integration/run.sh
lucas-p-bernardes: I did this check on last pushs! Thanks @rborn-tx!

rborn-tx (approved):
LGTM

lucas-akira (approved):
Latest revision looks good to me. Approving.

leonheldattoradex:
@lucasbernardestoradex please guarantee that the latest changes are working in a pipeline on our GitLab instance. You probably did this, but just to checklist an item to avoid rework later.

lucas-p-bernardes:
@rborn-tx @jsrc27 Can we have this merged?

jsrc27 (approved):
LGTM. Merging now.

PR #37, opened 2024-12-04
Make device-required tests independent of tests that don't use a device

lucas-akira:
Some tests that require a device use data created by previous tests that do not require one. This causes problems when only running device-required tests via the 'requires-device' BATS tag, as the necessary data wouldn't be created in this case.

This commit makes these two types of tests independent of one another.

The AVAL-related pipeline only run tests that require a device, so this PR fixes test errors caused by the problem described above.

### DIFF
--- a/tests/integration/testcases/build.bats
+++ b/tests/integration/testcases/build.bats
@@ -164,9 +164,33 @@ teardown_file() {
     requires-image-version "$DEFAULT_TEZI_IMAGE" "5.3.0"
     requires-device
 
-    # This test case assumes the previous one was executed.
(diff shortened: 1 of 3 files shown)


lucas-p-bernardes:
I noticed the pipeline on Gitlab instance passed on AVAL jobs.
LGTM.

jsrc27 (approved):
LGTM

lucas-p-bernardes:
@rborn-tx @jsrc27 Can we have this merged? I need this changes to work on a ticket!

rborn-tx (approved):
LGTM

PR #40, opened 2024-12-12
backend/platform: fetch latest director root.json on lockbox download

bclouser:
The director service in torizon cloud will ensure the root.json expires later than all offline-metadata as part of servicing the request to GET:api/v1/admin/repo/root.json. If necessary, the director will create a new version of root.json with a later expiration.

It is imperative to trigger this validation during the creation of a lockbox to prevent issues where a user cannot install a valid lockbox because the root metadata has expired.

Fetching the root metadata directly with the version (ie: 1.root.json) will not trigger this validity check, and therefore must be preceeded by fetching the naked root.json

### DIFF
--- a/tcbuilder/backend/platform.py
+++ b/tcbuilder/backend/platform.py
@@ -866,28 +866,46 @@ def fetch_director_metadata(lockbox_name, director_url, dest_dir, access_token=N
     # ---
     # Get all versions of root metadata.
     # ---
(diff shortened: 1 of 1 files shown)


lucas-akira (approved):
LGTM

jsrc27 (approved):
LGTM

rborn-tx (approved):
LGTM as well.

PR #41, opened 2024-12-26
Add support for passing environment variables to DIND

rborn-tx:
Add support for passing environment variables to the Docker-in-Docker container used by the tool to fetch container images for bundling.

Passing environment variables to that container is useful in many cases. For example, one can set the `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` variables to define which proxy to use in different cases.

In this PR we introduce a new switch `--dind-env` to commands `bundle` and `platform lockbox` plus we add properties `dind-param` and `dind-env` to the `tcbuild` file schema. With this, the commands `bundle`, `platform lockbox` and `build` now support passing both parameters and environment variables to the DinD instance used for bundling.

Another feature introduced here is the ability to forward environment variables from the TorizonCore Builder container into the DIND container. This will be useful to simplify using the tool with a proxy server.


rborn-tx:
Merging this without review to have it available on the next early-access version.

PR #43, opened 2025-01-30
Update TCB tests to run on latest TOS versions/ Be able to run AVAL on kirkstone

lucas-p-bernardes:
Related-to: AVAL-18
Related-to: TCB-502

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -52,7 +52,6 @@ services:
 stages:
   - pre-build
   - build
-  - test-platform
   - test
   - build-multiarch
   - deploy
@@ -183,7 +182,7 @@ build-arm64:
 
 test-host-tc6-platform:
   extends: .test-base
-  stage: test-platform
+  stage: test
   variables:
     TCB_TESTCASE: platform
     YOCTO_BRANCH: "kirkstone-6.x.y"
@@ -194,7 +193,7 @@ test-host-tos7-platform:
   needs:
     - test-host-tc6-platform
   extends: .test-base
-  stage: test-platform
+  stage: test
   variables:
     TCB_TESTCASE: platform
     YOCTO_BRANCH: "scarthgap-7.x.y"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -27,13 +27,6 @@ variables:
   DOCKER_BUILDKIT: 1
   GITLAB_DOCKERREGISTRY_SUFFIX: ${CI_COMMIT_REF_SLUG}-${CI_PIPELINE_ID}
   GITLAB_DOCKERREGISTRY_SUFFIX_LATEST: ${CI_COMMIT_REF_SLUG}-latest
-  # TC image default versions:
-  TEST_IMAGES_TC6_VERSION: "6.2.0"
-  TEST_IMAGE_TC6_32BIT: "torizon-core-docker-colibri-imx7-emmc-Tezi_6.2.0-devel-20230324+build.220.tar"
-  TEST_IMAGE_TC6_64BIT: "torizon-core-docker-apalis-imx8-Tezi_6.2.0-devel-20230324+build.220.tar"
-  TEST_IMAGES_TOS7_VERSION: "7.0.0"
-  TEST_IMAGE_TOS7_32BIT: "torizon-docker-apalis-imx6-Tezi_7.0.0-devel-20240830+build.40.tar"
-  TEST_IMAGE_TOS7_64BIT: "torizon-docker-apalis-imx8-Tezi_7.0.0-devel-20240830+build.40.tar"
   # Common Torizon .wic/.img images:
   TEST_IMAGES_TC_COMMON_VERSION: "6.6.0-common"
   TEST_IMAGE_TC_COMMON_INTEL: "torizon-core-common-docker-dev-intel-corei7-64-20240226043251.rootfs.wic"
@@ -168,12 +161,9 @@ build-arm64:
     - when: on_success
   variables:
     IMAGE_NAME: torizoncore-builder-amd64
-    # Don't run 'platform' tests
-    TCB_TESTCASES_32BIT: "torizoncore-builder images images-unpack images-download images-serve union deploy isolate dt dto build bundle bundle-registry combine kernel ostree push splash"
-    TCB_TESTCASES_64BIT: "dt dto kernel"
   before_script:
     # setup environment for the tests
-    - apk update && apk add git sshpass zstd wget bash tar openssl zip
+    - apk update && apk add git sshpass zstd wget bash tar openssl zip curl jq
     - cd tests/integration/
     - mkdir -p workdir/reports && rm -Rf workdir/reports/*
     - TCB_SKIP_PULL=1 ./setup.sh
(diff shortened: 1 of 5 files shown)


lucas-p-bernardes:
Refer to latest pipeline of `AVAL-18_TCB-502` branch on my gitlab fork to see the results.

jsrc27 (approved):
LGTM

tests/integration/get_tezi_images.sh
```diff
@@ -1,94 +1,84 @@
...
-verdin-imx8mp:torizon:torizon-docker \
-verdin-am62:torizon:torizon-docker \
-"
-
-# location to save images
+# Location to save images
 OUTDIR="$PWD/workdir/images"
-TMPDIR="$OUTDIR/tmp"
 STAMP="$OUTDIR/.images_downloaded"
+ARTIFACTORY_URL="http://artifactory-horw.int.toradex.com"
```
rborn-tx: This `get_tezi_images.sh` script is for any developer to use (even people outside the company) but here we're referencing the internal Artifactory. Is that intentional??

tests/integration/get_tezi_images.sh
```diff
@@ -1,94 +1,84 @@
...
+        exit 1
+    fi
 
-# CONFIGME: uncomment for monthly images
-#BUILD_RELEASE="prerelease"
-#BUILD_TYPE="monthly"
-#VERSION="${TCVERSION}-devel-${BUILD_DATE}+build.${BUILD_NUMBER}"
+    # TODO:This feature of catch latest build is only available on internal TEZI yet. When this is ready on public one, adapt this code to use
+    # the public
+    wget --no-verbose -P workdir/ "https://tezi.int.toradex.com/tezifeed/build/image-torizoncore-$YOCTO_BRANCH-$TARGET_BUILD_TYPE-matrix/latest"
```
rborn-tx: I see the comment here, but is it acceptable to access the internal Tezi feed?

tests/integration/get_tezi_images.sh
```diff
@@ -1,94 +1,88 @@
...
-        if ! wget --no-verbose -P $TMPDIR $LINK; then
-            echo "Error: could not download $IMAGE_FILE."
-            exit 1
-        fi
-        mv $TMPDIR/$IMAGE_FILE $OUTDIR/
-        echo ""
-    fi
-}
+        version_number=$(echo "$json_content" | jq -r '.version' | sed 's#+build.*##')
+        build_number=$(echo "$json_content" | jq -r '.version' | sed -E 's#.*+build\.([0-9]+)#\1#')
```
rborn-tx: Bad indentation.

tests/integration/get_tezi_images.sh
```diff
@@ -1,94 +1,88 @@
...
-    rm -Rf $TMPDIR
-}
+        torizon_name=$(echo "$img_json_url_no_eval" | sed -E 's#.*/(torizon|torizon-upstream)/(.*)\/teziimage/.*#\2#')
+        device_name=$(echo "$img_json_url_no_eval" | sed -E 's#.*/(.*)/(torizon|torizon-upstream)/.*#\1#')
 
-download_tezi_image() {
-    OE_MACHINE=$1
-    OE_DISTRO=$2
-    OE_IMAGE=$3
+        if [[ "$TCB_MACHINE" = "$device_name" ]]; then
```
rborn-tx: I'd suggest negating this logic to reduce one level of nesting of the rest of the block. Also I see no reason for using the `[[` operator here (i.e. avoid unnecessary Bashism). Something like this:

```
if [[ "$TCB_MACHINE" = "$device_name" ]]; then
    # <big block>
else
    # <small block>
fi
```

Becomes:

```
if [ "$TCB_MACHINE" != "$device_name" ]; then
    # <small block>
    continue
fi
# <big block (with reduced nesting)>
```

tests/integration/get_tezi_images.sh
```diff
@@ -1,94 +1,88 @@
...
-# branch
-BRANCH="scarthgap-7.x.y"
+export image_count=""
+export img_json_url_eval=""
+export img_json_url_no_eval=""
+export device_name=""
+export json_content=""
+export version_number=""
+export build_number=""
+export torizon_url=""
```
rborn-tx: Shouldn't all these variable be local to `download()`?

tests/integration/testcases/setup.bats
```diff
@@ -0,0 +1,18 @@
+bats_load_library 'bats/bats-support/load.bash'
+bats_load_library 'bats/bats-assert/load.bash'
+bats_load_library 'bats/bats-file/load.bash'
+
+
+
```
rborn-tx: Extraneous empty lines.

tests/integration/testcases/setup.bats
```diff
@@ -0,0 +1,18 @@
...
+# bats test_tags=requires-device
+@test "Check if versions of file and installed on device matches" {
+    requires-device
+
+    DEVICE_INFO=$(device-shell cat /etc/os-release)
+    installed_version=$(echo "$DEVICE_INFO" | grep VERSION_ID | cut -d= -f2 | tr -d '"')
+
+    downloaded_version=$(basename "$(ls images)")
```
rborn-tx: Couldn't a user have multiple files and/or directories under `images`? I imagine they could in which case this wouldn't work very well right?

tests/integration/testcases/setup.bats
```diff
@@ -0,0 +1,18 @@
...
+    downloaded_version=$(echo "$downloaded_version" | tr ' ' '\n' | grep "torizon.*-"$TCB_MACHINE"-Tezi_.*\.tar" | sed 's/.*Tezi_//;s/\.tar//' | sed 's/+/-/')
```
rborn-tx: You could split the line here to avoid too wide lines. E.g.:

```
downloaded_version=$(echo "$downloaded_version" | tr ' ' '\n' |
			 grep "torizon.*-"$TCB_MACHINE"-Tezi_.*\.tar" |
			 sed 's/.*Tezi_//;s/\.tar//' | sed 's/+/-/')
```

tests/integration/aval/aval-tests.yml
rborn-tx, line 1: As a note, I guess we could be running many test jobs in parallel perhaps in the same runner which could be very resource intensive. Not sure if we should be concerned about it anyways.

.gitlab-ci.yml
```diff
@@ -168,78 +160,81 @@ build-arm64:
...
-    TCB_TESTCASES_32BIT: "platform"
-    TCB_TESTCASES_64BIT: "torizoncore-builder"
+    # Don't run 'platform' tests
+    TCB_TESTCASE: torizoncore-builder images images-unpack images-download images-serve union deploy isolate dt dto build bundle bundle-registry combine kernel ostree push splash
+    YOCTO_BRANCH: "kirkstone-6.x.y"
+    TARGET_BUILD_TYPE: "release"
+    TCB_MACHINE: "apalis-imx8"
 
-test-torizoncore6:
+test-host-tos7-32bits:
```
rborn-tx: To keep the naming pattern: `test-host-tos7-32bits` -> `test-host-tos7-32bit`

.gitlab-ci.yml
```diff
@@ -168,78 +160,81 @@ build-arm64:
...
-    TEST_IMAGE_32BIT: ${TEST_IMAGE_TC6_32BIT}
-    TEST_IMAGE_64BIT: ${TEST_IMAGE_TC6_64BIT}
+    # Don't run 'platform' tests
+    TCB_TESTCASE: torizoncore-builder images images-unpack images-download images-serve union deploy isolate dt dto build bundle bundle-registry combine kernel ostree push splash
+    YOCTO_BRANCH: "scarthgap-7.x.y"
+    TARGET_BUILD_TYPE: "release"
+    TCB_MACHINE: "apalis-imx6"
 
-test-torizonos7:
+test-host-tos7-64bits:
```
rborn-tx: To keep the naming pattern: `test-host-tos7-64bits` -> `test-host-tos7-64bit`

rborn-tx (changes requested):
Nice changes! I've raised some points for review though...

tests/integration/get_tezi_images.sh
lucas-p-bernardes: I requested to Automation to implement this feature in public TEZI. They said they will priorize this in the next sprint, which starts 5th February. 
I think we can wait this, so I will move this PR to "Draft" until they launch this feature.

tests/integration/aval/aval-tests.yml
lucas-p-bernardes: AFAIK, in the `/etc/gitlab-runner/config.toml` file of the runners, there is a config that tells the runner the number of parallel jobs it can run. If the number of requested jobs exceeds this limit, some jobs will stay in the "awaiting runner" state.
So, I don't think this is a problem here. If the resources are available at the moment, I don't see a reason not to use them. If they are not available, GitLab CI will handle this with the "awaiting runner" status.

Otherwise, the default is run only 4 AVAL tests:
- 64 bit kirkstone
- 64 bit scarthgap
- 32 bit kirkstone
- 32 bit scarthgap

If the user wanna run tests in all devices for any reason, he needs to define `$AVAL_TEST_ALL = true`. For me, it's better to have the option to do this than don't have.

tests/integration/testcases/setup.bats
lucas-p-bernardes: Good point, it's better to use $DEFAULT_TEZI_IMAGE, that is the image used in test scripts, instead of trying to determine what image is from the directory. I will solve ths.

tests/integration/get_tezi_images.sh
lucas-p-bernardes: Done!

tests/integration/get_tezi_images.sh
lucas-p-bernardes: Done!

tests/integration/get_tezi_images.sh
lucas-p-bernardes: Yeas, should be!
Done!

tests/integration/testcases/setup.bats
lucas-p-bernardes: Done!

tests/integration/testcases/setup.bats
lucas-p-bernardes: Done!

.gitlab-ci.yml
lucas-p-bernardes: Done!

.gitlab-ci.yml
lucas-p-bernardes: Done!

tests/integration/get_tezi_images.sh
lucas-p-bernardes: I changed the script to use public Artifact, with `HTML` parsing.

tests/integration/get_tezi_images.sh
lucas-p-bernardes: I decided the best way here was to change `get_tezi_images.sh` to not depend on this feature of Automation. Now, it is using direct `HTML` parsing.
I will re-open the MR.

lucas-p-bernardes:
@rborn-tx @lucas-akira I changed `get_tezi_images.sh` to use `HTML` parsing instead of `latest` feature from TEZI.
The reason was that I noticed this feature doesn't fit the needs for TCB. This catches the images for the latest buid number. Then, if we have only one machine image in a release (as we have now for Aquila AM69), this feature will get only this image and ignore the other machines.
What I needed was the latest images for every machine, not the latest images for the latest build number.

Please review considering this new `get_tezi_images.sh` script.

tests/integration/aval/aval-tests.yml
rborn-tx: Thanks for the clarification.

rborn-tx:
> Please review considering this new `get_tezi_images.sh` script.

This new version looks sensible to me.

rborn-tx (approved):

lucas-akira (approved):
Looks good to me as well.

PR #45, opened 2025-02-05
Update Aktualizr to a version supporting commitmeta objects plus cosmetic changes

rborn-tx:
- Follow the same indentation pattern in all RUN commands.
- Update aktualizr to add support for commitmeta objects; this means switching to the downstream aktualizr. In the future we may go back to the upstream aktualizr if https://github.com/uptane/aktualizr/pull/129 gets merged.
- Update to uptane-sign 3.2.6.

See commit messages for details.

Since the first commit is basically indentation involving the whole Dockerfile, I suggest reviewing it separately.

### DIFF
--- a/torizoncore-builder.Dockerfile
+++ b/torizoncore-builder.Dockerfile
@@ -17,11 +17,12 @@ RUN if [ "$APT_PROXY" != "" ]; then \
 # installing them twice (build-dep would install them too)
 # This are all dependencies from the regular Debian OSTree packages except
 # AVAHI.
-RUN apt-get -q -y update && apt-get -q -y --no-install-recommends install \
-    libarchive13 libassuan0 libfuse2 libglib2.0-0  libgpg-error0  libgpgme11 \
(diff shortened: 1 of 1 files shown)


jsrc27 (approved):
LGTM no major comments.

One small question though. I assume there was no breaking changes in the uptane sign tool? Just wondering since that's quite a version leap from the previous version we were using here.

lucas-akira (approved):
LGTM; No issues on my side.

rborn-tx:
> LGTM no major comments.
> 
> One small question though. I assume there was no breaking changes in the uptane sign tool? Just wondering since that's quite a version leap from the previous version we were using here.

I'm assuming the same :wink:.

Basically this is the version we're already using in Yocto. All our TCB tests pass with that new version of the CLI tools so my hope is that this is enough to say the new version is compatible with the old one at least for the small set of commands we use from `uptane-sign`. The [project](https://github.com/uptane/ota-tuf) itself doesn't seem to have a Release Notes document but checking the commit history I couldn't find messages about breaking changes.

PR #47, opened 2025-02-11
Add functionality to push new fuse type packages

jsrc27:
Did a number of run-time tests so I'm fairly confident on the behavior at least.

By the way I know we've already said this before. But, working on this really reminded me how "busy" the `platform` command is.

### DIFF
--- a/tcbuilder/backend/fuse.schema.yaml
+++ b/tcbuilder/backend/fuse.schema.yaml
@@ -1,0 +1,74 @@
+title: "Fuse update package configuration"
+description: "Schema for secure boot fuse update package files."
+type: object
+properties:
+  fuses:
+    type: object
+    description: "Fuse values to be programmed in-order"
+    properties:
+      fuse-close:
+        type: boolean
+        description: "Whether to set closing fuse on device"
+      fuse-val1:
+        type: number
+        description: "Hex representation of value to set to 1st fuse"
+      fuse-val2:
+        type: number
+        description: "Hex representation of value to set to 2nd fuse"
+      fuse-val3:
+        type: number
+        description: "Hex representation of value to set to 3rd fuse"
+      fuse-val4:
+        type: number
+        description: "Hex representation of value to set to 4th fuse"
+      fuse-val5:
+        type: number
+        description: "Hex representation of value to set to 5th fuse"
+      fuse-val6:
+        type: number
+        description: "Hex representation of value to set to 6th fuse"
+      fuse-val7:
+        type: number
+        description: "Hex representation of value to set to 7th fuse"
+      fuse-val8:
+        type: number
+        description: "Hex representation of value to set to 8th fuse"
+      fuse-val9:
+        type: number
+        description: "Hex representation of value to set to 9th fuse (AHAB only)"
+      fuse-val10:
+        type: number
+        description: "Hex representation of value to set to 10th fuse (AHAB only)"
+      fuse-val11:
+        type: number
+        description: "Hex representation of value to set to 11th fuse (AHAB only)"
+      fuse-val12:
+        type: number
+        description: "Hex representation of value to set to 12th fuse (AHAB only)"
+      fuse-val13:
+        type: number
+        description: "Hex representation of value to set to 13th fuse (AHAB only)"
+      fuse-val14:
+        type: number
+        description: "Hex representation of value to set to 14th fuse (AHAB only)"
+      fuse-val15:
+        type: number
+        description: "Hex representation of value to set to 15th fuse (AHAB only)"
+      fuse-val16:
+        type: number
+        description: "Hex representation of value to set to 16th fuse (AHAB only)"
+    additionalProperties: false
+    # Both HAB & AHAB requires the following
+    required:
+      - fuse-close
+      - fuse-val1
+      - fuse-val2
+      - fuse-val3
+      - fuse-val4
+      - fuse-val5
+      - fuse-val6
+      - fuse-val7
+      - fuse-val8
+additionalProperties: false
+required:
+  - fuses
--- a/tcbuilder/backend/platform.py
+++ b/tcbuilder/backend/platform.py
@@ -1246,40 +1246,16 @@ def push_compose(credentials, target, version, compose_file,
     if compatible_with:
         custom_metadata["compatibleWith"] = compatible_with
 
-    log.info(f"Pushing '{os.path.basename(push_file)}' with package version "
-             f"{version} to OTA server. You should keep this file under your "
+    log.info(f"You should keep '{os.path.basename(push_file)}' under your "
              "version control system.")
 
-    run_uptane_command(["uptane-sign", "init",
-                        "--credentials", credentials,
-                        "--repo", TUF_REPO_DIR], verbose)
-
-    run_uptane_command(["uptane-sign", "targets", "pull",
-                        "--repo", TUF_REPO_DIR], verbose)
-
-    run_uptane_command(["uptane-sign", "targets", "upload",
-                        "--repo", TUF_REPO_DIR,
-                        "--input", push_file,
-                        "--name", target,
-                        "--version", version,
-                        "--timeout", UPTANE_SIGN_UPLOAD_TIMEOUT], verbose)
-
-    run_uptane_command(["uptane-sign", "targets", "add-uploaded",
-                        "--repo", TUF_REPO_DIR,
-                        "--input", push_file,
-                        "--name", target,
-                        "--version", version,
-                        "--hardwareids", "docker-compose",
-                        "--customMeta", json.dumps(custom_metadata)], verbose)
-
-    run_uptane_command(["uptane-sign", "targets", "sign",
-                        "--repo", TUF_REPO_DIR,
-                        "--key-name", "targets"], verbose)
-
-    run_uptane_command(["uptane-sign", "targets", "push",
-                        "--repo", TUF_REPO_DIR], verbose)
-
-    log.info(f"Successfully pushed {os.path.basename(push_file)} to OTA server.")
+    uptane_sign_push(credentials=credentials,
+                     push_file=push_file,
+                     target=target,
+                     version=version,
+                     hardwareids_str="docker-compose",
+                     custom_metadata=custom_metadata,
+                     verbose=verbose)
 
     if description is not None:
         update_description(description, target, version, credentials)
@@ -1307,39 +1283,13 @@ def push_generic(credentials, target, version, generic_file,
     if compatible_with:
         custom_meta["compatibleWith"] = compatible_with
 
-    log.info(f"Pushing '{os.path.basename(generic_file)}' with package version "
-             f"{version} to OTA server.")
-
-    run_uptane_command(["uptane-sign", "init",
-                        "--credentials", credentials,
-                        "--repo", TUF_REPO_DIR], verbose)
-
-    run_uptane_command(["uptane-sign", "targets", "pull",
-                        "--repo", TUF_REPO_DIR], verbose)
-
-    run_uptane_command(["uptane-sign", "targets", "upload",
-                        "--repo", TUF_REPO_DIR,
-                        "--input", generic_file,
-                        "--name", target,
-                        "--version", version,
-                        "--timeout", UPTANE_SIGN_UPLOAD_TIMEOUT], verbose)
-
-    run_uptane_command(["uptane-sign", "targets", "add-uploaded",
-                        "--repo", TUF_REPO_DIR,
-                        "--input", generic_file,
-                        "--name", target,
-                        "--version", version,
-                        "--hardwareids", hardwareids_str,
-                        "--customMeta", json.dumps(custom_meta)], verbose)
-
-    run_uptane_command(["uptane-sign", "targets", "sign",
-                        "--repo", TUF_REPO_DIR,
-                        "--key-name", "targets"], verbose)
-
-    run_uptane_command(["uptane-sign", "targets", "push",
-                        "--repo", TUF_REPO_DIR], verbose)
-
-    log.info(f"Successfully pushed {os.path.basename(generic_file)} to OTA server.")
+    uptane_sign_push(credentials=credentials,
+                     push_file=generic_file,
+                     target=target,
+                     version=version,
+                     hardwareids_str=hardwareids_str,
+                     custom_metadata=custom_meta,
+                     verbose=verbose)
 
     if description is not None:
         update_description(description, target, version, credentials)
@@ -1561,5 +1511,53 @@ def get_shared_provdata(dest_file, repo_url, director_url, access_token=None):
         set_output_ownership(dest_file)
         log.info(f"Shared data archive '{dest_file}' successfully generated.")
 
+def uptane_sign_push(credentials, push_file, target, version,
+                     hardwareids_str, custom_metadata, verbose):
+    """
+    Use Uptane sign to push a file-based package to the OTA server
+
+    :param credentials: The credentials file corresponding to the OTA
+                        server that will be pushed to.
+    :param push_file: The file-based pacakge payload
+    :param target: The name of the package
+    :param version: The version of the package
+    :param hardwareids_str: A string of hardware-ids
+    :param custom_metadata: Custom metadata to be pushed with package
+    :param verbose: Whether uptane-sign will execute with verbose output
+    """
+
+    log.info(f"Pushing '{os.path.basename(push_file)}' with package version "
+             f"{version} to OTA server.")
+
+    run_uptane_command(["uptane-sign", "init",
+                        "--credentials", credentials,
+                        "--repo", TUF_REPO_DIR], verbose)
+
+    run_uptane_command(["uptane-sign", "targets", "pull",
+                        "--repo", TUF_REPO_DIR], verbose)
+
+    run_uptane_command(["uptane-sign", "targets", "upload",
+                        "--repo", TUF_REPO_DIR,
+                        "--input", push_file,
+                        "--name", target,
+                        "--version", version,
+                        "--timeout", UPTANE_SIGN_UPLOAD_TIMEOUT], verbose)
+
+    run_uptane_command(["uptane-sign", "targets", "add-uploaded",
+                        "--repo", TUF_REPO_DIR,
+                        "--input", push_file,
+                        "--name", target,
+                        "--version", version,
+                        "--hardwareids", hardwareids_str,
+                        "--customMeta", json.dumps(custom_metadata)], verbose)
+
+    run_uptane_command(["uptane-sign", "targets", "sign",
+                        "--repo", TUF_REPO_DIR,
+                        "--key-name", "targets"], verbose)
+
+    run_uptane_command(["uptane-sign", "targets", "push",
+                        "--repo", TUF_REPO_DIR], verbose)
+
+    log.info(f"Successfully pushed {os.path.basename(push_file)} to OTA server.")
 
 # EOF
--- a/tcbuilder/backend/platform.py
+++ b/tcbuilder/backend/platform.py
@@ -28,6 +28,7 @@ from tcbuilder.errors import \
 from tcbuilder.backend import ostree, sotaops
 from tcbuilder.backend.bundle import \
     (DindManager, login_to_registries, show_pull_progress_xterm)
+from tcbuilder.backend.build import parse_config_file
 from tcbuilder.backend.common import \
     (get_host_workdir, get_own_network, set_output_ownership, run_with_loading_animation,
      validate_compose_file)
@@ -59,6 +60,19 @@ TUF_REPO_DIR = "/deploy/tuf-repo"
 
 # SHA256 Hash Regex
 HASH_REGEX = re.compile(r"^[0-9a-f]{64}$")
+
+FUSE_HARDWAREIDS = {
+    "verdin-imx8mp-fuses": "hab",
+    "verdin-imx8mm-fuses": "hab",
+    "apalis-imx8-fuses": "ahab",
+    "apalis-imx6-fuses": "hab",
+    "colibri-imx8x-fuses": "ahab",
+    "colibri-imx7-emmc-fuses": "hab",
+    "colibri-imx6-fuses": "hab",
+    "colibri-imx6ull-emmc-fuses": "hab"
+}
+
+FUSE_SCHEMA_FILE = "fuse.schema.yaml"
 
 
 def load_metadata(fname, ftype=None, maxlen=DEFAULT_METADATA_MAXLEN):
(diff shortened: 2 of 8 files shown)


tcbuilder/backend/fuse.schema.yaml
```diff
@@ -0,0 +1,75 @@
+title: "Fuse update package configuration"
+description: "Schema for secure boot fuse update package files."
+type: object
+properties:
+  fuses:
+    type: object
+    description: "Fuse values to be programmed in-order"
+    oneOf:
```
rborn-tx: Since there's only one sub-schema here, we probably don't need this `oneOf`. In fact, IIRC having the `oneOf` ends up causing the parser to generate less useful error messages. For example, instead of saying that some property is missing it would say that the input did not match any of the schemas...

tcbuilder/backend/fuse.schema.yaml
```diff
@@ -0,0 +1,75 @@
...
+        required:
+          - fuse-close
+          - fuse-val1
+          - fuse-val2
+          - fuse-val3
+          - fuse-val4
+          - fuse-val5
+          - fuse-val6
+          - fuse-val7
+          - fuse-val8
```
rborn-tx: Since with some machines we'll want `fuse-val{1..8}` to be specified and `fuse-val{9..16}` not to be specified while for other machines we'll want all fuse values to be specified, maybe we could leave the `fuse-val{1..8}` out of the schema and do the said checks only in code so as to provide consistent error messages.

What do you think?

tcbuilder/backend/platform.py
```diff
@@ -24,10 +24,14 @@
 
 from tcbuilder.errors import \
     (TorizonCoreBuilderError, InvalidDataError, OperationFailureError,
-     FetchError)
+     FetchError, ParseError, ParseErrors)
 from tcbuilder.backend import ostree, sotaops
 from tcbuilder.backend.bundle import \
     (DindManager, login_to_registries, show_pull_progress_xterm)
+from tcbuilder.backend.build import \
+    (parse_config_file)
```
rborn-tx: In general we break imports in multiple lines and use parentheses only if it doesn't fit into one line - see the import for `ostree` and `sotaops` above.

tcbuilder/backend/platform.py
```diff
@@ -24,10 +24,14 @@
...
+from tcbuilder.cli.build import \
+    (l2_pref)
```
rborn-tx: General rule: the backend should never import from the CLI (since this is a layering violation).

I see this import is here because `push_fuse()` is trying to format messages for displaying purposes which is a task for the CLI layer. If we move the exception handling/error formatting to the CLI layer then this import wouldn't be needed here.

tcbuilder/backend/platform.py
```diff
@@ -60,6 +64,14 @@
 # SHA256 Hash Regex
 HASH_REGEX = re.compile(r"^[0-9a-f]{64}$")
 
+FUSE_HARDWAREIDS = [
+    "verdin-imx8mp-fuses", "verdin-imx8mm-fuses", "apalis-imx8-fuses",
+    "apalis-imx6-fuses", "colibri-imx8x-fuses", "colibri-imx7-emmc-fuses",
+    "colibri-imx6-fuses", "colibri-imx6ull-emmc-fuses"
+]
```
rborn-tx: I'd suggest describing the fusing capabilities of the various machines as data because it's much easier (and less error prone) to change/maintain data than code. With this all the logic to validate the fuse file (in `validate_fuse_file()`) could be made more generic.

For example, we could have:

```
FUSE_TECHNOLOGY_PER_MACHINE = {
    "apalis-imx8x": "ahab",
    "verdin-imx8mm": "hab",...
}
```

We probably don't need the various `verdin-imx8mp-fuses`, `verdin-imx8mm-fuses`, ... names to appear anywhere in the since they follow a formation rule `<machine>-fuses`; instead we could encode the formation rule as code.

tcbuilder/cli/platform.py
```diff
@@ -541,12 +541,9 @@ def do_platform_push(args):
...
     package_info, compatible_with = _check_compatible_with_param(args.compatible_with, credentials)
-    if args.ref.endswith(".yml") or args.ref.endswith(".yaml"):
-        if args.hardwareids and any(hwid != "docker-compose" for hwid in args.hardwareid):
-            raise InvalidArgumentError(
-                "Error: The hardware ID for a docker-compose package can "
-                "only be \"docker-compose\".")
-
+    if ((args.ref.endswith(".yml") or args.ref.endswith(".yaml")) and
+            (not args.hardwareids or
+             (len(args.hardwareids) == 1 and args.hardwareids[0] == "docker-compose"))):
```
rborn-tx: Suggestion for readability and improved UX: move the check of the hwid to a separate function (e.g. `is_compose_hwid(hwids)`) that:

- Returns True if `hwids` is empty.
- Returns True if `len(hwids)>=1` and all ids are equal to `docker-compose`.
- Returns False if `docker-compose` is not in `hwids`.
- Throws an exception otherwise (meaning someone is mixing `docker-compose` IDs with something else).

tcbuilder/cli/platform.py
```diff
@@ -559,6 +556,22 @@ def do_platform_push(args):
             compose_file=args.ref,
             compatible_with=compatible_with,
             canonicalize=args.canonicalize, force=args.force, verbose=args.verbose)
+    elif ((args.ref.endswith(".yml") or args.ref.endswith(".yaml")) and
+          (set(args.hardwareids).issubset(FUSE_HARDWAREIDS))):
```
rborn-tx: Similar suggestion for readability and improved UX: move the check of the hwid to a separate function (e.g. is_fuse_hwid(hwids)) that:

Returns True is all `hwids` end with `-fuses`.

Perhaps we could also already enforce that all IDs are related to the same fuse "technology" (i.e. throw and exception if someone is mixing different technologies).

tcbuilder/cli/platform.py
```diff
@@ -637,13 +653,13 @@ def add_common_push_arguments(subparser):
         required=False, default=[])
     subparser.add_argument(
         metavar="REF", dest="ref",
-        help="OSTree reference or file (docker-compose or generic package) to push to "
-             "Torizon OTA.")
+        help="OSTree reference or file (docker-compose, fuse package, or generic package) "
+             "to push to Torizon OTA.")
```
rborn-tx: Another occurrence of `Torizon OTA` here.

tcbuilder/cli/platform.py
```diff
@@ -637,13 +653,13 @@ def add_common_push_arguments(subparser):
...
     subparser.add_argument(
         "--canonicalize", dest="canonicalize", action=argparse.BooleanOptionalAction,
         help=("Generates a canonicalized version of the docker-compose file, changing "
```
rborn-tx, line 684: Confirmation: this switch would be irrelevant for `fuse` packages (we'd assume it's always true), right?

tcbuilder/backend/fuse.schema.yaml
```diff
@@ -0,0 +1,75 @@
...
+  fuses:
+    type: object
+    description: "Fuse values to be programmed in-order"
+    oneOf:
+      - properties:
+          fuse-close:
+            type: boolean
+            description: "Whether to set closing fuse on device"
+          fuse-val1: 
+            type: number
```
rborn-tx: Maybe it would be less error prone to make this field a string and even check if it starts with "0x".

Imagine the user enters `10` in such field, how would we interpret it? As the decimal number 10 which would be canonicalized as 0xa or as the hex 0x10?

What if the user enters `beef` in such field, how would we interpret it? Testing on a yaml to json converter I see this input is parsed as a string rather than a number. Does this case work with the current code?

tcbuilder/backend/platform.py
```diff
@@ -1439,56 +1451,68 @@ def set_images_hash(compose_file_data):
         svc_spec['image'] = image_parsed.get_name_with_tag()
 
 
-def canonicalize_compose_file(compose_file, force=False):
```
rborn-tx: This comment regards both `canonicalize_compose_file` and `is_canonicalized`.

@jsrc27 As you mentioned in the PR description, we know that the `platform push` command is already too busy and we have intentions to improve this situation on a future major release of TorizonCore Builder.

Looking at your proposed code here I see you're modifying `canonicalize_compose_file` and `is_canonicalized` to make them suitable to handle fuse packages as well; this was done by passing a flag and testing it in multiple places.

I'd propose a different approach of creating new "versions" of those functions specific to the fuse package. There are some of reasons for that:

- The individual functions would be cleaner/easier to read.
- Even though we can say that both file types are in canonical form, compose files and fuse files don't have to follow the same canonicalization rules. Using the same code path increases the chances of changes aiming one type impacting the other unintentionally and makes it harder for us to handle the differences (we end up placing `if`s in multiple places as you can see).
- Keeping the handling of different package types into different functions will make it easier for us to split the platform command into sub-commands.

With this being said, one may argue that my proposal here will cause code duplication which is partially true. However, this can be solved by identifying the relevant parts of the code that are worth reusing (if any) and moving them to reusable functions.

Does all of that make sense to you?

tcbuilder/backend/platform.py
```diff
@@ -1561,5 +1592,153 @@ def restrict_perms(dirname):
...
+                        "--version", version,
+                        "--hardwareids", hardwareids_str,
+                        "--customMeta", json.dumps(custom_metadata)], verbose)
+
+    run_uptane_command(["uptane-sign", "targets", "sign",
+                        "--repo", TUF_REPO_DIR,
+                        "--key-name", "targets"], verbose)
+
+    run_uptane_command(["uptane-sign", "targets", "push",
+                        "--repo", TUF_REPO_DIR], verbose)
```
rborn-tx: This is likely a stretch of code that could be reused between compose and fuse packages because they are both just files from the perspective of the platform.

tcbuilder/backend/platform.py
```diff
@@ -1561,5 +1592,153 @@ def restrict_perms(dirname):
...
+    hab = False
+    if "colibri-imx8x-fuses" in hardwareids or "apalis-imx8-fuses" in hardwareids:
+        ahab = True
+        fuse_num = 16
+
+    if ("verdin-imx8mp-fuses" in hardwareids or "verdin-imx8mm-fuses" in hardwareids or
+            "apalis-imx6-fuses" in hardwareids or "colibri-imx7-emmc-fuses" in hardwareids or
+            "colibri-imx6-fuses" in hardwareids or "colibri-imx6ull-emmc-fuses" in hardwareids):
+        hab = True
+        fuse_num = 8
```
rborn-tx: As a reminder, this could be made more generic/machine independent by describing the "fuse technology" as data.

rborn-tx (changes requested):
Finished 1st pass.

tcbuilder/backend/fuse.schema.yaml
jsrc27: > maybe we could leave the fuse-val{1..8} out of the schema

Did you mean to say leave 9-16 out of the schema? I'm not sure why we would leave 1-8, which are always required, out of the schema.

I also think it would still be helpful to leave all 1-16 in the schema. It would still be a check to make sure all the fields in the provided yaml are valid and listed in the schema (i.e fuse-val17 wouldn't pass the schema)

tcbuilder/backend/platform.py
jsrc27: > We probably don't need the various verdin-imx8mp-fuses, verdin-imx8mm-fuses, ... names to appear anywhere in the since they follow a formation rule <machine>-fuses; instead we could encode the formation rule as code.

I'm not sure I understand this part of the suggestion. Wouldn't we still want the full `<machine>-fuses` somewhere to check against? Only a subset of the Toradex supported machines are compatible with `*-fuses` package type.

tcbuilder/cli/platform.py
jsrc27: Does that really need to be it's own whole function? Seems like it's just wrapping an if check in a function structure. Unless we reuse the same exact check in more than one spot

> Throws an exception otherwise (meaning someone is mixing docker-compose IDs with something else).

Looking again at the previous logic I can see this was the behavior previously. Thinking about it now, is there a reason we don't allow mixing of IDs for docker compose? Shouldn't the only thing that matter is whether at least 1 of the hardwareids is `docker-compose`?

tcbuilder/cli/platform.py
jsrc27: > Returns True is all hwids end with -fuses.

Wouldn't we still need to check each full hardwareid string rather than just checking if it ends in `*-fuses`? Otherwise someone could just do `foo-fuses` or whatever.

tcbuilder/cli/platform.py
jsrc27: Correct, I mean I guess the user could still provide the switch along with a fuse-type package and there wouldn't be any explicit error, or warning. But, it also wouldn't really do anything since we just always implicitly canonicalize for this package type.

As we discussed earlier there's no real practical reason not to just canonicalize always with this package type. Anyways canonicalizing here doesn't cause any significant work, like getting the image hashes does for docker-compose canonicalization.

tcbuilder/backend/fuse.schema.yaml
jsrc27: I feel like we'd have to check more things to be safe, if we also allow strings in this field. For one this would mean users would need to supply the values quoted to make them strings in yaml. Which seems less intuitive to me. I mean yaml can natively support hex numbers, why should I quote a number in a yaml file would be my thought.

> Imagine the user enters 10 in such field, how would we interpret it? As the decimal number 10 which would be canonicalized as 0xa or as the hex 0x10?

Well right now pyyaml will load anything that looks like a number (according to the yaml standard) and translate it as a decimal during the loading process. If a user provides a 10, I mean I guess we'd be making the assumption that they forgot the 0x. Either that or for some reason they provided the fuse values as decimals wanting to translate them to hex via tcb...

> What if the user enters beef in such field, how would we interpret it? Testing on a yaml to json converter I see this input is parsed as a string rather than a number. Does this case work with the current code?

Well right now this would get caught by the current schema since we only allow numbers in these fields.

If there really is a worry about users forgetting the `0x`. I guess we could add another check that parses the yaml file as a "normal" file and makes sure each line of `fuse-val*` has a `0x`.

tcbuilder/backend/platform.py
jsrc27: I'm not sure I see it the same way. The few `if` checks added for the different behavior between compose and fuse yaml files are minimal in my opinion and don't contort the function structure much. The actions needed to canonicalize both types of yaml files are 95% the same logic.

> Even though we can say that both file types are in canonical form, compose files and fuse files don't have to follow the same canonicalization rules

I'm not sure what you mean by different rules here. Don't we really only have 1 rule/method for canonicalizing any yaml file? The only differences would be the handling of specific fields (i.e the image hashes in compose files and the hex values in fuse). And these differences are external to the yaml standard itself. Also the differences are so different it would be difficult to mix them up between the two files.

I did do manual tests where I passed a fuse yaml but told tcb it was docker-compose and vice versa. Both cases failed as expected since the checks for each package type look for fields in the yaml that would only be found in that specific type of yaml.

> Keeping the handling of different package types into different functions will make it easier for us to split the platform command into sub-commands.

As it is now, when/if we split the `platform` command wouldn't we just move these common functions to some common backend module/flle. Then we just keep the specific functions (i.e `restore_hex`) in their own specific files.

tcbuilder/backend/platform.py
jsrc27: I did consider this, but then I thought maybe there was a reason it was like this since `push_generic` and `push_compose` already duplicate these `uptane-sign` executions.

tcbuilder/backend/platform.py
jsrc27: Just to clarify when you say "as data" you mean using a dictionary structure like you showed in one of the above threads?

(i.e):
```
FUSE_TECHNOLOGY_PER_MACHINE = {
    "apalis-imx8x": "ahab",
    "verdin-imx8mm": "hab",...
}
```

tcbuilder/backend/fuse.schema.yaml
rborn-tx: Sorry what I said was confusing :zipper_mouth_face:.

What I meant was leave 1-8 out of the `required` property (that's why I selected that range of lines).

So, the schema validator wouldn't check what `fuse-val<N>` properties are required and instead we would do it in code. For example, if a user specified values {1,2,11,12} with HAB we could show "useful" messages such as:

```
error: fuse values {3,..,8} are missing and must be specified with HAB
error: fuse values {9,..,16} cannot be specified with HAB
```

which the parser alone wouldn't be able to provide.

tcbuilder/backend/fuse.schema.yaml
rborn-tx: I agree that the field being a string is less intuitive but recall that these fields will be fill out by the user by copying data from a file that we provide them. They should never have to think of those fields as numbers and consider in what numeric base they are specified. So, if someone provides a number without the 0x it's more likely that they failed to correctly copy the number from `fuse-cmds.txt`.

Besides that when allowing the fields to be numbers we open possibilities for more errors and inconsistencies. For example:

```
fuse-val1: 100    --> interpreted as decimal 100 -> 0x64
fuse-val1: 0100   --> interpreted as decimal 64  -> 0x40
fuse-val1: beef   --> interpreted as the string "beef" -> parse error
```

With the field being a string we solve all those cases with the small cost (IMHO) of having to put quotes around the values. In fact we would even ensure the numbers are already in the canonical form since we expect them to be (if we assume that users would just copy those values from `fuse-cmds.txt`).

Of course if we don't like the need for quotes we could preprocess the file before feeding into the yaml parser, but the disadvantage is that this would not be something represented in the schema which could cause difficulties for us in the future - say we wanted to use that schema under vscode.

tcbuilder/backend/platform.py
rborn-tx: The gist of the idea is concentrating the information about the fusing capabilities on a single variable. Say we have a machine that we recognize as valid but that doesn't have the fusing capability (e.g. verdin-am62), then we could:

```
FUSE_TECHNOLOGY_PER_MACHINE = {
    "apalis-imx8x": "ahab",
    "verdin-imx8mm": "hab",
    "verdin-am62": None
}
```
The valid hwid names could be easily generated dynamically like this:
```
>>> [f"{mach}-fuses" for mach, tech in FUSE_TECHNOLOGY_PER_MACHINE.items() if tech]
['apalis-imx8x-fuses', 'verdin-imx8mm-fuses']
```

Though it's probably wiser to do the opposite: given a hwid, ensure it ends in `-fuses`, remove the `-fuses` from the end and check if the machine exists in the `FUSE_TECHNOLOGY_PER_MACHINE` dict. If the machine is not found then tell the user we don't know about the machine and if it's found but the technology is `None` then tell them that the machine does not support fusing.

tcbuilder/backend/platform.py
rborn-tx: When we say "canonical composefs file" we mean:

- the yaml file is in the low-level canonical form;
- the file is known to be a Docker compose file and certain fields are interpreted as references to images which are translated into sha-based references.

When we say "canonical fuse file" we mean:

- the yaml file is in the low-level canonical form;
- the file is known to be a fuse package and certain fields are required to be in a specific format that do not match the usual low-level canonicalization.

What is common between them is the low-level canonicalization which is done in the lower part of `canonicalize_file()` but that could be refactored into a separate function and reused by the two cases (compose and fuse).

As for `is_canonicalized()`, honestly there's almost nothing common between the two cases. So I don't see a good reason for trying to handle compose and fuse files with the same function.

> I did do manual tests where I passed a fuse yaml but told tcb it was docker-compose and vice versa. Both cases failed as expected since the checks for each package type look for fields in the yaml that would only be found in that specific type of yaml.

I'm pretty sure it works but here I'm more concerned with the design and long-term maintenance.

Therefore, in terms of design, the handling of the two case with the same functions makes the code harder to read and maintain, at least in my experience. With small refactoring we can properly reuse code without the need for the extra conditionals.

tcbuilder/backend/platform.py
rborn-tx: Hmm, that's something to fix. We probably didn't notice this code duplication when `push_generic` was added.

tcbuilder/backend/platform.py
rborn-tx: Exactly. Here we wouldn't have any machine names appearing in the function implementation.

Neither would we have the number of required fuse values written in code. We'll likely need something like:

```
FUSE_KEY_VALUES_PER_TECHNOLOGY = {
    "hab": 8,
    "ahab": 16
}
```

tcbuilder/cli/platform.py
rborn-tx: I proposed a function because we don't have just the checks to ensure the right condition is handled but also the check to provide a proper error message to guide the user.

tcbuilder/cli/platform.py
rborn-tx: Any additional checks to improve UX are welcome.

The additional check I mentioned of the same technology could cover this case you mentioned: if we don't know the technology it means we don't know how to handle that package.

tcbuilder/cli/platform.py
rborn-tx: Agree.

tcbuilder/cli/platform.py
rborn-tx: > is there a reason we don't allow mixing of IDs for docker compose?

I forgot to answer: I imagine it doesn't make much sense to allow a docker-compose package to be assigned to any secondary other than the docker-compose one (making such assignment would most likely be a operator error in the platform UI, for example).

Of course, in theory we could imagine some other secondary that also processes a compose file but in practice we don't have that case...

tcbuilder/backend/fuse.schema.yaml
jsrc27: I'm still confused on what this exactly would accomplish that the current implementation does not already cover.

So currently the schema requires `fuse-val{1-8}` that's fine we can guarantee the first 8 are present which are always needed. The 2nd eight are listed in the schema but not required. This covers the case if someone adds or mistypes something like `fuse-val17` which would alert them during the schema check.

For the AHAB case to guarantee all 16 fuses values are present, we check this manually in code. As per the schema we know 1-8 are present. We just need need to check 9-16, for this it should be enough to just count how many `fuse-val*` keys were parsed from the file via pyyaml. The schema guarantees any additional keys past the first 8 need to also be listed in the schema. In addition to the fact that keys parsed are unique, if we count 16 `fuse-val*` keys then we know all `fuse-val{1-16}` are present. If not then we already print an error message stating how many keys were expected and how many were found.

From that it should be enough for any user to determine what is missing. If a user gets the message that "9 values were found but 16 were expected" and they specified {1,2.....,9} it should be pretty apparent they just need to continue the fields up to `fuse-val16`.

Or in the example you proposed, they provide only 4 values {1,2,11,12}. Ignoring that the current schema would complain about not having 1-8, they would then get a message about only having 4 values when needing 8. We really shouldn't need to point out that we need 8 values (i.e `fuse-val{1-8,}`) and not something like {1,3,5,7,9,11,13,15}. I mean if you ask somebody to count 8 values and they're counting out of order instead of 1-8, why.

tcbuilder/backend/platform.py
jsrc27: > Though it's probably wiser to do the opposite: given a hwid, ensure it ends in -fuses, remove the -fuses from the end and check if the machine exists in the FUSE_TECHNOLOGY_PER_MACHINE dict. If the machine is not found then tell the user we don't know about the machine and if it's found but the technology is None then tell them that the machine does not support fusing.

Doesn't this unnecessarily block any hardwareids of the form `<something>-fuses`? Why not just have the dict with `<machine>-fuses` like so: 

```
FUSE_TECHNOLOGY_PER_MACHINE = {
    "apalis-imx8x-fuses": "ahab",
    "verdin-imx8mm-fuses": "hab",
}
```

I don't know if we should even block the Toradex supported non-A/HAB machines yet either since nothing has been decided for those yet. It should be the full hardware id string that has a special meaning not just a subset of the string, which is why I was saying we should check the full string rather than just if it ends in `-fuses`.

tcbuilder/cli/platform.py
jsrc27: If the issue is the lack of the exception case, that can be added back similar to how it was before where there wasn't a need for a dedicated function.

* If the input is a yaml and no hwid then it's docker-compose
* If it's yaml and there is a hwid and it's only docker-compose then docker-compose
* If it's yaml and a mix of hwid I guess it goes to the generic package part of the branch statement.
  * We could add the exception throwing and check here if we really don't want to allow a mix of docker-compose with any other hwid

tcbuilder/backend/fuse.schema.yaml
jsrc27: I don't follow the logic here. So we assume the user is copying the values from `fuse-cmds.txt`, which does not have the hex values enclosed by quotes. Or, even if they manually get the list of hex values via `hexdump` this also does not have quotes. We're saying here it's reasonable they know to copy-paste, and then add additional quotation characters that were not there to begin with.

> is that this would not be something represented in the schema which could cause difficulties for us in the future - say we wanted to use that schema under vscode.

I'm not even sure why this is being mentioned as it's not a requirement and out of scope.

tcbuilder/backend/platform.py
jsrc27: Do you just want me to just completely de-couple the fuse package logic from the docker-compose logic? It's increasingly unclear to me what is considered acceptable code reuse. 

If this is going to be a blocker for this PR I rather just de-couple the logic and implement separately, than attempt a refactor on something relatively important (i.e the docker-compose canonicalization logic).

tcbuilder/backend/fuse.schema.yaml
rborn-tx: I've just tested to confirm. E.g. (case 1):

```
fuses:
  fuse-val1: 0x1A
  fuse-val2: 0x2B
  fuse-val3: 0x3C
  fuse-val4: 0x4D
  fuse-val5: 0x5E
  fuse-val6: 0x6F
  fuse-val7: 0x7a
  fuse-close: False
```

gives me the error:

```
=> Parsing errors found:
fuses-missing.yaml: {'fuse-val1': 26, 'fuse-val2': 43, 'fuse-val3': 60, 'fuse-val4': 77, 'fuse-val5': 94, 'fuse-val6': 111, 'fuse-val7': 122, 'fuse-close': False} is not valid under any of the given schemas, while parsing /fuses
```

which is unfriendly and not clear, while (case 2):

```
fuses:
  fuse-val1: 0x1A
  fuse-val2: 0x2B
  fuse-val3: 0x3C
  fuse-val4: 0x4D
  fuse-val5: 0x5E
  fuse-val6: 0x6F
  fuse-val7: 0x7a
  fuse-val8: 0x7a
  fuse-val9: 0x7a
  fuse-close: False
```

throws:

```
Error: Found 9 fuse values in provided yaml file, but expected to find 8 instead.
```

So, as I said initially, we don't get consistent messages if we depend on the parser (or a mix between parser and code). Now, if you follow my other proposal of removing the `oneOf` then probably the parser would produce a more friendly message saying fuse-val8 is missing which would already be an improvement but if the whole error check was done in code then we could get consistent messages in both cases.

So, for example, if we got:

```
(case 1) Error: Found 7 fuse values in provided yaml file, but expected to find 8 instead.
(case 2) Error: Found 9 fuse values in provided yaml file, but expected to find 8 instead.
```

that IMO would already be consistent and clear enough. The other error messages I gave were just examples of what we could do.

Makes sense?

tcbuilder/backend/platform.py
rborn-tx: Not sure if I understand. If we don't know the technology, how can we tell if the provided fuse values are correct? Or are you proposing not to check the fuse values in case the technology is unknown?

In my view, the hardwareid ending in `-fuses` is reserved for Toradex. If a user wanted to create their own fusing secondary for a machine we don't support (yet) they could but they would need to use a hwid ending in something else. So, we wouldn't be blocking other use cases, right? But for the machines/technologies known to us, we make sure all information provided is correct and prevent users from shooting their feet.

tcbuilder/backend/fuse.schema.yaml
rborn-tx: Btw, when testing the cases above I noticed the program took many seconds to run, which seemed odd since it never got to the pushing part...

tcbuilder/backend/fuse.schema.yaml
rborn-tx: Ok, as reviewer I'm trying to think of places where users could make a mistake and how we can minimize the chances of those mistakes going through. This particular point is about the inconsistencies in number parsing in yaml. If for you this is not a problem, I'm okay to close this thread as I imagine/hope such errors to be rare.

tcbuilder/backend/platform.py
rborn-tx: My preference is creating specific versions of `canonicalize_compose_file` and `is_canonicalized` for the fuse package. We're taking about less than 50 lines of code and about 25 statements which is not terrible to duplicate -> modify. The specific versions for the fuse package will be simpler and slightly smaller so the refactoring for reusing code is not strictly needed if you don't feel like doing it.

tcbuilder/cli/platform.py
rborn-tx: That might work as well.

Let me show you some code translating what I wrote so you can evaluate what's best (covering both compose and fuse packages):

```python
    # This can be a local function.
    def _is_compose_hwid(hwids):
        if not hwids:
            return True
        if "docker-compose" in hwids:
            if any(hwid != "docker-compose" for hwid in hwids):
                raise InvalidArgumentError("user is mixing up docker-compose with other things")
            return True
        return False

    # This can be a local function.
    def _is_fuse_hwid(hwids):
        fuses_str = "-fuses"
        if not hwids:
            return False
        if not any(hwid.endswith(fuses_str) for hwid in hwids):
            return False
        if not all(hwid.endswith(fuses_str) for hwid in hwids):
            raise InvalidArgumentError("user is mixing up fuse packages with something else")
        # Ensure all hwids relate to the same technology.
        techs = set()
        for hwid in hwids:
            mach = hwid[:-len(fuses_str)]
            if mach not in FUSE_TECHNOLOGY_PER_MACHINE:
                raise InvalidArgumentError("unknown machine for fuse package...")
            if FUSE_TECHNOLOGY_PER_MACHINE[mach] is None:
                raise InvalidArgumentError("fuse package for machine is not supported...")
            techs.add(FUSE_TECHNOLOGY_PER_MACHINE[mach])
        if len(techs) > 1:
            raise InvalidArgumentError("user is mixing different technologies")
        return True

    package_info, compatible_with = _check_compatible_with_param(args.compatible_with, credentials)
    if ((args.ref.endswith(".yml") or args.ref.endswith(".yaml")) and
        _is_compose_hwid(args.hardwareids)):
        #...

    elif ((args.ref.endswith(".yml") or args.ref.endswith(".yaml")) and
          _is_fuse_hwid(args.hardwareids)):
        #...
    elif os.path.isfile(args.ref):
        #...
    else:
```

tcbuilder/backend/platform.py
lucas-akira: > In my view, the hardwareid ending in -fuses is reserved for Toradex. If a user wanted to create their own fusing secondary for a machine we don't support (yet) they could but they would need to use a hwid ending in something else. So, we wouldn't be blocking other use cases, right? But for the machines/technologies known to us, we make sure all information provided is correct and prevent users from shooting their feet.

I guess the main point of discussion here is whether or not we should reserve harwareids with the `-fuses` suffix. If we do, then we need to make sure this is properly documented to users to avoid confusion.

rborn-tx: That's absolutely a valid discussion.

IMO yes we should reserve the suffix because it adds no real limitations to the user while it avoids clashes with names we may define in the future. Say in the future Toradex starts producing machine `banana` then we'd know that when we define the hwid `banana-fuses` this would never clash with a user-defined hwid.

tcbuilder/backend/fuse.schema.yaml
jsrc27: changed in v2

tcbuilder/backend/platform.py
jsrc27: changed in v2

tcbuilder/backend/platform.py
jsrc27: changed in v2

tcbuilder/cli/platform.py
jsrc27: changed in v2

tcbuilder/backend/platform.py
jsrc27: changed in v2

tcbuilder/backend/platform.py
jsrc27: In V3 I changed the variable to be a dict as suggested. But I'm choosing not to block any hardware-ids that end in `*-fuses`. I don't think it makes sense to wildcard block hardwareid strings, unless the full string is an exact match for a hardwareid we currently use.

It's not like we chose to prohibit hardwareids that end in `*-bootloader`. Also users can upload whatever hardwareid ending in `*-fuses` via the web UI, so this just seems like a moot point.

tcbuilder/cli/platform.py
jsrc27: Added the exception back in V3 so now users can't mix `docker-compose` with other hardwareids.

tcbuilder/cli/platform.py
jsrc27: Closing thread as discussion is covered in another thread.

tcbuilder/backend/platform.py
jsrc27: Completely separated the canonicalization logic in V3.

tcbuilder/backend/platform.py
jsrc27: Using new dict for this check in V3.

rborn-tx: Ok

tcbuilder/backend/platform.py
rborn-tx: Ok

tcbuilder/backend/platform.py
```diff
@@ -1307,39 +1297,13 @@ def push_generic(credentials, target, version, generic_file,
...
-                        "--repo", TUF_REPO_DIR], verbose)
-
-    log.info(f"Successfully pushed {os.path.basename(generic_file)} to OTA server.")
+    uptane_sign_push(credentials=credentials,
+                     push_file=generic_file,
+                     target=target,
+                     version=version,
+                     hardwareids_str=hardwareids_str,
+                     custom_metadata=custom_meta,
+                     verbose=verbose)
```
rborn-tx, line 1306: Since this refactoring is fixing a previous issue unrelated to the fuse implementation, could you make it as a separate commit?

jsrc27: Done

tcbuilder/backend/fuse.schema.yaml
rborn-tx: Checking the program output with v3 I believe they are clear enough now.

Also, regarding the delay I mentioned above, I checked a previous version of TCB and found the same delay is present. So, it's unrelated to the present implementation.

So, closing this thread.

rborn-tx (approved):
All threads solved.

lucas-akira (approved):

PR #48, opened 2025-02-13
tests/integration: Add --ignore-different-secondaries-between-updates to AVAL

lucas-p-bernardes:
A new secondary for fuse was added to Torizon. Because of this, when AVAL tries this update path:

(image with secondary) -> (image withouth secondary) -> (image with secondary)

Aktualizr/Uptane breaks due to an artificial limitation imposed by the platform to prevent security issues. Thus we must always make sure to remove secondaries that are not in the intersection between the images. In the current case, this is only the `fuses` secondary.
This flag allow AVAL to delete /var/sota/storage/fuse path, as a workaround for this.

Related-to: TCB-505
Related-to: TOR-3635


lucas-p-bernardes:
@rborn-tx @lucas-akira Without this change, probably the AVAL pipelines will fail because the updates won't work, be aware of this!

rborn-tx (approved):
LGTM

lucas-akira (approved):
LGTM

PR #51, opened 2025-02-20
tests/integration: Enable "images serve: check zeroconf" test in CI plus minor improvements

lucas-p-bernardes:
Pipeline: https://gitlab.int.toradex.com/lucas.bernardes/torizoncore-builder/-/pipelines/118698

### DIFF
--- a/tests/integration/functions.sh
+++ b/tests/integration/functions.sh
@@ -44,8 +44,9 @@ torizoncore-builder-bg() {
     # Replace the "run" in "docker run ..." with "run -d" to run it in detached mode
     # and to output the ID of the container.
     local CMD=$(eval echo ${TCBCMD/ run / run -d })
+    # Print alias definition if test fail
+    echo "torizoncore-builder-bg alias was defined as: $CMD $@"
     # Run container in the background
-    # echo "# Running $CMD $@"
     run $CMD $@
     assert_success
     # Save the ID of the container.
--- a/tests/integration/functions.sh
+++ b/tests/integration/functions.sh
@@ -37,13 +37,21 @@ export TCB_BG_CONTAINER=""
 
 # run torizoncore-builder-bg
 torizoncore-builder-bg() {
+    # If this function was called by torizoncore-builder-bg-alt_network,
+    # use --network=$TCB_BG_ALT_NETWORK, otherwise use --network=host
+    local tcbcmd
+    if [ -z "${TCBCMD_BG_OVERRIDE}" ]; then
+	    tcbcmd="${TCBCMD}"
+    else
+	    tcbcmd="${TCBCMD_BG_OVERRIDE}"
+    fi
     # Make sure no other process is running in the background
     assert [ -z $TCB_BG_CONTAINER ]
     # Make sure the docker command does not have the "-it" parameter(s)
     [[ $TCBCMD == *" -it"* ]] && assert false
     # Replace the "run" in "docker run ..." with "run -d" to run it in detached mode
     # and to output the ID of the container.
-    local CMD=$(eval echo ${TCBCMD/ run / run -d })
+    local CMD=$(eval echo ${tcbcmd/ run / run -d })
     # Print alias definition if test fail
     echo "torizoncore-builder-bg alias was defined as: $CMD $@"
     # Run container in the background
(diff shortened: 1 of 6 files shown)


lucas-akira (approved):

lucas-p-bernardes:
Hello @rborn-tx @jsrc27, can we have this merged?

jsrc27 (approved):

tests/integration/functions.sh
```diff
@@ -42,10 +42,20 @@ torizoncore-builder-bg() {
...
+    if [ -n "$TCBCMD_NETWORK" ]; then
+        local CMD=$(eval echo ${TCBCMD_NETWORK/ run / run -d })
+    else
+        local CMD=$(eval echo ${TCBCMD/ run / run -d })
+    fi
+    # If other function call torizoncore-builder-bg, the default TCBCMD will be used
+    unset TCBCMD_NETWORK
+
+    # Print alias definition if test fail
+    echo "torizoncore-builder-bg alias was defined as: $CMD $@"
```
rborn-tx: This `torizoncore-builder-bg` is a generic function that can be used in many tests and this new stretch of code looks too specific IMO. A better design would be to make it even more generic and keep the specific parts somewhere else.

Here's a way to do that:

```
torizoncore-builder-bg() {
    local tcbcmd
    if [ -z "${TCBCMD_BG_OVERRIDE}" ]; then
	tcbcmd="${TCBCMD}"
    else
	tcbcmd="${TCBCMD_BG_OVERRIDE}"
    fi
    # And replace TCBCMD -> tcbcmd in the rest of the function
    ...
}
```

The other function that starts TCB with a specific network would then do something like this:

```
torizoncore-builder-bg-tcb_network() {
    local tcbcmd="${TCBCMD/--network=host/--network=tcb_network}"
    TCBCMD_BG_OVERRIDE="${tcbcmd}" torizoncore-builder-bg "$@"
}
export -f torizoncore-builder-bg-tcb_network
```

Notice we don't need any global variable in this version (the value of `${TCBCMD_BG_OVERRIDE}` is passed only to `torizoncore-builder-bg`) which is another improvement.

---

As a second point, notice that `torizoncore-builder-bg-tcb_network` is coupled with `tcb_create_docker_network` which is in another module (`setup.sh`) and the connection between the two is not visible. If someone decides to change the name of the network, things would simply stop working and one would need to figure out why. It would be better to keep the name of the network in some variable exported by `setup.sh` say `TCB_BG_ALT_NETWORK` and use that variable in both functions.

In this case, I'd rename `torizoncore-builder-bg-tcb_network` to say `torizoncore-builder-bg-alt-network` to indicate that this new function will start TCB on the background on an "alternative" network.

tests/integration/testcases/images-serve.bats
```diff
@@ -24,12 +24,17 @@ teardown() {
...
 @test "images serve: check zeroconf tezi service response." {
-    skip-under-ci
-
     IMAGE_DIR="samples/images"
-    torizoncore-builder-bg images serve $IMAGE_DIR
-
-    run avahi-browse-domains -a -t
+    torizoncore-builder-bg-tcb_network images serve $IMAGE_DIR
+
+    run docker run -i --rm --privileged --network=tcb_network alpine:latest sh -c "
```
rborn-tx: Here again we'd use `${TCB_BG_ALT_NETWORK}`.

rborn-tx (changes requested):
Sorry for the delay - I overlooked this PR. I have some suggestions for better design/code reuse.

tests/integration/functions.sh
lucas-p-bernardes: Done!

tests/integration/testcases/images-serve.bats
lucas-p-bernardes: Done!

lucas-p-bernardes:
Hello @rborn-tx, I fixed your points, and rebase the branch. Please take a look if everything was fixed correctly and merge if it's OK.
This is the pipeline with this last changes: https://gitlab.int.toradex.com/lucas.bernardes/torizoncore-builder/-/jobs/937051

lucas-p-bernardes:
@rborn-tx The tests start failing in Github (and only in Github) when I defined the name of the docker network using the variable `TCB_BG_ALT_NETWORK`, defined in the `tcb_create_docker_network()` script in `setup.sh`, because the variable  is not persisting across steps, since each `run:` command in GitHub Actions is executed in a new shell, meaning that exported variables do not persist between steps.
Do you have a suggestion for fix this?

rborn-tx:
> @rborn-tx The tests start failing in Github (and only in Github) when I defined the name of the docker network using the variable `TCB_BG_ALT_NETWORK`, defined in the `tcb_create_docker_network()` script in `setup.sh`, because the variable is not persisting across steps, since each `run:` command in GitHub Actions is executed in a new shell, meaning that exported variables do not persist between steps. Do you have a suggestion for fix this?

Looking at the other steps I see the environment variables are passed between steps by appending to a file referenced by variable `GITHUB_ENV`. Then the step that does the setup would become:

```
      - name: Run setup script
        run: |
          TCB_SKIP_PULL=1 ./setup.sh
          echo "TCB_BG_ALT_NETWORK=${TCB_BG_ALT_NETWORK}" >> ${GITHUB_ENV}
```

In the future we could think of something more elaborate such as asking the setup script itself to dump the exported variables.

lucas-p-bernardes:
@rborn-tx The bug of Github Actions failing was solved!

rborn-tx (approved):
LGTM

PR #52, opened 2025-02-24
aval-tests: Change the workaround for mismatch between database and OSTree

lucas-p-bernardes:
The previous workaround was generating a problematic edge case. Let’s suppose the existence of two parallel tests, A and B. With this previous workaround, test A runs the TCB tests, releases the device, and then takes it again to update it to a different version, aiming to sync the Aktualizr’s database with OSTree. However, if test B takes the device after A’s TCB tests but before A’s workaround update and attempts an update to the Torizon version before A’s TCB tests, the device will fall into the same mismatch bug.

Because of this, the workaround was changed to delete the Aktualizr’s database, which forces Aktualizr to recreate it, this time based on OSTree, bringing both into sync again. With this new workaround, the device is not released until the workaround is completed, there is no risk of encountering these edge cases, and it also speeds up testing by saving one update.

Related-to: TOR-3518
Related-to: AVAL-53
Related-to: TCB-490

### DIFF
--- a/tests/integration/aval/aval-tests.yml
+++ b/tests/integration/aval/aval-tests.yml
@@ -55,25 +55,11 @@ build-aval-docker:
     - ${T} docker login -u "${CI_REGISTRY_USER}" -p "${CI_REGISTRY_PASSWORD}" "${CI_REGISTRY}"
     - ${T} docker pull "${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}"
     - echo -e "\e[0Ksection_end:$(date +%s):pull_eval_tcb_section\r\e[0K"
+    # To circumvent the mismatch between Aktualizr's database and ostree when an external entity manages ostree without
+    # Aktualizr, delete the database files, which forces Aktualizr to create database again, this time synced with ostree
     - cd tests/integration && python /aval/main.py --delegation-config $DELEGATION_CONFIG --ignore-different-secondaries-between-updates
       --run-before-on-host "./run_all.sh --device --machine $TCB_MACHINE --report --tcb-tags requires-device
-      --tcb-custom-image ${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}"
-      # To circumvent the mismatch between Aktualizr's database and ostree when an external entity manages ostree without
-      # Aktualizr, install a different version than the one Aktualizr thinks it's currently installed (the one before TCB
-      # deploy), which externally forces both Aktualizr and ostree to sync. In particular, if we're running a nightly, install
-      # a release build and vice-versa
-    - |
-      toggle_release_type() {
-        local current_release_type="$1"
-        if [ "$current_release_type" == "release" ]; then
-          echo "nightly"
-        elif [ "$current_release_type" == "nightly" ]; then
-          echo "release"
-        fi
-      }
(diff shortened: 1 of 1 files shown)


lucas-p-bernardes:
Pipeline: https://gitlab.int.toradex.com/lucas.bernardes/torizoncore-builder/-/pipelines/119239

jsrc27 (approved):

lucas-p-bernardes:
Hello @lucas-akira @rborn-tx. Can you take a look at this MR first? This is causing issues with some boards stuck in "Update in Progress", what is very annoying for maintenance.

lucas-akira (approved):

tests/integration/aval/aval-tests.yml
```diff
@@ -55,25 +55,11 @@ build-aval-docker:
...
-        if [ "$current_release_type" == "release" ]; then
-          echo "nightly"
-        elif [ "$current_release_type" == "nightly" ]; then
-          echo "release"
-        fi
-      }
-    - TARGET_BUILD_TYPE=$(toggle_release_type "$TARGET_BUILD_TYPE") python /aval/main.py --delegation-config $DELEGATION_CONFIG --ignore-different-secondaries-between-updates
-      "echo 'Updating to a different TOS version than the TCB tests version, to circumvent the mismatch between Aktualizr'\''s database and ostree when an external entity manages ostree without Aktualizr'"
-    - (! grep "^not ok" workdir/reports/*)
+      --tcb-custom-image ${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}" "echo $DEVICE_PASSWORD | sudo -S rm -rf /var/sota/sql.db /var/sota/storage/*"
```
rborn-tx, line 62: We could try this approach but honestly I'm not sure this would totally solve the issue. If we just delete the primary database, wouldn't aktualizr still use information from the other databases? And with this possibly try to re-register an ECU with the same id, which the platform would disallow?

rborn-tx (approved):
As I said the other day, the reliable solution would be to delete the whole `/var/sota` directory and then reprovision the device so it looks like a new device on the platform side.

Anyways, I'll approve this since this does not impact customers, but I would seriously consider a more reliable solution.

lucas-p-bernardes:
> As I said the other day, the reliable solution would be to delete the whole `/var/sota` directory and then reprovision the device so it looks like a new device on the platform side.
> 
> Anyways, I'll approve this since this does not impact customers, but I would seriously consider a more reliable solution.

I agree, I plan to create this more robust solution in the AVAL repo itself, but this will require a bit more effort. For now, I believe this solves this specific problem.

tests/integration/aval/aval-tests.yml
lucas-p-bernardes: The original problem was that the sql.db and OSTree wasn't synced. I'm deleting sql.db expecting they synced again, so I'm expecting Aktualizr use information from other databases.

> And with this possibly try to re-register an ECU with the same id, which the platform would disallow?

About this, I thought Aktualizr just recreate the local sql.db using OSTree information, and not re-register the device.

rborn-tx: > The original problem was that the OSTree and libostree wasn't synced.

`OSTree` == `libostree`, so I don't think I understand what you mean.

> About this, I thought Aktualizr just recreate the local sql.db using libostree information, and not re-register the device.

The database has many other pieces of information besides the currently deployed OSTree commit and would all be lost. For example, how would Aktualizr know what ECUs are already registered with the platform? This information is in the database, thus I'd expect it to retry and register the ECUs again. Notice this is something the platform would allow -- it doesn't allow the re-registration of an ECU that was previously de-registered. Also, notice I'm saying ECU, not "device".

lucas-p-bernardes: > OSTree == libostree, so I don't think I understand what you mean.

Sorry, I confused the names, I updated the comment.

PR #54, opened 2025-02-25
aval-tests: Don't run AVAL tests when the source is a trigger token

lucas-p-bernardes:
A side effect of commit c27a999c48e6b8625b18cc069eb608799f36b309, which removes the rule $CI_PIPELINE_SOURCE == "schedule", is that AVAL tests started running by default in all created pipelines, including the pipeline triggered by Jenkins.

However, it does not make sense to test AVAL with this pipeline created by Jenkins in the Torizon nightlies, since it runs TCB tests using the nightly created by Jenkins as the base image for the tests. On the other hand, AVAL runs tests using the release or the nightly available in the Cloud, which is not necessarily the same as the one created by Jenkins, making these tests unnecessary and redundant with those already performed when a commit is created or with the Early-Access pipeline.

To solve this, a rule was added to prevent AVAL tests from running when the job source is a trigger token.

### DIFF
--- a/tests/integration/aval/aval-tests.yml
+++ b/tests/integration/aval/aval-tests.yml
@@ -25,6 +25,8 @@ build-aval-docker:
       allow_failure: true
     - if: '$AVAL_DISABLED =~ /^true$/i'
       when: never
+    - if: $CI_PIPELINE_SOURCE == "trigger"
+      when: never
     - when: on_success
   variables:
     IMAGE_NAME_AVAL: aval-docker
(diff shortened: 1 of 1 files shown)


lucas-p-bernardes:
Pipeline triggered by a new commit: https://gitlab.int.toradex.com/lucas.bernardes/torizoncore-builder/-/pipelines/119492
Pipeline triggered using a trigger token: https://gitlab.int.toradex.com/lucas.bernardes/torizoncore-builder/-/pipelines/119493

Note the AVAL jobs doesn't exist in the second.

jsrc27 (approved):

lucas-akira (approved):

lucas-p-bernardes:
Hello @rborn-tx @jsrc27, can we have this merged?

PR #57, opened 2025-03-22
Add new ubootenv fuses command

jsrc27:
Add new command to allow users to provision fusing data to their OS image's U-Boot environment.

Also add basic tests to for this new command.


tcbuilder/backend/ubootenv.py
```diff
@@ -0,0 +1,182 @@
...
+        log.warning("Warning: Unable to determine image version in input directory. "
+                    "Proceeding anyways.")
+    elif (major == 7 and minor < 2) or major < 7:
+        log.warning("Warning: Provisioning secure fuse data is not supported on Torizon "
+                    "OS before 7.2.0. Proceeding anyways, but this will most likely not "
+                    "do anything.")
+
+    initial_env = subprocess.check_output(
+        ["find", input_dir, "-type", "f",
+         "-name", "u-boot-initial-env*", "-print", "-quit"], text=True)
```
rborn-tx: Shouldn't the name of the file be taken from `image.json` (field `u_boot_env`)?

tcbuilder/backend/ubootenv.py
```diff
@@ -0,0 +1,182 @@
...
+    # Check to see if initial u-boot env already contains fuse variables
+    if any(fuse_var in env for fuse_var in FUSE_VARIABLES):
+        raise InvalidStateError(
+            "Input image already contains fuse related u-boot variables")
+
+    board = find_board(env)
+    hwid = [board + "-fuses"]
+    if not set(hwid).issubset(set(FUSE_HARDWAREIDS.keys())):
+        raise InvalidStateError(
+            f"Hardware type \"{board}\" is not supported by this command.")
```
rborn-tx: Can you foresee `find_board` ever returning a list of boards? If not, this could be made simpler:

```python
hwid = board + "-fuses"
if hwid not in FUSE_HARDWAREIDS:
    ...
```

tcbuilder/backend/ubootenv.py
```diff
@@ -0,0 +1,182 @@
...
+
+    # Validate and check fuse file before proceeding
+    validate_fuse_file(fuse_file, hwid)
```
rborn-tx: Notice here you could write `[hwid]` if hwid was a scalar type.

tcbuilder/backend/ubootenv.py
```diff
@@ -0,0 +1,182 @@
...
+        raise TorizonCoreBuilderError(
+            f"Error: '{fuse_file}' is not in canonical form, which is expected "
+            "for files with the '.lock' extension.")
+
+    if not is_lockfile:
+        fuse_file_canon = canonicalize_fuse_file(fuse_file, force)
+    else:
+        log.debug(f"Input file '{fuse_file}' is already in canonical form, "
+                  "proceeding.")
+        fuse_file_canon = fuse_file
```
rborn-tx: Since the file is not going to be pushed to the platform, what's the point in canonicalizing it or checking if it's in canonical form?

tcbuilder/backend/ubootenv.py
```diff
@@ -0,0 +1,182 @@
...
+    else:
+        if os.path.exists(output_dir) and force:
+            shutil.rmtree(output_dir)
+
+        log.debug("Creating copy of Torizon OS input image.")
+        shutil.copytree(input_dir, output_dir)
+        output_env = subprocess.check_output(
+            ["find", output_dir, "-type", "f",
+             "-name", "u-boot-initial-env*", "-print", "-quit"], text=True)
+        output_env = output_env.rstrip()
```
rborn-tx: If we took the env file name from `image.json` then this variable would be set to `os.path.join(output_directory, <env-file-name>)`.

tcbuilder/backend/ubootenv.py
```diff
@@ -0,0 +1,182 @@
...
+        log.debug("Updating Torizon OS image in place.")
+        output_dir = input_dir
+        output_env = initial_env
+        inplace = True
+    else:
+        if os.path.exists(output_dir) and force:
+            shutil.rmtree(output_dir)
+
+        log.debug("Creating copy of Torizon OS input image.")
+        shutil.copytree(input_dir, output_dir)
```
rborn-tx, line 76: Notice that with this `copytree` call here, if the process is terminated (by any signal) the output directory would be left behind.

tcbuilder/backend/ubootenv.py
```diff
@@ -0,0 +1,182 @@
...
+    close = ""
+    if close_bool == "true":
+        close = "1"
+    elif close_bool == "false":
+        close = "0"
+
+    var_list = [
+        "fuse_prog_list=" + val_list, "fuse_prog_close=" + close, "fuse_status=pending"
+    ]
+    return var_list
```
rborn-tx: Shouldn't we use the yaml parser here rather than parse the file as plain text?

I see the intention here was to load the canonical file, but as I said on a previous question, I'm not sure why we want the canonical fuse file to generate the environment. If we didn't then the file could even be loaded only once (in `validate_fuse_file`).

tcbuilder/backend/ubootenv.py
```diff
@@ -0,0 +1,182 @@
...
+            shutil.rmtree(output_dir)
+        raise TorizonCoreBuilderError(
+            "Failed to update u-boot-initial-env* file.")
+# pylint: enable=too-many-locals
+
+
+def find_board(env):
+    """Determine the hardware type via the u-boot-initial-env* file
+
+    :param env: File contents of u-boot-initia-env*
```
rborn-tx: initia -> initial

rborn-tx (commented):
UX looks good.

I've raised a couple of questions mostly about the design.

tcbuilder/backend/ubootenv.py
jsrc27: Yeah that's a good point. I was focused on the `u-boot-initial-env*` file so I didn't even look at what was in `image.json`.

tcbuilder/backend/ubootenv.py
jsrc27: Hmm probably not. The `board` U-Boot variable is controlled by `CONFIG_SYS_BOARD` in U-Boot I believe. So someone would  have to change this in a custom U-Boot for it to return something unexpected like a list of boards. Such a scenario is probably very highly unlikely.

tcbuilder/backend/ubootenv.py
jsrc27: I'm confused, isn't `hwid` already being changed to a type list several lines above?

Or wait, are you suggesting I do this change, if I also do the change in the thread above?

tcbuilder/backend/ubootenv.py
jsrc27: Hmm I guess it's not strictly needed. My thought here was to kind of "enforce" the utilization of the canonical form for the fuse yaml file. Since there's not really any benefit or point, for users to not have it be canonicalized. Kinda like a promote best practices kind of thing. 

Also wanted to ensure the fuse hex values being written to the u-boot-initial-env file were consistent. In the sense that any letter characters in the hex value was lower-case, since that's what we've been going for. Again it's not strictly required the hex values have lower case letter characters, but I was thinking more along the lines of consistency.

I guess we can just keep the canonical yaml data internal to TCB and not create the canonical file for the user? For example load the input via the yaml parser, restore the hex values, force everything to lower-case. Which is essentially what the canonicalization process already does, but then just work with the yaml data and not output it to the user.

Would that be a better idea?

tcbuilder/backend/ubootenv.py
jsrc27: Isn't this the same as what is done for `combine` and `images provision`?

tcbuilder/backend/ubootenv.py
jsrc27: > If we didn't then the file could even be loaded only once (in validate_fuse_file).

Are you suggesting to change `validate_fuse_file` to return the parsed yaml data that it was validating?

That could work, would have to remember to restore the numbers back to hex form, but there's already a function for that. 

Between this and the other thread, I guess we more or less are working with the canonical form of the file, internal to the tool. We just don't create the lock file then?

tcbuilder/backend/ubootenv.py
rborn-tx: Exactly, it's just a heads up if you took my previous suggestion - in that case, the variable would become scalar but here we'd need to pass a list due to the expectation of `validate_fuse_file`.

tcbuilder/backend/ubootenv.py
rborn-tx: IMO, yes. I mean, would prefer not force users into using the canonical form if that is not required. For pushing to the platform, the canonicalization was required for proper functionality (and that functionality would be at some point visible to the user), but here just ensuring the file is valid seems good enough and we avoid the generation of an extra output for the command.

If for us is important or better to handle the contents in canonical format then we could keep that internal to the tool which could be achieve in the way you mentioned.

tcbuilder/backend/ubootenv.py
rborn-tx: Good point. Guess I never paid attention to that :weary:. Never mind then.

tcbuilder/backend/ubootenv.py
rborn-tx: Ideally we would need to load the file only once and that's why I mentioned `validate_fuse_file()` which is the first place the file is loaded and parsed.

Since in the end we only care about the fuse values, the most straightforward path seems to be:

- Make `validate_fuse_file()` return the parsed yaml file.
- Pass the yaml to `get_fuse_vars` which would grab the hex values from there.
- Canonicalize the hex values when generating the output list of `get_fuse_vars()`.

And then we don't even need a canonical version of the whole file even internally to the tool. Wouldn't that work?

tcbuilder/backend/ubootenv.py
```diff
@@ -0,0 +1,182 @@
...
+        fuse_file_canon = fuse_file
+
+    var_list = get_fuse_vars(fuse_file_canon)
+
+    inplace = False
+    if output_dir is None:
+        log.debug("Updating Torizon OS image in place.")
+        output_dir = input_dir
+        output_env = initial_env
+        inplace = True
```
lucas-akira: In the `combine` command, updating an image in place requires either the input directory being the same as the output one, or no output directory being specified:

https://github.com/torizon/torizoncore-builder/blob/3.12.0/tcbuilder/backend/combine.py#L227

But for either case the `--force` flag is always required. Maybe we should have the same behavior here.

tcbuilder/backend/ubootenv.py
jsrc27: Done in V2

tcbuilder/backend/ubootenv.py
jsrc27: Done in V2

tcbuilder/backend/ubootenv.py
jsrc27: Done in V2

tcbuilder/backend/ubootenv.py
jsrc27: Adjusted in V2, a lock file is no longer outputted and we keep the yaml data internal for parsing.

tcbuilder/backend/ubootenv.py
jsrc27: Done in V2

tcbuilder/backend/ubootenv.py
jsrc27: Done in V2

tcbuilder/backend/ubootenv.py
jsrc27: Done in V2

tcbuilder/backend/ubootenv.py
jsrc27: So as I was  doing this, things got a bit "weird".

First of all turns out I did not need to ad an explicit check here since in `cli/ubootenv.py`, there is already a check for if the output directory exists and `--force` is not passed. This would also check the in-place operation if the output directory that was passed is also the input (i.e in-place).

Anyways the part that got weird was when I was checking the case where no output directory is specified, therefore implicitly the tool should assume a in-place operation. Except when I do that I get:
```
usage: torizoncore-builder ubootenv fuses [-h] --fuse-file FUSE_FILE [--force] INPUT_DIRECTORY OUTPUT_DIRECTORY
torizoncore-builder ubootenv fuses: error: the following arguments are required: OUTPUT_DIRECTORY
```

Which confused me, but maybe I just forgot to test this case the first time. I then checked `combine` and `images provision` that also should have in-place support for when the output directory is not specified. But, if you don't pass the output directory you get the same issue:
```
$ torizoncore-builder combine torizon-docker-verdin-imx8mp-Tezi_7.1.0+build.4/
usage: torizoncore-builder combine [-h] [--bundle-directory BUNDLE_DIRECTORY] [--force] [--raw-rootfs-label LABEL] [--image-name IMAGE_NAME] [--image-description IMAGE_DESCRIPTION] [--image-licence LICENCE_FILE] [--image-accept-licence | --no-image-accept-licence] [--image-release-notes RELEASE_NOTES_FILE]
                                   [--image-autoinstall | --no-image-autoinstall] [--image-autoreboot | --no-image-autoreboot]
                                   IMAGE_PATH OUTPUT_PATH
torizoncore-builder combine: error: the following arguments are required: OUTPUT_PATH
$ torizoncore-builder images provision --mode offline torizon-docker-verdin-imx8mp-Tezi_7.1.0+build.4/
usage: torizoncore-builder images provision [-h] --mode {offline,online} [--force] [--shared-data SHARED_DATA_FILE] [--online-data ONLINE_DATA] [--hibernated] INPUT_DIRECTORY OUTPUT_DIRECTORY
torizoncore-builder images provision: error: the following arguments are required: OUTPUT_DIRECTORY
```

So now I'm questioning whether this in-place stuff ever worked in the scenario where you don't pass the output directory. In both `combine` and `images provision` there is explicit logic to handle the case where output directory was not passed, but based on this it seems like this logic would never actually be executed.

In any case, for now I've just made it so that the way the in-place operation works here, is when the user passes an input and output directory, and they refer to the same directory.

CC: @rborn-tx

jsrc27: Oh wait never mind on further inspection the in-place operation where output directory is not specified is only used for the `build` command, not the standalone commands.

That's why I got confused since this is only implementing the standalone command portion.

Well I guess for now it's sensible to have the only in-place operation be when the user uses the standalone command and passes the same directory as input and output. When the integration for the `build` command gets added, then it would make sense to add the ability for the backend of this command to handle the case where the output directory is none.

lucas-akira: > Oh wait never mind on further inspection the in-place operation where output directory is not specified is only used for the build command, not the standalone commands.

Right, I recall now that for the `combine` command it's only possible to use the "no output directory" in-place operation with the `build` command.

> Well I guess for now it's sensible to have the only in-place operation be when the user uses the standalone command and passes the same directory as input and output.

Looking at the code again, the "input equal to the output directory" in-place operation can already be done by the current implementation as you said, so my initial suggestion didn't make much sense. 

One minor thing I would suggest is having the `Updating Torizon OS image in place.` message showing up when this operation happens, but I see you already did it.

> When the integration for the build command gets added, then it would make sense to add the ability for the backend of this command to handle the case where the output directory is none.

Yeah, no problem. All good then!

tcbuilder/backend/ubootenv.py
```diff
@@ -0,0 +1,175 @@
...
+    """
+
+    image_json_path = os.path.join(tezi_dir, "image.json")
+    if not os.path.exists(image_json_path):
+        raise PathNotExistError(f"File {image_json_path} does not exist.")
+
+    with open(image_json_path, 'r') as file:
+        jsondata = json.load(file)
+
+    env_file = jsondata["u_boot_env"]
```
rborn-tx: You could use a `get()` here to avoid an exception if the field did not exist; without it the `if` below would be reached only if the field was present but empty.

tcbuilder/backend/ubootenv.py
```diff
@@ -0,0 +1,175 @@
...
+    if not env_file:
+        raise FileContentMissing(
+            f"Could not find u_boot_env feild in {image_json_path}")
```
rborn-tx: typo

rborn-tx (commented):
I've finished a second pass with minor findings only. Apart from those, everything else looks good to me.

rborn-tx:
Somehow GitHub managed to duplicate by comments :thinking:

tcbuilder/backend/ubootenv.py
jsrc27: Done in V3

tcbuilder/backend/ubootenv.py
jsrc27: Done in V3

jsrc27:
By the way guys, will this be merged after the 3.13 release? I don't have preference either way, I was just wondering since I know we are in the middle of the release process.

lucas-akira:
We haven't started the tests for the 3.13 release yet, as we need to merge #58 to have the test pipeline up and running. 

If this PR gets merged before that, then it will be part of the 3.13 release.

rborn-tx (approved):
V3 LGTM.

lucas-akira (approved):

PR #59, opened 2025-04-02
Fix one 'platform push' test, add 'ubootenv' tests to the CI pipeline, plus other improvements

lucas-akira:
See commit messages for more details.

Related-to: TCB-508

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -179,20 +179,22 @@ build-arm64:
     - ${T} docker login -u "${CI_REGISTRY_USER}" -p "${CI_REGISTRY_PASSWORD}" "${CI_REGISTRY}"
     - ${T} docker pull "${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}"
     - echo -e "\e[0Ksection_end:$(date +%s):pull_eval_tcb_section\r\e[0K"
-  script:
-    - ./run_all.sh --machine $TCB_MACHINE --report --testcase "$TCB_TESTCASE"
-      --tcb-custom-image ${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}
+    # get all .bats files (excluding raw/wic ones) except platform.bats
+    - export TCB_HOST_TESTS="torizoncore-builder $(cd testcases/ && ls -1 *.bats | sed -e 's/\.bats//' -e '/^\(platform\|torizoncore-builder\)$/d' | sort)"
+  after_script:
     # check results
-    - (! grep "^not ok" workdir/reports/*)
+    - (! grep "^not ok" tests/integration/workdir/reports/*)
 
 test-host-tc6-platform:
   extends: .test-base
   stage: test
   variables:
-    TCB_TESTCASE: platform
     YOCTO_BRANCH: "kirkstone-6.x.y"
     TARGET_BUILD_TYPE: "release"
     TCB_MACHINE: "apalis-imx8"
+  script:
+    - ./run_all.sh --machine $TCB_MACHINE --report --testcase platform
+      --tcb-custom-image ${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}
 
 test-host-tos7-platform:
   needs:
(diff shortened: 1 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -179,11 +179,14 @@ build-arm64:
     - ${T} docker login -u "${CI_REGISTRY_USER}" -p "${CI_REGISTRY_PASSWORD}" "${CI_REGISTRY}"
     - ${T} docker pull "${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}"
     - echo -e "\e[0Ksection_end:$(date +%s):pull_eval_tcb_section\r\e[0K"
+    # get all .bats files (excluding raw/wic ones) except platform.bats
+    - export TCB_HOST_TESTS="torizoncore-builder $(find ./testcases -maxdepth 1 -type f -name "*.bats" | sed -E "s#\./testcases/(.*).bats#\1#" | grep -v torizoncore-builder | grep -v platform | sort)"
```
rborn-tx: Any particular reason for always putting test `torizoncore-builder` at the beginning of the list? Our tests should be idempotent so I don't think this should be necessary...

As a suggestion, the assignment could be simplified to something like:

```
export TCB_HOST_TESTS="$(cd testcases/ && ls -1 *.bats | sed -e 's/\.bats//' -e '/^platform$/d' | sort)"
```

And notice with this approach we could more easily add tests to be excluded by replacing `platform` above by `\(platform\|test1\|test2\|...\)`.

Anyways, I wish the could achieve the same goal by using the test tags...

rborn-tx (approved):
LGTM. I've added a suggestion for an optional improvement.

.gitlab-ci.yml
lucas-akira: > Any particular reason for always putting test torizoncore-builder at the beginning of the list? Our tests should be idempotent so I don't think this should be necessary...

I put it at the beginning mostly because it corresponds to the first Python file that gets executed.

If it fails then probably something is very wrong with the core application itself, so I think it would be useful to have it running first so we know right off the bat if TCB is working at a very basic level.

> As a suggestion, the assignment could be simplified to something like:
> And notice with this approach we could more easily add tests to be excluded by replacing platform above by \(platform\|test1\|test2\|...\).

That's a nice suggestion. I'll apply it, thanks!

jsrc27 (approved):

.gitlab-ci.yml
```diff
@@ -179,11 +179,14 @@ build-arm64:
...
   script:
     - ./run_all.sh --machine $TCB_MACHINE --report --testcase "$TCB_TESTCASE"
       --tcb-custom-image ${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}
```
lucas-p-bernardes: Since you are already overwritting the `script` in `test-host-tc6-32bit`, `test-host-tc6-64bit`, `test-host-tos7-32bit` and `test-host-tos7-64bit`, sounds more reasonable for orgatization to remove the `script` in `.test-base` and also define in `test-host-tc6-platform` and `test-host-tos7-platform`.
This keeps the code more standardized, avoiding unnecessary overwriting, but it is just a suggestion, feel free to do what you think it's better.

lucas-akira: Right, makes sense. I've followed your suggestion, thanks!

.gitlab-ci.yml
```diff
@@ -179,67 +179,75 @@ build-arm64:
     - ${T} docker login -u "${CI_REGISTRY_USER}" -p "${CI_REGISTRY_PASSWORD}" "${CI_REGISTRY}"
     - ${T} docker pull "${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}"
     - echo -e "\e[0Ksection_end:$(date +%s):pull_eval_tcb_section\r\e[0K"
-  script:
-    - ./run_all.sh --machine $TCB_MACHINE --report --testcase "$TCB_TESTCASE"
-      --tcb-custom-image ${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}
+    # get all .bats files (excluding raw/wic ones) except platform.bats
+    - export TCB_HOST_TESTS="torizoncore-builder $(cd testcases/ && ls -1 *.bats | sed -e 's/\.bats//' -e '/^\(platform\|torizoncore-builder\|\)$/d' | sort)"
```
rborn-tx:
```
\(platform\|torizoncore-builder\|\)
                               ^^
```

Is the empty OR intended here?

lucas-akira: Oops, definitely not intentional. I've removed it, thanks!

PR #61, opened 2025-04-10
Maintenance in CI pipelines

lucas-p-bernardes:


tests/integration/run_all.sh
```diff
@@ -126,23 +126,37 @@ if [ $? -ne 0 ]; then
...
+            exit 1
+        fi
     fi
+fi
 
-    if [ -f "./get_raw_images.sh" ]; then
-        echo "Downloading raw images..."
-        ./get_raw_images.sh
-    fi
+if [ -n "${IS_WIC}" ]; then
```
lucas-akira: As it is right now if `IS_WIC` equals the string `false` it will still enter this `if` statement, given that the variable length is non-zero.

I suggest replacing this condition with `if [ "${IS_WIC}" = "true" ]` instead.

tests/integration/run_all.sh
```diff
@@ -126,23 +126,37 @@ if [ $? -ne 0 ]; then
     exit 1
 fi
 
-if [ ! -e workdir/images/.images_downloaded ] || [ ! -e workdir/images/.raw_images_downloaded ]; then
-    echo "Images not found. Attempting to download images..."
-
-    if [ -f "./get_tezi_images.sh" ]; then
-        echo "Downloading TEZI images..."
-        ./get_tezi_images.sh
+if [ -z "${IS_WIC}" ]; then
```
lucas-akira: If `IS_WIC` is set with any non-empty string (e.g. `a`, `1`, `0`, `false`, etc.) this `if` statement won't get executed.

I suggest replacing this condition with `if [ "${IS_WIC}" != "true" ]` instead.

jsrc27 (approved):
LGTM

tests/integration/aval/aval-tests.yml
```diff
@@ -32,13 +32,16 @@ build-aval-docker:
...
+    - docker login -u "${CI_REGISTRY_USER}" -p "${CI_REGISTRY_PASSWORD}" "${CI_REGISTRY}"
     - docker login -u "${CI_DOCKER_HUB_PULL_USER}" -p "${CI_DOCKER_HUB_PULL_PASSWORD}"
   script:
     - ${B} docker build --build-arg CI_REGISTRY="${CI_REGISTRY}"
                         --build-arg CI_PROJECT_NAMESPACE="${CI_PROJECT_NAMESPACE}"
                         -f "${DOCKERFILE_NAME}"
-                        -t "${CI_REGISTRY_IMAGE}/${IMAGE_NAME_AVAL}:main" .
+                        -t "${CI_REGISTRY_IMAGE}/${IMAGE_NAME_AVAL}:${GITLAB_DOCKERREGISTRY_SUFFIX}" .
+    # Authenticate using password in necessary to push to torizoncore-builder repo
+    - docker login -u "${CI_REGISTRY_USER}" -p "${CI_REGISTRY_PASSWORD}" "${CI_REGISTRY}"
```
rborn-tx: Isn't this login already covered by the `docker login` in the `before_script` section?

.gitlab-ci.yml
```diff
@@ -283,6 +285,8 @@ test-torizoncore:
       when: never
     - if: $BUILD_MACHINE
   script:
+    # TODO: Since this creates a sub-shell, the environment variables created in setup.sh aren't persisting, because of this 'export TCBCMD' is required
+    - TCB_SKIP_PULL=1 ./setup.sh
     # TODO: Switch to Bash and use tcb-env-setup.sh script later
     - export TCBCMD="docker run --rm -v /deploy -v $(pwd)/workdir:/workdir -v storage:/storage --net=host -v /var/run/docker.sock:/var/run/docker.sock ${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}"
     - if [ -z ${INTERNAL_ARTIFACTORY_URL} ]; then echo "Missing variable INTERNAL_ARTIFACTORY_URL." && exit1; fi
```
rborn-tx: This wasn't part of your changes but I've just noticed a typo here: "exit1" -> "exit 1".

.gitlab-ci.yml
```diff
@@ -283,6 +285,8 @@ test-torizoncore:
       when: never
     - if: $BUILD_MACHINE
   script:
+    # TODO: Since this creates a sub-shell, the environment variables created in setup.sh aren't persisting, because of this 'export TCBCMD' is required
```
rborn-tx: I think the information in this comment is not correct in a few of ways:

- This does not create a subshell; it simply starts another program (that happens to be another shell).
- If the goal was to persist the variables then we could source the script, but this would require that script to be compatible with the outer (caller) shell (which is covered by the comment we have right below).
- Even if we sourced the script and the TCBCMD variable persisted, we'd want to change its value because we're taking the image from a specific registry that is different from the one selected by the setup script.

So, unless I'm missing something my suggestion is to drop this comment from here and from the place this was copied.

tests/integration/run_all.sh
```diff
@@ -126,23 +126,37 @@ if [ $? -ne 0 ]; then
...
-    fi
+if [ -n "${IS_WIC}" ]; then
+    if [ ! -e workdir/images/.raw_images_downloaded ]; then
+        echo "Raw Images not found. Attempting to download images..."
 
-    if [ ! -e workdir/images/.images_downloaded ] || [ ! -e workdir/images/.raw_images_downloaded ]; then
-        echo "Error: Images not found and could not be downloaded automatically."
-        echo "Please download images manually and place them in workdir/images/."
-        exit 1
+        if [ -f "./get_raw_images.sh" ]; then
```
rborn-tx: Why do we need a conditional here? Isn't the script part of the repo?

rborn-tx (changes requested):
Finished a pass.

tests/integration/run_all.sh
lucas-p-bernardes: Done, thanks!!

tests/integration/run_all.sh
lucas-p-bernardes: Done, thanks

tests/integration/aval/aval-tests.yml
lucas-p-bernardes: You are correct, I removed this. Thanks!

.gitlab-ci.yml
lucas-p-bernardes: Done in `gitlab-ci: Fix typo in test-torizoncore job` commit, thanks!!

tests/integration/run_all.sh
lucas-p-bernardes: Indeed this is unnecessary, I removed this! Thanks!

.gitlab-ci.yml
lucas-p-bernardes: Thanks for the explanation, I removed the comment! Thanks!

lucas-p-bernardes:
All the point were fixed, can we merge this?

rborn-tx (approved):
LGTM

PR #62, opened 2025-05-05
aval-tests: Fix rules override preventing AVAL_ALLOW_FAILURE from working

lucas-p-bernardes:


tests/integration/aval/aval-tests.yml
```diff
@@ -46,14 +46,6 @@ build-aval-docker:
...
     IMAGE_NAME: torizoncore-builder-amd64
     IMAGE_NAME_AVAL: aval-docker
-  rules:
-    - if: '$AVAL_ALLOW_FAILURE =~ /^true$/i'
-      allow_failure: true
-    - if: '$AVAL_DISABLED =~ /^true$/i'
-      when: never
-    - if: $CI_PIPELINE_SOURCE == "trigger"
-      when: never
-    - when: on_success
```
rborn-tx: Hmmm, the repetition being added by this solution looks really bad IMO :weary:...

A solution I see (that would need to be tested) would look like this:

1. Add the extra `if: $AVAL_TEST_ALL...` to be base template (since this is needed by most jobs):
   ```
   .aval-template:
     stage: test
     variables:
       IMAGE_NAME: torizoncore-builder-amd64
       IMAGE_NAME_AVAL: aval-docker
     rules:
       - if: '$AVAL_ALLOW_FAILURE =~ /^true$/i'
         allow_failure: true
       - if: '$AVAL_DISABLED =~ /^true$/i'
         when: never
       - if: $CI_PIPELINE_SOURCE == "trigger"
         when: never
       - when: on_success
       - if: '$AVAL_TEST_ALL =~ /^true$/i'       # <= Add this line to the template    since most jobs will need it
   ...
   ```
2. Create a `rules` template that does not include the  `if $AVAL_TEST_ALL` line:
   ```
   .rules-template1:
     rules: &aval-template-alt-rules
       - if: '$AVAL_ALLOW_FAILURE =~ /^true$/i'
         allow_failure: true
       - if: '$AVAL_DISABLED =~ /^true$/i'
         when: never
       - if: $CI_PIPELINE_SOURCE == "trigger"
         when: never
       - when: on_success
   ```
3. Remove the `rules` section of all existing jobs having the `if: '$AVAL_TEST_ALL...` rule.
4. Modify the other jobs to refer to the rules section of `rules-template1`, e.g.:
   ```
   aval-apalis-imx6q-scarthgap-release:
     extends: .aval-template
     rules: *aval-template-run-alt-rules
   ```

Shouldn't this work (and make things more concise)?

lucas-p-bernardes:
I found another issue, I tried this pipeline in `bullseye` with this variables:
```
AVAL_TEST_ALL = FALSE
AVAL_DISABLED = TRUE
AVAL_ALLOW_FAILURE = TRUE
RELEASE_TYPE = early-access
```
https://gitlab.com/toradex/rd/torizon-core/torizoncore-builder/-/pipelines/1810046530

Note that I got AVAL execution regardless of whether I set AVAL DISABLED = TRUE.

Thent I added one more paragraph in the commit message about it (and the fix):

"Further, previously, using AVAL_ALLOW_FAILURE=true caused jobs to run
regardless of other rules due to the implicit 'when: on_success' applied to all
rules. So, the AVAL_ALLOW_FAILURE=true caused the job to run even if
AVAL_DISABLED=true, for example. This commit changes the order of the
rules and reverse the behavior of AVAL_TEST_ALL `if` statement to
guarantee the correct evaluation of the rules."

tests/integration/aval/aval-tests.yml
lucas-p-bernardes: Done!

lucas-p-bernardes:
Pipelines:
```
AVAL_TEST_ALL = FALSE
AVAL_DISABLED = FALSE
AVAL_ALLOW_FAILURE = TRUE
RELEASE_TYPE = early-access
```
Result: https://gitlab.com/toradex/rd/torizon-core/torizoncore-builder/-/pipelines/1812965566

```
AVAL_TEST_ALL = TRUE
AVAL_DISABLED = FALSE
AVAL_ALLOW_FAILURE = TRUE
RELEASE_TYPE = early-access
```
Result: https://gitlab.com/toradex/rd/torizon-core/torizoncore-builder/-/pipelines/1812966272

```
AVAL_TEST_ALL = TRUE
AVAL_DISABLED = TRUE
AVAL_ALLOW_FAILURE = TRUE
RELEASE_TYPE = official
```
Result: https://gitlab.com/toradex/rd/torizon-core/torizoncore-builder/-/pipelines/1813443296

lucas-akira (approved):
Latest revision looks good to me

rborn-tx (approved):
LGTM

jsrc27 (approved):

PR #63, opened 2025-05-12
backend/dt: Show DTC output in debug log level when building custom DTs

lucas-akira:
In addition, if running the command at the default log level and DTC has any warnings, TCB will inform the user about them, as well as indicate how to view them.

Related-to: TCB-443


rborn-tx (approved):
Nice addition. LGTM.

jsrc27 (approved):

PR #64, opened 2025-05-14
Implement JUnit report and XRay integration for CI tests

lucas-p-bernardes:
Gitlab CI pipeline: https://gitlab.com/toradex/rd/torizon-core/torizoncore-builder/-/pipelines/1818264201

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -14,6 +14,9 @@ variables:
   # RUNNER_TAG defined in schedule pipeline will overwrite this variable. This one is useful for commit
   # pipelines.
   RUNNER_TAG: ""
+  ENABLE_XRAY:
+    value: false
+    description: "Whether to send test results to JIRA using XRay"
 
   DEBIAN_RELEASE: "bullseye-slim"
   # Pipeline Settings
@@ -61,6 +64,7 @@ stages:
   - pre-build
   - build
   - test
+  - report
   - build-multiarch
   - deploy
   - deploy-multiarch
@@ -188,7 +192,7 @@ build-arm64:
     paths:
       - tests/integration/workdir/reports/report-${CI_JOB_NAME}.xml
 
-test-host-tc6-platform:
+test-host-tc6-platform-release:
   extends: .test-base
   stage: test
   variables:
@@ -199,9 +203,9 @@ test-host-tc6-platform:
     - ./run_all.sh --machine $TCB_MACHINE --report --testcase platform
       --tcb-custom-image ${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}
 
-test-host-tos7-platform:
+test-host-tos7-platform-release:
   needs:
-    - test-host-tc6-platform
+    - test-host-tc6-platform-release
   extends: .test-base
   stage: test
   variables:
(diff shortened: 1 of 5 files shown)


tests/integration/run.sh
```diff
@@ -223,8 +223,18 @@ echo -e "Starting integration tests...\n"
 
 # run tests
 if [ "$TCB_REPORT" = "1" ]; then
-    $BATS_CMD 2>&1 | tee $REPORT_FILE
+    $BATS_CMD
+    # ensure the report writing is finished
+    if ! is_report_complete "$REPORT_DIR/report.xml"; then
+        echo "Error: $REPORT_DIR/report.xml not created or incomplete!"
+        exit 6
+    fi
```
rborn-tx: @lucasbernardestoradex Can you explain why we need this logic here? If BATS finished executing then the report must have been fully written already, right?

.gitlab-ci.yml
```diff
@@ -58,16 +61,16 @@ services:
...
   - build
   - test
+  - report
   - build-multiarch
   - deploy
   - deploy-multiarch
   - tag
 
 lint-dockerfiles:
-  stage: pre-build
```
rborn-tx: Based on the commit message, I see that the goal of merging the `pre-build` stage into the `build` one was to speed up execution which makes sense.

However, my view for us having both of those stages was that we wanted to avoid  spending computing resources on building the image if there are any issues with the Dockerfile or the Python code and this idea conflicts with the merging of the two stages. So, here I think we should clarify if it's okay to spend time building an image even in the presence of issues on the Dockerfile and/or Python code. Maybe @jsrc27 could help with his view.

Anyways, if the goal is to speed up things, I wanted to ask if you @lucasbernardestoradex were able to understand why those basic jobs in the `pre-build` stage were taking long. Checking the executions on the internal Gitlab I see the stage usually took 4-5 minutes to run but now on `gitlab.com` it's taking 20+ minutes, about the same time the build stage takes. I wonder if the caching on the `tcbuilder-dev-latest` we're employing is actually working...

rborn-tx (commented):
Overall looks good. I've raised a few questions though.

tests/integration/run.sh
lucas-p-bernardes: @rborn-tx I removed this and executed a pipeline to show what's was happening: https://gitlab.com/toradex/rd/torizon-core/torizoncore-builder/-/pipelines/1827712459/

Note that in https://gitlab.com/toradex/rd/torizon-core/torizoncore-builder/-/pipelines/1827712459/test_report you can see that some reports failed due to "JUnit XML parsing failed". If you open one of this pipelines, and download the XML file in "Job Artifacts" area, you notice that the content of the file is just this:

```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites time="916">
```

Seems the process was finishing before fully write XML file, and returning this XML with missing parts.
This is the problem, but I don't fully understand why the CI was cutting the XML and maybe this solution is not the better one.

.gitlab-ci.yml
jsrc27: > Maybe @jsrc27 could help with his view.

For me I did appreciate having the standalone stage that checks the Dockerfile and runs the Python linter. Let's me get feedback early on any issues with PRs so I can quickly fix things without waiting too long. Now of course this was for the Github pipeline and this change is for Gitlab. Though similar principal would apply.

As @rborn-tx said though on Github usually this initial stage takes maybe several minutes tops, so it's strange that it takes so long on Gitlab.

lucas-p-bernardes: Ok, I agree with you.
Since this is not the initial purpose of the MR, which was the JUnit and XRay integration, I'm droping a12b75e1c4d93f40a0d8e266cd6a47bc76a5ad18 commit by now.

About the point raised that it is taking much longer after migration, I created TCB-690 ticket for this, so we can work in this later avoiding delaying this MR.

lucas-akira: I agree that the `pre-build` issue should be dealt with separately, as it needs a more thorough investigation on what is causing it.

tests/integration/run.sh
rborn-tx: @lucasbernardestoradex I'd suggest further investigating this issue to determine its cause. My concern is that we could be hiding a bigger problem by simply waiting the file to be written since we don't have an explanation for why this wait is needed.

lucas-p-bernardes: Ok, I'm taking a look!

lucas-p-bernardes: @rborn-tx Indeed, there was a bug in bats-core, that was already fixed. So I needed to bump bats-core to catch this fix.
Added commit, that I explain what happened: 
`tests/integration: Bump BATS versions to get fix in --report-formatter junit`

lucas-p-bernardes: Pipeline: https://gitlab.com/toradex/rd/torizon-core/torizoncore-builder/-/pipelines/1860905628/

lucas-p-bernardes:
Hello @lucas-akira @rborn-tx @jsrc27, can we have another round of review here?

lucas-akira (approved):

jsrc27 (approved):
LGTM

tests/integration/run.sh
rborn-tx: Glad to hear.

rborn-tx (approved):
LGTM

PR #65, opened 2025-05-23
Allow Docker images to be specified by tag+digest inside compose files

rborn-tx:
In theory, specifying a Docker image by a tag plus a digest (such as in `image:tag@sha256:<hex-digest`) doesn't make much sense, but Docker does accept it by basically ignoring the tag. However, the implementation in TorizonCore Builder did not work in such cases. With the present commit we try to mimic Docker's behavior by also ignoring the tag. This has an effect both when canonicalizing a compose file (via `platform push`) and when generating Lockboxes (via `platform lockbox`).

Related-to: TCB-689

### DIFF
--- a/tcbuilder/backend/platform.py
+++ b/tcbuilder/backend/platform.py
@@ -389,7 +389,8 @@ def get_referenced_images(compose):
         parsed_name = parse_image_name(image)
         assert parsed_name.tag.startswith(SHA256_PREFIX), \
             f"Image '{image}' not specified by digest"
-        image_per_service[svc_name] = (image, image_platform)
+        image_per_service[svc_name] = \
+            (parsed_name.get_name_with_tag(), image_platform)
 
     log.debug(f"Images being used in docker-compose: {image_per_service}")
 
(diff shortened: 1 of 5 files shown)


jsrc27 (approved):
LGTM.

Really weird that Docker "allows" this kinda notation.

lucas-akira (approved):
Looks good to me as well. Docker ignoring the tag when pulling an image in format `image:tag@digest` appears to be a deliberate decision, as indicated by one of the release notes for Docker Engine: https://github.com/moby/moby/releases/tag/v20.10.20

And apparently one of reasons for this notation to be allowed is to "give users a visible 'clue' of what tag was originally used to resolve the digest": https://github.com/moby/moby/issues/39769#issuecomment-522527395. Just keep in mind this comment was made several years ago.

rborn-tx:
> And apparently one of reasons for this notation to be allowed is to "give users a visible 'clue' of what tag was originally used to resolve the digest": [moby/moby#39769 (comment)](https://github.com/moby/moby/issues/39769#issuecomment-522527395). Just keep in mind this comment was made several years ago.

That's good to know. Thanks for checking!

PR #66, opened 2025-06-10
Use skopeo to generate Lockbox tarballs

rborn-tx:
Switch to `skopeo` to generate the Docker tarballs that go into a Lockbox. Also, introduce a couple of related improvements and upgrade the DinD version used by TorizonCore Builder.

For more detail, please refer to the individual commit messages.

The following tests have been done manually since they are not covered by our current CI tests:

- Build Lockboxes with different versions of DinD (19.03.8, 20.10.24, 23.0.6, 25.0.3) and install then into a device running Torizon OS 6.8.2 and 7.2.0. All combinations have been tested to make sure we can suggest users to use a different version in case they ever bump into compatibility issues.
- Build Torizon OS with bundled containers and install on a device, with all combinations above.

Related-to: TCB-692

### DIFF
--- a/tcbuilder/backend/bundle.py
+++ b/tcbuilder/backend/bundle.py
@@ -379,6 +379,14 @@ class DindManager(DockerManager):
         dind_client = docker.DockerClient(base_url=self.docker_host, tls=tls_config)
         return dind_client
 
+    def get_client_env(self):
+        env = {
+            "DOCKER_HOST": self.docker_host,
+            "DOCKER_TLS_VERIFY": "1",
+            "DOCKER_CERT_PATH": os.path.join(self.cert_dir, 'client')
+        }
+        return env
+
     def save_tar(self, output_file):
         """Create compressed tar archive of the Docker images"""
 
--- a/tcbuilder/backend/bundle.py
+++ b/tcbuilder/backend/bundle.py
@@ -137,7 +137,7 @@ class DindManager(DockerManager):
     # - "23.0.6-dind":   version supporting zstd compression of layers
     # - "25.0.3-dind":   version used by Torizon OS 7.x.y
     #
-    DIND_IMAGE_VERSION_DEFAULT = "19.03.8-dind"
+    DIND_IMAGE_VERSION_DEFAULT = "25.0.3-dind"
 
     DIND_VOLUME_NAME = "dind-volume"
     DIND_CONTAINER_NAME = "tcb-fetch-dind"
--- a/tcbuilder/backend/bundle.py
+++ b/tcbuilder/backend/bundle.py
@@ -128,7 +128,17 @@ class DindManager(DockerManager):
     Installer tool.
     """
 
-    DIND_CONTAINER_IMAGE = "docker:19.03.8-dind"
+    DIND_IMAGE_NAME_DEFAULT = "docker"
+
+    # Noteworthy versions:
+    #
+    # - "19.03.8-dind":  old version used before TCB 3.11
+    # - "20.10.24-dind": (closest to the) version used by Torizon OS 6.x.y
+    # - "23.0.6-dind":   version supporting zstd compression of layers
+    # - "25.0.3-dind":   version used by Torizon OS 7.x.y
+    #
+    DIND_IMAGE_VERSION_DEFAULT = "19.03.8-dind"
+
     DIND_VOLUME_NAME = "dind-volume"
     DIND_CONTAINER_NAME = "tcb-fetch-dind"
     TAR_CONTAINER_NAME = "tcb-build-tar"
@@ -203,6 +213,7 @@ class DindManager(DockerManager):
                 f"{self.cert_dir} is a shared location between this script and "
                 "the Docker host.")
 
+    # pylint: disable=too-many-locals
     def start(self, network_name="fetch-dind-network",
               default_platform=None, dind_params=None, dind_env=None):
         """Start manager
@@ -279,8 +290,11 @@ class DindManager(DockerManager):
 
         log.debug(f"Environment variables for DinD: {environ}")
         log.debug(f"Running DinD container: ports={ports}, network={network_name}")
+        dind_image = "{}:{}".format(
+            os.environ.get("DIND_IMAGE_NAME", self.DIND_IMAGE_NAME_DEFAULT),
+            os.environ.get("DIND_IMAGE_VERSION", self.DIND_IMAGE_VERSION_DEFAULT))
         self.dind_container = self.host_client.containers.run(
-            self.DIND_CONTAINER_IMAGE,
+            dind_image,
             privileged=True,
             environment=environ,
             mounts=mounts,
@@ -297,6 +311,7 @@ class DindManager(DockerManager):
             dind_ip = self.dind_container.attrs \
                 ["NetworkSettings"]["Networks"][network_name]["IPAddress"]
             self.docker_host = "tcp://{}:22376".format(dind_ip)
+    # pylint: enable=too-many-locals
 
     def _stop_container(self):
         log.info("Stopping DIND container")
--- a/tcbuilder/backend/platform.py
+++ b/tcbuilder/backend/platform.py
@@ -720,7 +720,8 @@ def fetch_compose_target(target, repo_url, images_dir, metadata_dir,
     # Fetch the manifests of all images.
     manifests_dir = os.path.join(metadata_dir, sha256 + ".manifests")
     os.mkdir(manifests_dir)
-    manifests_per_image = fetch_manifests(images, manifests_dir)
+    manifests_per_image = fetch_manifests(
+        images, manifests_dir, req_platforms=req_platforms)
 
     # Determine (image, platform) pairs referenced in the compose file.
     image_platform_pairs = set(image_per_service.values())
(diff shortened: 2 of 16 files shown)


torizoncore-builder.Dockerfile
```diff
@@ -75,13 +75,50 @@ RUN cd aktualizr && \
...
+
+# hadolint ignore=SC2046
+RUN echo "Building skopeo..." && \
+    cd ${GOPATH}/src/github.com/containers/skopeo && \
+    B="$(pwd)" && D="${B}/install-dir" && \
+    make clean && \
+    make EXTRA_LDFLAGS="-w -s" bin/skopeo && \
+    make EXTRA_LDFLAGS="-w -s" DESTDIR="${D}" DISABLE_DOCS="1" install && \
+    tar cjvf /root/skopeo.tar.bz2 -C "${D}" $(cd "${D}" && echo *)
+
```
lucas-akira, line 114: Couldn't we use skopeo from the Debian feeds instead of building it?

tcbuilder/backend/platform.py
```diff
@@ -588,7 +588,227 @@ def select_unique_images(image_platform_pairs, manifests_per_image,
...
+        log.debug("Transformed repositories data: {repos_data}")
+
+        with open(repos_path, "w") as fhandle:
+            json.dump(repos_data, fhandle, separators=(",", ":"))
+
+
+def _apply_aktualizr_blkread_wkaround(tbdir):
+    """Workaround for bug found in image loading in Aktualizr.
+
+    See: https://toradex.atlassian.net/browse/TOR-3854
```
lucas-akira: Should we reference internal links in the comments? Only people at Toradex would be able to see it.

tcbuilder/backend/bundle.py
```diff
@@ -279,8 +290,11 @@ def start(self, network_name="fetch-dind-network",
 
         log.debug(f"Environment variables for DinD: {environ}")
         log.debug(f"Running DinD container: ports={ports}, network={network_name}")
+        dind_image = "{}:{}".format(
+            os.environ.get("DIND_IMAGE_NAME", self.DIND_IMAGE_NAME_DEFAULT),
+            os.environ.get("DIND_IMAGE_VERSION", self.DIND_IMAGE_VERSION_DEFAULT))
```
jsrc27, line 295: Is the idea here we set the default and users should adjust this themselves as needed? Or will there be an upcoming mechanism that will change these to the appropriate version?

tcbuilder/backend/platform.py
```diff
@@ -588,7 +588,227 @@ def select_unique_images(image_platform_pairs, manifests_per_image,
...
+    else:
+        log.debug(f"No need to add dummy file to '{tbdir}'.")
+
+
+def amend_image_tarball(image_fname, image_spec, skopeo_generated=True):
+    # Apply the workaround for Skopeo canonicalizing the image names which
+    # prevents the loading by Aktualizr since the names are validated against
+    # the docker-compose file.
+    canon_wkaround = (skopeo_generated and
+                      os.environ.get("APPLY_SKOPEO_CANON_WKAROUND", "1") == "1")
```
jsrc27, line 735: Is there a reason a user would not want this enabled?

tcbuilder/backend/platform.py
```diff
@@ -588,7 +588,227 @@ def select_unique_images(image_platform_pairs, manifests_per_image,
...
+
+    # Apply the workaround for a bug in the Docker image validation which happens if the image is
+    # slightly bigger than a multiple of the block reading size (256KB).
+    blkread_wkaround = (os.environ.get("APPLY_AKTUALIZR_BLKREAD_WKAROUND", "1") == "1")
```
jsrc27, line 739: Similar question here, what's the case where this may not be wanted?

tcbuilder/backend/platform.py
```diff
@@ -657,14 +878,29 @@ def build_docker_tarballs(unique_images, target_dir, host_workdir,
...
-            with open(image_fname, "wb") as outf:
-                for image_data in image_obj.save(named=True):
-                    outf.write(image_data)
+
+            if os.path.exists(image_fname):
+                # Sanity check: images in a Lockbox should be unique.
+                raise TorizonCoreBuilderError(
+                    "Container image file '{image_fname}' already exists")
+
+            tar_generator = os.environ.get("LOCKBOX_TARBALL_GENERATOR", "skopeo")
```
jsrc27, line 905: What's the reason to keep the old method with `docker`? Just for backward compatibility, or is there another reason?

tcbuilder/backend/platform.py
```diff
@@ -730,6 +967,19 @@ def fetch_compose_target(target, repo_url, images_dir, metadata_dir,
...
+    # sure the digests alone are also unique. If not, it means the same image is
+    # being referenced in two different ways which the code in Aktualizr would
+    # not handle correctly. Users can easily work around this corner case by
+    # always referencing an image in the same way in their compose file.
+    digests_unique = set()
+    for _spec, sel_digest in images_selection:
+        if sel_digest in digests_unique:
+            raise TorizonCoreBuilderError(
+                f"Image with digest {sel_digest} has been selected more than once"
+                " through different names which is currently not supported.")
```
jsrc27: Is the log message here the correct message we want to send?

It says "currently not supported" which softly implies it's something we do plan to support. Is that true or do we never plan to support a compose file that references the same image more than once?

jsrc27 (commented):
Left some questions about UX/use-cases regarding the new adjustable environment variables.

tcbuilder/backend/bundle.py
rborn-tx: I made these two variables configurable basically for two reasons:

- So we can test with different versions without having to change the code.
- So that we can ask users to switch to a different version in case they face any issues and we suspect that it may be related to the DinD version.

In the manual tests I mentioned, I made a compose file referencing 10 different images including OCI/non-OCI (Docker), with/without provenance, few/many layers and the such and things seemed to work fine with all combinations of TOS x DinD versions, but we never know...

To set these variables or the others, one would basically pass the `-e<VAR>=<VALUE>` parameter to the `tcb-env-setup.sh` script and then the alias produced would use it on all invocations of `torizoncore-builder`.

tcbuilder/backend/platform.py
rborn-tx: I don't expect users to change this particular variable. It's useful mainly for our testing.

tcbuilder/backend/platform.py
rborn-tx: That's another variable useful mainly for testing. For this particular workaround, I've created a ticket to solve the issue at the Aktualizr side. So, for testing that fix we'd leverage this particular variable.

tcbuilder/backend/platform.py
rborn-tx: The main reason for now is the possibility of issues with Skopeo which, despite all the testing, could pop up. So I wanted to have a way to exactly replicate the old behavior by just setting variables.

Another reason for the long term is that we could at some point consider implementing the new "docker save" format in case new features are added to the images that cannot be represented by the "v2s2" format that Aktualizr currently understands. If we made that implementation, in the field there would be versions of AKtualizr that support and others that don't support the new format in which case we might document this variable for general use.

tcbuilder/backend/platform.py
rborn-tx: Good point! Let me remove that word.

We could always decide to remove that limitation at some point but it would involve changes in TorizonCore Builder and Aktualizr. Since the workaround is so easy for users, it's probably not worth doing it though.

lucas-akira (commented):
Lots of work done here, looks pretty good. I only have a few questions.

tcbuilder/backend/platform.py
jsrc27: Ahh okay makes sense then for internal usage.

tcbuilder/backend/platform.py
jsrc27: Makes sense.

tcbuilder/backend/platform.py
jsrc27: Ahh okay, so the thought process is more along the lines of "just in case". That way we still retain the original code as it's currently written.

tcbuilder/backend/bundle.py
jsrc27: Basically we have users use the default `19.03.8-dind` unless there are issues?

Would it be better to have the default set to match the latest Torizon OS? Or even the latest Torizon OS LTS? Or does it make sense to have the default be this older `19.03.8-dind` version?

tcbuilder/backend/platform.py
rborn-tx: Exactly.

tcbuilder/backend/bundle.py
rborn-tx: OMG, thanks for noticing. It looks like I managed to drop the commit that separately bumped the default version of DinD :weary:. I've just restored it.

torizoncore-builder.Dockerfile
rborn-tx: Unfortunately the version from the Debian feeds is too old to work with DinD 25.0.3, even on bookworm.

tcbuilder/backend/platform.py
rborn-tx: Good point! Indeed I see no other links around the code so it's probably best to avoid this one. Let me shortly describe the bug as part of the comment instead.

torizoncore-builder.Dockerfile
lucas-akira: Hmm I see. In that case building it is really the only option we have.

jsrc27 (approved):

lucas-akira (approved):

PR #67, opened 2025-06-27
Add new 'secboot sign-kernel' and 'secboot sign-bootloader-hab' commands

lucas-akira:
Add new command that re-signs existing signed Torizon OS 7 images. Specifically the command will sign the kernel fitImage and the bootloader container (`flash.bin`/`imx-boot`) with keys provided by the user. The end result is a copy of the input OS image with the aforementioned components re-signed.

For now the only supported board is the Verdin i.MX8M Plus. Other boards are planned to be supported in the future.

The kernel fitImage is signed with a user-provided directory with the RSA key and certificate. The bootloader container is signed with a directory tree with keys, certificates, SRK table and SRK fuse hash generated beforehand using the NXP Code Signing Tool.

Test cases were also added for this new command.

Related-to: TCB-158


tcbuilder/backend/secure_boot.py
```diff
@@ -0,0 +1,644 @@
...
+
+    sig_match = re.search(r"Signature written to '(.*)'", mkimage_output, re.IGNORECASE)
+    pub_match = re.search(r"Public key written to '(.*)'", mkimage_output, re.IGNORECASE)
+
+    if sig_match and pub_match:
+        log.info(f"Signature written to '{sig_match.group(1)}'")
+        log.info(f"Public key written to '{pub_match.group(1)}'")
+
+        if SUPPORTED_SIGNING_MACHINES[board] == "imx8m":
+            copy_sig_node_to_config_of_list(input_binaries_dir, board)
```
jsrc27: Is there a reason we check specifically for imx8m here? I know currently this only supports imx8mp, but in the future won't this function be used on other machines as well? Or is there a reason it's imx8m specific right now?

tcbuilder/backend/secure_boot.py
```diff
@@ -0,0 +1,644 @@
...
+UBOOT_DTB_BINARY = "u-boot.dtb.out"
+
+SUPPORTED_SIGNING_MACHINES = {"verdin-imx8mp": "imx8m"}
+
+SECURE_BOOT_FILES_DIR = "/builder/tcbuilder/secure_boot_files"
+COPY_SIG_NODE_SCRIPT = "copy_signature_node_to_fdts.sh"
+
+FUSE_CMD_TXT_NAME = "fuse-cmds.txt"
+
+def check_if_file_exists(filename, directory):
```
jsrc27: This seems like a generic function that could go in `common.py`.

tcbuilder/backend/secure_boot.py
```diff
@@ -0,0 +1,644 @@
...
+    if wrong_keyname_match:
+        expected_keyname = wrong_keyname_match.group(1)
+        raise FileContentMissing(f"mkimage expected a key pair named ({expected_keyname}.key, "
+                                 f"{expected_keyname}.crt) to sign the kernel fitImage "
+                                 "(key-name-hint property). Aborting.")
+
+    sig_match = re.search(r"Signature written to '(.*)'", mkimage_output, re.IGNORECASE)
+    pub_match = re.search(r"Public key written to '(.*)'", mkimage_output, re.IGNORECASE)
+
+    if sig_match and pub_match:
```
jsrc27: I assume this if doesn't need an else branch because it would be impossible for that to happen?

tcbuilder/cli/secure_boot.py
```diff
@@ -0,0 +1,156 @@
...
+CST_SRK_DEFAULT_TABLE = "SRK_1_2_3_4_table.bin"
+CST_SRK_DEFAULT_FUSE = "SRK_1_2_3_4_fuse.bin"
+
+def do_sign(args):
+    """Run 'secure-boot sign' subcommand"""
+
+    for directory in (args.input_dir, args.kernel_key_dir, args.cst_dir):
+        if not os.path.isdir(directory):
+            raise InvalidArgumentError(
+                f"Directory \"{directory}\" does not exist: aborting.")
```
jsrc27: Possible idea. I understand the `cst_dir` must already have all the keys/certs generated prior by the user. These get checked in the bash signing script later on in the process.

Would it possibly be better to roughly check whether the key/cert files exist early on, just to give an early error and exist. Otherwise the user would have to wait later on in the process to get an error.

Or we just keep it as is and just check the keys/certs later, which I don't think is a big issue all in all.

tcbuilder/cli/secure_boot.py
```diff
@@ -0,0 +1,156 @@
...
+              "directory must be specified with the --cst-dir argument. Keys, certificates, SRK "
+              "table, and e-fuse hash must be generated beforehand by following the NXP "
+              "documentation."),
+        description=("Currently supported machines: "
+                     f"{', '.join(secure_boot.SUPPORTED_SIGNING_MACHINES)}"))
+
+    subparser.add_argument(
+        dest="input_dir",
+        metavar="INPUT_DIRECTORY",
+        help="Path of Torizon OS image (in Toradex Easy Installer format) to be signed. U-Boot "
```
jsrc27: According to the commit message, we're only creating this new command for Torizon OS 7.X images specifically. Perhaps we should specify here. Also if we want to enforce this we should also probably check the version of the input image.

jsrc27 (commented):
Gave it an initial review and left some comments/questions.

One more question though. Is there a way to check whether the input Tezi image isn't already signed? Could be a nice thing to check early on in the command. At least if it's easy to check, otherwise I don't think it would be a big deal since the command is pretty transparent in what it will do.

tcbuilder/backend/secure_boot.py
lucas-akira: This function just calls a shell script that I basically copied from [meta-toradex-security](https://github.com/toradex/meta-toradex-security/tree/8646da8e4686fdfcb54eb7abfea2d7489fd48303/recipes-bsp/u-boot/files).

For the signing process I've tried to replicate what meta-toradex-security does, and given that in it [this script is only called for the iMX8M machines](https://github.com/toradex/meta-toradex-security/blob/8646da8e4686fdfcb54eb7abfea2d7489fd48303/recipes-bsp/u-boot/u-boot-fit-signature-nxp.inc#L46), that's what I did.

To be honest I don't really know how this procedure is done for other machines, or if it's actually needed if not using `binman` to assemble the bootloader container, as I've only focused on the current Verdin i.MX8M Plus implementation.

For the function I passed the machine name as an argument (`board`), just in case other similar scripts are added for other machines. Of course, the if statement there would need to be updated as well.

tcbuilder/backend/secure_boot.py
lucas-akira: Hmm, actually I forgot to consider this case when I wrote the function. I think the code should raise an exception there, just to be safe. I'll add it, thanks.

tcbuilder/backend/secure_boot.py
lucas-akira: I agree, makes sense. I'll put it there then, thanks.

tcbuilder/cli/secure_boot.py
lucas-akira: Right, I'll add the version check and clarify that only Torizon OS 7 images are supported.

tcbuilder/cli/secure_boot.py
lucas-akira: That's a good idea. I'll add this verification right after the initial image support check.

torizoncore-builder.Dockerfile
rborn-tx, line 1: Regarding the commit "Add tools and dependencies related to secure boot support", I'd suggest adding the prefix "Dockerfile:" to its commit title.

torizoncore-builder.Dockerfile
```diff
@@ -198,6 +207,13 @@ RUN groupadd --gid $USER_GID $USERNAME && \
 
 FROM tcbuilder-base
 
+# Build U-Boot tools for secure boot support (U-Boot commit hash: 3f772959501c99fbe5aa0b22a36efe3478d1ae1c)
+RUN git clone https://source.denx.de/u-boot/u-boot.git -b v2024.07 /u-boot-repo && \
+    cd /u-boot-repo && make tools-only_defconfig && make tools-only && make scripts
+
+RUN mkdir /u-boot && mv /u-boot-repo/tools /u-boot && \
+    mv /u-boot-repo/scripts /u-boot && rm -rf /u-boot-repo
```
rborn-tx: Shouldn't these two RUN commands be collapsed on a single one to avoid wasting space in the TCB container?

torizoncore-builder.py
```diff
@@ -25,7 +25,7 @@
 import traceback
 
 from tcbuilder.cli import (bundle, build, combine, deploy, dt, dto, images, isolate,
-                           kernel, ostree, platform, push, splash, ubootenv, union)
+                           kernel, ostree, platform, push, secure_boot, splash, ubootenv, union)
```
rborn-tx: This is mostly personal taste but let me ask anyway: how about renaming `secure_boot.py -> secboot.py` more in line with the short module names we've used so far? Btw, the name of the file does not necessarily need to match the name of the command.

Regardless of the above, maybe even the name of the command could also be shortened in the same way, i.e. instead of writing `torizoncore-builder secure-boot`, the user would write `torizoncore-builder secboot` (again, inline with the short names we've used such as the recent `ubootenv` command). But of course this is related to UX, so we could ask @leograba's opinion here.

tests/integration/samples/signing_keys/cst_dirs.tar.gz
rborn-tx, line 1: Regarding this tarball, I have a couple of suggestions:

- When creating the file, use root/root as user/group avoid data with personal info.
- To avoid having multiple copies of the CST binary, put the binaries into a separate directory and then copy them to the desired destinations after unpacking.
- Add a new level to separate HAB from the future AHAB test data.

With this we would go from:

```
cst_dirs/
├── {cst-3.4.1_tcb_test_ecdsa_p384,cst-3.4.1_tcb_test_ecdsa_p521_no_ca,cst-3.4.1_tcb_test_rsa_1024_no_ca,cst-3.4.1_tcb_test_rsa_2048}
│   ├── crts
│   ├── keys
│   ├── linux32
│   │   └── bin
│   └── linux64
│       └── bin
```

to something like this:

```
cst_dirs/
├── cst-3.4.1
│   ├── linux32
│   │   └── bin
│   └── linux64
│       └── bin
└── hab
    └── {cst-3.4.1_tcb_test_ecdsa_p384,cst-3.4.1_tcb_test_ecdsa_p521_no_ca,cst-3.4.1_tcb_test_rsa_1024_no_ca,cst-3.4.1_tcb_test_rsa_2048}
        ├── crts
        └── keys
```

tcbuilder/cli/secure_boot.py
```diff
@@ -0,0 +1,156 @@
...
+
+    subparser.add_argument(
+        "--force", dest="force",
+        default=False, action="store_true",
+        help="Force program output, overwriting any existing file/directory.")
+
+    subparser.add_argument(
+        "--kernel-key-dir", dest="kernel_key_dir",
+        help="Kernel fitImage key directory path.",
+        required=True)
```
rborn-tx: I think we're missing the kernel key name, no? The name `dev` that we have today is just an example that shouldn't be used except for development.

tcbuilder/backend/secure_boot.py
lucas-akira: Done. I've also moved `get_tezi_uenv_txt_vars()` to `common.py` as it is a similar generic function as well.

lucas-akira:
I've changed `--cst-dir` to no longer be a required argument, as the CST directory should only be necessary when signing HAB-compatible (and eventually, AHAB-compatible) modules. Now the directory is only checked after TCB detects that the input image is for a HAB-compatible machine. 

I've also changed the command help description to indicate that the CST directory is for iMX-based modules.

> One more question though. Is there a way to check whether the input Tezi image isn't already signed? Could be a nice thing to check early on in the command. At least if it's easy to check, otherwise I don't think it would be a big deal since the command is pretty transparent in what it will do.

I thought about it, but I'm not really sure if there's a way to easily verify if the input image is signed. The `uEnv.txt` contents of a signed image have additional variables and functions, but I'm not sure if there's something reliable enough to confirm if the image is signed or not.

tcbuilder/cli/secure_boot.py
```diff
@@ -0,0 +1,155 @@
...
+        "secure-boot",
+        help=("Sign different components of a specified Torizon OS image in "
+              "Toradex Easy Installer format."),
+        allow_abbrev=False)
+
+    subparsers = parser.add_subparsers(title='Commands', required=True, dest='cmd')
+
+    # secure-boot sign
+    subparser = subparsers.add_parser(
+        "sign",
```
rborn-tx: I see here we are considering a single "sign" command that will do the signing of all components that might need signing with all possible technologies and I wonder if we wouldn't quickly get to the same situation we have today with the `platform push` command (which has many switches that are mutually exclusive and depend on the type of package you're trying to push).

A solution to this would be to make the signing more granular and technology specific. For example:

```
# Commands for signing the bootloader binary only:
torizoncore-builder sign-bootloader-hab <parameters-required-for-HAB>
torizoncore-builder sign-bootloader-ahab <parameters-required-for-AHAB>
# Commands for signing the kernel:
torizoncore-builder sign-kernel <parameters-required-for-kernel-signing>
```

Of course this would lead to the use of multiple commands for signing all components but this goes in line with the idea of providing lower level control through individual CLI commands and then later putting them together with the `build` command.

Anyways, I think we should discuss this topic on a meeting.

tcbuilder/cli/secure_boot.py
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    secure_boot.sign(
+        input_dir=args.input_dir,
+        kernel_key_dir=args.kernel_key_dir,
+        output_dir=args.output_dir,
+        force=args.force,
+        cst_dir=args.cst_dir,
+        cst_args=cst_args
+    )
```
rborn-tx: Is there anything preventing us from making the changes in-place? We don't necessarily need to provide that possibility to the user, but it would be very useful when we implement the signing via the `build` command.

tcbuilder/backend/common.py
```diff
@@ -803,3 +806,71 @@ def validate_compose_file(compose_file_data):
...
+        tar_compress_options = get_tar_compress_program_options(rootfs_tar)
+        tarcmd = [
+            "tar",
+            "-xf", rootfs_tar,
+            "-C", tempdir,
+        ] + tar_compress_options
+        log.debug(f"Running tar command: {shlex.join(tarcmd)}")
+        subprocess.check_output(tarcmd, stderr=subprocess.STDOUT)
+
+        uenv_txt_path = os.path.join(tempdir, "boot/loader/uEnv.txt")
```
rborn-tx: How about only unpacking the needed file (`uEnv.txt`)? Or just the directory where it's expected to be `boot/loader/`?

tcbuilder/backend/secure_boot.py
rborn-tx, line 1: I'll review this file on a second pass because the items I've raised may lead to structural changes impacting it.

rborn-tx (changes requested):
I've finished my 1st pass. Good job so far!

torizoncore-builder.py
leograba: For this single command, I don't have a strong preference or user indication on long versus short form.

Overall, it's best to keep a unified look and feel. If all commands follow a pattern of being short, let's opt for that.

lucas-akira: I don't have any particular preference for either name, but since we already have a command in short form, we can change both the filename and the command itself to `secboot`.

tests/integration/samples/signing_keys/cst_dirs.tar.gz
lucas-akira: > When creating the file, use root/root as user/group avoid data with personal info.

Thanks, I'll change that.

> To avoid having multiple copies of the CST binary, put the binaries into a separate directory and then copy them to the desired destinations after unpacking.

I thought about tolerating the multiple copies of the binary given their small size, but on second thought, it really doesn't make sense to do this. I'll follow your suggestions.

tcbuilder/backend/secure_boot.py
jsrc27: Ahh I see now, that makes sense then.

tcbuilder/cli/secure_boot.py
lucas-akira: In-place changes can be done if the provided output directory is the same as the input one and `--force` is applied. The backend part of the command deals with this logic.

tcbuilder/backend/common.py
lucas-akira: Right, that would be considerably faster than unpacking the whole rootfs tar.

Given that `loader` is a symlink to directory loader._n_, where _n_ is a number that can change, I think we would have to unpack something like `./boot/loader*`, right?

I've tried passing `./boot/loader/uEnv.txt` directly to the extraction command, but that gives me a `Not found in archive` error, even when using the `--dereference`/`-h` option.

Apparently `--dereference`/`-h` only applies when compressing files, not when extracting them.

tcbuilder/cli/secure_boot.py
lucas-akira: >  I wonder if we wouldn't quickly get to the same situation we have today with the `platform push` command (which has many switches that are mutually exclusive and depend on the type of package you're trying to push)

That's a good point. With other boards having different secure boot technologies a single sign command would quickly get cluttered with lots of technology-specific arguments as other machines are supported.

> A solution to this would be to make the signing more granular and technology specific. For example:
> 
> ```
> # Commands for signing the bootloader binary only:
> torizoncore-builder sign-bootloader-hab <parameters-required-for-HAB>
> torizoncore-builder sign-bootloader-ahab <parameters-required-for-AHAB>
> # Commands for signing the kernel:
> torizoncore-builder sign-kernel <parameters-required-for-kernel-signing>
> ```
> 
> Of course this would lead to the use of multiple commands for signing all components but this goes in line with the idea of providing lower level control through individual CLI commands and then later putting them together with the `build` command.
> 
> Anyways, I think we should discuss this topic on a meeting.

Yeah, I agree with the solution. So these sign sub-commands would be grouped inside a common `secboot` command (as you suggested in the other topic), like:

```
torizoncore-builder secboot sign-kernel

torizoncore-builder secboot sign-bootloader-hab
```

It shouldn't be very difficult to implement this. I'll work on it.

tcbuilder/cli/secure_boot.py
lucas-akira: Hm, I didn't implement the feature of changing the key name. I know that `mkimage` will not sign if it cannot find key files that match the value of `key-name-hint`.

In order to change the expected key name, TCB would have to change the `key-name-hint` field with the new key name for all configuration nodes of the kernel FIT image, right?

tcbuilder/backend/common.py
rborn-tx: You could unpack `./boot/loader*/uEnv.txt`, for example. Or even `./boot/` would already be a gain.

tcbuilder/cli/secure_boot.py
rborn-tx: > In order to change the expected key name, TCB would have to change the key-name-hint field with the new key name for all configuration nodes of the kernel FIT image, right?

Yes, I believe so. The name is configurable on the Yocto side, so I think we'd need to do that anyway, otherwise we wouldn't support images where the name was changed to something else.

Also, even if we only supported input images where the key name was `dev`, users would likely want to sign their output images with a different key - say `prod` - to distinguish their development images from the production ones. This would still require updating the `key-name-hint` property, unfortunately.

torizoncore-builder.Dockerfile
lucas-akira: Good catch. I'll put everything into a single RUN instruction.

torizoncore-builder.Dockerfile
lucas-akira: Right, it would be clearer to indicate in the subject line where the changes are located. I'll add it.

lucas-akira: Done!

tests/integration/samples/signing_keys/cst_dirs.tar.gz
lucas-akira: OK, done!

tcbuilder/backend/common.py
lucas-akira: > You could unpack `./boot/loader*/uEnv.txt`, for example. Or even `./boot/` would already be a gain.

First suggestion did the trick!

lucas-akira:
I've made some significant changes to the code according to the suggestions:

- Renamed the `secure-boot` command to `secboot`;

- Removed the general `sign` sub-command and divided the kernel and the bootloader signing into two distinct sub-commands: `sign-kernel` and `sign-bootloader-hab`;

- In the kernel signing command, added the feature to put a new key name chosen by the user. TCB does this by doing the following:
    - Any references to previous existing key names in the U-Boot DTB and DTBs in CONFIG_OF_LIST are removed, as otherwise U-Boot will try to validate them when booting the kernel of the re-signed image;
    - All configuration nodes of the kernel FIT are modified to accept the new key name.

- Adapted the tests to all the changes made above.

For the rest of the suggestions I've addressed them in their own conversation thread. I've also rebased the code to be up-to-date with the latest commits from `bullseye`.

Because of all of these changes, I'm requesting a new review.

jsrc27 (commented):
Question.

Now that we can separately sign the bootloader, or the fitimage. If a user wants to do both, does the order they do the commands matter?

Since the signing of the fitimage requires modification of U-Boot (DTB), I would think then the U-Boot would need to be re-signed if it was signed prior.

For example if a user goes:

* `tcb secboot sign-kernel`
* `tcb secboot sign-bootloader-hab`

That should be okay, but if they go:

* `tcb secboot sign-bootloader-hab`
* `tcb secboot sign-kernel`

Then the resulting bootloader wouldn't be properly signed anymore, or would it? I remember there was a weird order of operations thing here with the signing when we did this in Yocto.

lucas-akira:
The order the commands are done matters, yes. If a user wants to sign both the kernel FIT and the bootloader, they should do `sign-kernel` first then sign the resulting OS image with `sign-bootloader-hab`, as you said.

> Then the resulting bootloader wouldn't be properly signed anymore, or would it? I remember there was a weird order of operations thing here with the signing when we did this in Yocto.

~~As I understand it you are correct: if the order of the commands is reversed i.e. sign the bootloader first then the kernel, the bootloader signature would become invalid.~~

~~When signing the kernel, the U-Boot DTB needs to be updated with the public key. As the U-Boot DTB is part of the bootloader container, changing it will change the hash to be different than the expected one, resulting in a failed bootloader signature verification.~~

EDIT: My understanding here was conceptually incorrect, as explained below.

rborn-tx:
@jsrc27 @lucas-akira Guys,

On a chain of trust system where each component has the public key for verifying the next component, the order of signing **must not matter**. That's what allows the components to be updated independently.

Keep in mind that each component in the chain of trust has only the pub key for validating the next component. There are no hashes related to the next component at all.

In Yocto, there was a particular order in which the signing was performed simply because of tooling limitations, namely the `mkimage` tool responsible for signing the kernel is/was also responsible for updating the U-Boot DTB with the pub key. However, there are other ways to use the tool and in later versions of `openembedded-core` they changed the usage it so that the update of the U-Boot DTB is done independently/without the need of a kernel. If we took some of the signing logic from Yocto we could have inherited the limitation but then it's just a matter of following the the latest approach to fix the situation.

lucas-akira:
> Keep in mind that each component in the chain of trust has only the pub key for validating the next component. There are no hashes related to the next component at all.

Hmm you're right, I mixed up the concepts. The hashing of the binary is done at execution time when verifying the signature. It doesn't matter if one component in the chain of trust changes as long as it's signed with a key that has its corresponding public key in the previous component.

----

In the current implementation of the `secboot sign-kernel` command, the bootloader container is recreated to update the public key in the U-Boot DTB, and the current one is overwritten.

The recreated bootloader when running this command is unsigned, so as a result of this implementation, the signing order ends up being relevant.

Maybe it would be better to keep the bootloader container unaltered when signing the kernel and pass the kernel public key in the bootloader signing command instead.

jsrc27:
> In the current implementation of the secboot sign-kernel command, the bootloader container is recreated to update the public key in the U-Boot DTB, and the current one is overwritten.

Okay so just to summarize currently the current order of commands/signing **does** matter, yes?

But, we want the order to **not** matter.

If I understood the situation correctly, then I'm glad I asked about this point so we're all on the same page.

rborn-tx:
> Okay so just to summarize currently the current order of commands/signing **does** matter, yes?
> 
> But, we want the order to **not** matter.
> 
> If I understood the situation correctly, then I'm glad I asked about this point so we're all on the same page.

Absolutely, that was an very important question and I think your summary is correct.

~I believe a more accurate view is that currently~ The two steps are simply tied together by implementation details. In other words, the two steps are independent in concept but the implementation made them coupled (perhaps due to following the previous implementation in `openembedded-core`). By ~splitting the commands from the users perspective and~ making the steps independent (thus not reliant on ordering) we'd have a more sound/principled implementation.

PS.: When I originally answered this I didn't realize the splitting was already done.

lucas-akira:
Alright, then I'll work on making the commands independent from one another. I'll remove the public key update step on `sign-kernel` (i.e. for this command the bootloader won't be modified) and I'll put it as an optional step in `sign-bootloader-hab`: if the user provides the kernel pubkey, it's assumed that the U-Boot DTB + CONFIG_OF_LIST DTBs should be updated with it before signing the bootloader.

lucas-akira:
Just pushed the requested changes; now the order the `sign-kernel` and `sign-bootloader-hab` commands are executed on a given input Torizon OS image shouldn't matter: the resulting OS image will have both components re-signed.

I did test both cases of signing the bootloader first then the kernel, and signing the kernel first followed by the bootloader: For each case I've flashed the resulting image to a Verdin i.MX8M Plus and checked for HAB events on U-Boot then checked if the kernel FIT passes all hash integrity tests and boots successfully.

For both cases no HAB events were found and the kernel booted without issues.

tcbuilder/cli/secboot.py
```diff
@@ -0,0 +1,239 @@
...
+        help="Kernel fitImage key directory path. If specified the U-Boot DTB and Control DTBs "
+             "will be updated with the public key before signing the bootloader. This directory "
+             "must contain a private key in .key format and a certificate with "
+             "the public key in .crt format, both with the same name.")
+
+    subparser.add_argument(
+        "--kernel-key-algo", dest="kernel_key_algo",
+        default=KERNEL_KEY_DEFAULT_ALGO,
+        help="Comma-separated pair of hashing and crypto algorithms used to sign the kernel, if "
+             f"updating the public key in U-Boot DTB. (default: {KERNEL_KEY_DEFAULT_ALGO})")
```
jsrc27: So if these optional arguments are present that triggers the keys being added to the bootloader dtbs. But it does not actually sign the fit-image, still requiring the user to run `sign-kernel`, correct?

So if a user wants everything signed they should always add these arguments whether `sign-bootloader-hab` is executed first or second, right?

Is there a good way to warn/caution the user that these arguments should be added in the use-case where they want everything signed? Cause with how many arguments this command has I can imagine it would be easy for users to forget, or not understand these optional arguments. Then if they forget these and the DTBs aren't probably updated prior to signing they'll have a "bad" image essentially, which I guess won't be apparent to them until they flash and run it.

lucas-akira: > So if these optional arguments are present that triggers the keys being added to the bootloader dtbs. But it does not actually sign the fit-image, still requiring the user to run sign-kernel, correct?

That's correct. In this case what is being added to the DTBs is just the public key, which the bootloader needs to verify the kernel signature.

> So if a user wants everything signed they should always add these arguments whether sign-bootloader-hab is executed first or second, right?

Exactly. If the bootloader container and the kernel FIT are to be signed with a new key, then yes, `sign-bootloader-hab` needs to have `--kernel-key-dir` (`--kernel-key-algo` can be omitted if the default value suits the user) . Otherwise the bootloader won't have the new public key and it will be unable to correctly verify the new kernel signature.

I can add a message at the end of the `sign-bootloader-hab` command if `--kernel-key-dir` wasn't provided, warning users that they should pass the argument if signing the kernel as well. I thought of writing something like this:

```
The bootloader DTBs were NOT updated with a new public key.
If the kernel fitImage will be signed with a new key, please re-run the command with --kernel-key-dir so the new kernel signature can be properly verified by the bootloader.
Otherwise, this message can be ignored.
```

jsrc27: Yeah that seems suitable as an initial attempt. We can always re-adjust based on feedback and observation on how customers use this.

tcbuilder/cli/secboot.py
```diff
@@ -0,0 +1,239 @@
...
+
+    subparser.add_argument(
+        "--kernel-key-dir", dest="kernel_key_dir",
+        help="Kernel fitImage key directory path. If specified the U-Boot DTB and Control DTBs "
+             "will be updated with the public key before signing the bootloader. This directory "
+             "must contain a private key in .key format and a certificate with "
+             "the public key in .crt format, both with the same name.")
+
+    subparser.add_argument(
+        "--kernel-key-algo", dest="kernel_key_algo",
```
rborn-tx: Aren't we missing the kernel signing key name here? From the help text I'm guessing we're automatically determining the key name from the directory contents, but that's probably not a good idea because users might (should) have multiple keys in the directory.

In fact, thinking a bit ahead, we have intentions to allow users to easily rotate the kernel signing keys at some point and that could be achieved by adding multiple keys into the U-Boot DTB. So, in the future I'd expect us to extend the CLI to allow users to pass multiple pubkeys to `secboot sign-bootloader-hab`. Of course we don't need to do this now, but it would be good to ensure the currently CLI does not make things harder for those future additions.

Also, let me confirm, once we know the name (`<name>`) of the key, we would produce a DTB like below, right?

```
/ {
    ...
    signature {
        key-<name> {
            key-name-hint = "<name>";
            algo = "<algo>";
            ...
        };
    };
```

rborn-tx (commented):
@lucas-akira This looks really great! Congrats!

I've raised only one point to be discussed.

tcbuilder/cli/secboot.py
lucas-akira: > Also, let me confirm, once we know the name (\<name\>) of the key, we would produce a DTB like below, right?

Exactly, `--kernel-key-algo` would correspond to the value of the `algo` field in the signature node, and the key name would be part of the node name (`key-<name>`) and the value of `key-name-hint`.

> Aren't we missing the kernel signing key name here? From the help text I'm guessing we're automatically determining the key name from the directory contents, but that's probably not a good idea because users might (should) have multiple keys in the directory.
>
> In fact, thinking a bit ahead, we have intentions to allow users to easily rotate the kernel signing keys at some point and that could be achieved by adding multiple keys into the U-Boot DTB. So, in the future I'd expect us to extend the CLI to allow users to pass multiple pubkeys to `secboot sign-bootloader-hab`. Of course we don't need to do this now, but it would be good to ensure the currently CLI does not make things harder for those future additions.

Good points, I didn't think about the use case of having multiple keys in the directory. As we discussed earlier today @rborn-tx , I'll address these by doing the following changes:

- In `secboot sign-kernel`, key-related args will be prefixed with `kernel-` e.g. `--kernel-key-dir` instead of `--key-dir` as to make them identical to their counterparts in `sign-bootloader-hab`, which should make things easier when copying args from running one command after the other.

- In `secboot sign-kernel`, a new `--kernel-key` arg will be used to pass both the key name and the pair of algorithms, replacing `--key-name` and `--key-algo` e.g. 

```
--kernel-key 'name=prod;algo=sha256,rsa2048'
```

- `secboot sign-bootloader-hab` would have `--kernel-key` as well. For it this would facilitate a future feature of passing multiple pubkeys to the DTBs: With this implementation, the user would just repeat the argument if using more than one key, passing the required information for each key in one place:

```
# Possible feature to have in the future:
--kernel-key 'name=prod1;algo=sha256,rsa2048' --kernel-key 'name=prod2;algo=sha512,rsa2048' ...
```

jsrc27 (approved):
I gave it one final pass through. From my side everything looks good, no further comments.

I can merge it now if everything is settled upon.

lucas-akira:
In this last change I just improved some of the checks for the tests, and removed an extra blank line in `cli/secboot.py`.

I also forgot to mention that in the last push I changed in the Dockerfile the git clone source for U-Boot from `source.denx.de` to its GitHub mirror, as the former has had some connection instabilities lately.

jsrc27:
Any other changes, or are we good to merge here?

lucas-akira:
All tests in the latest CI pipeline of the GitLab mirror passed, so on my side the PR can be merged.

PR #68, opened 2025-07-01
deploy: Block deploying to incompatible device

LeoBelmont:
Related-to: TCB-530

### DIFF
--- a/tcbuilder/backend/deploy.py
+++ b/tcbuilder/backend/deploy.py
@@ -20,7 +20,7 @@ from gi.repository import Gio, OSTree
 
 from tcbuilder.backend import ostree
 from tcbuilder.backend.common import (get_rootfs_tarball, resolve_remote_host,
-                                      run_with_loading_animation)
+                                      run_with_loading_animation, run_command_without_sudo)
 from tcbuilder.backend.rforward import reverse_forward_tunnel, request_port_forward
 from tcbuilder.errors import TorizonCoreBuilderError, InvalidDataError
 from tezi.utils import find_rootfs_content
@@ -33,6 +33,7 @@ EXTRA_ROOTFS_SIZE_KB = 0
 # Value based on
 # https://github.com/torizon/meta-toradex-torizon/blob/953aacb8b3241ea26f98e853d1a1d4c8463636a4/recipes-images/images/torizon-core-common.inc#L11
 IMAGE_OVERHEAD_FACTOR = 2.3
+
 
 def create_sysroot(deploy_sysroot_dir):
     sysroot = OSTree.Sysroot.new(Gio.File.new_for_path(deploy_sysroot_dir))
@@ -174,8 +175,10 @@ def create_installed_versions(path, ref, branch):
         versioninfo[ref] = branch + "-" + ref
         json.dump(versioninfo, versionfile)
 
+
 def copy_tezi_image(src_tezi_dir, dst_tezi_dir):
     shutil.copytree(src_tezi_dir, dst_tezi_dir)
+
 
 def pack_rootfs_for_tezi(dst_sysroot_dir, output_dir):
     image_filename = get_rootfs_tarball(output_dir)
@@ -416,7 +419,8 @@ def deploy_raw_image(base_raw_img, src_sysroot_dir, src_ostree_archive_dir,
                               rootfs_size_kb, dst_sysroot_dir)
     log.info(f"Image {os.path.basename(output_raw_img)} created successfully!")
 
-def run_command_with_sudo(client, command, password):
+
+def run_command_with_sudo(client, command, password) -> dict[str, str]:
     stdin, stdout, stderr = client.exec_command("sudo -S -- " + command)
     stdin.write(f"{password}\n")
     stdin.flush()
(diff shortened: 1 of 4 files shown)


tcbuilder/backend/deploy.py
```diff
@@ -437,6 +441,43 @@ def run_command_with_sudo(client, command, password):
...
+        return False
+
+    image_machine = srcmeta["oe.machine"]
+    compatibility = image_machine == host_machine
+    if not compatibility:
+        log.warning(
+            f"Image for device '{image_machine}' is NOT compatible with machine '{host_machine}'"
+        )
+
+    return image_machine == host_machine
```
jsrc27: Slight suggestion. Could probably just return `compatibility` since the comparison was already done prior.

tcbuilder/backend/deploy.py
```diff
@@ -498,6 +539,13 @@ def deploy_ostree_remote(remote_host, remote_username, remote_password, remote_p
     forwarding_thread.daemon = True
     forwarding_thread.start()
 
+    is_compatible = is_ostree_compatible(srcmeta, client, remote_password)
+
+    if not is_compatible:
+        raise TorizonCoreBuilderError("Device is NOT compatible. Operation aborted.")
+
+    log.info("Device compatible with image.")
```
jsrc27: Maybe this could be in `log.debug` or something, I don't think we need to explicitly show anything on the default log level if there's nothing amiss.

jsrc27 (approved):
Overall looks good to me. Just left some minor non-critical feedback.

tcbuilder/backend/deploy.py
LeoBelmont: Good catch! Fixed it.

tcbuilder/backend/deploy.py
LeoBelmont: That does make it cleaner. Changed it :)

lucas-akira (approved):
LGTM

tcbuilder/backend/deploy.py
```diff
@@ -437,6 +441,43 @@ def run_command_with_sudo(client, command, password):
...
+        r"bash -c '. /etc/profile.d/machine.sh && echo $MACHINE'",
+        password,
+    )
+    host_machine = result.get("stdout", "").strip()
+
+    if not host_machine:
+        log.warning("machine.sh exists but MACHINE variable is not set.")
+        return False
+
+    image_machine = srcmeta["oe.machine"]
```
rborn-tx, line 475: Since you're using information from the OSTree metadata on the host (which is a good thing), I think a better approach would be to do the same on the device side too.

E.g., on a device you can:

```
torizon@verdin-am62-15133477:~$ ostree admin status
* torizon 3a9ba273a8fbdeefb63113f1868dd5403d555cbf051597aa85ee67cb6a1f2fee.0
    Version: 7.3.0-devel-20250630153904+build.0
    origin refspec: 3a9ba273a8fbdeefb63113f1868dd5403d555cbf051597aa85ee67cb6a1f2fee
  torizon 93b9280dfddb10c4dfaea3d9dc3b5507ea43eb6e0803c9ca841acede9d808901.0 (rollback)
    Version: 7.3.0-devel-20250628003908+build.0
    origin refspec: 93b9280dfddb10c4dfaea3d9dc3b5507ea43eb6e0803c9ca841acede9d808901

torizon@verdin-am62-15133477:~$ ostree show 3a9ba273a8fbdeefb63113f1868dd5403d555cbf051597aa85ee67cb6a1f2fee --print-metadata-key oe.machine
'verdin-am62'
```

This could be translated into a one-liner:

```
$ ostree show $(ostree admin status | sed -ne 's/^ *\* .* \([0-9a-f]\+\)\.[0-9]\+/\1/p') --print-metadata-key oe.machine
'verdin-am62'
```

and notice this command doesn't require `sudo`. It would be worth checking if the `--print-metadata-key` switch is available on `kirkstone` (Torizon OS 6) though.

tcbuilder/cli/deploy.py
```diff
@@ -21,6 +21,7 @@
 DEFAULT_OUTPUT_WIC_NAME = "tcb_common_torizon_os.wic"
 DEFAULT_OUTPUT_IMG_NAME = "tcb_common_torizon_os.img"
 
+
```
rborn-tx, line 24: General recommendation for the future: put formatting / aesthetic changes into a separate commit to simplify review.

rborn-tx (commented):
Overall looks good. I've added a suggestion though.

tcbuilder/backend/deploy.py
LeoBelmont: @rborn-tx Thank you for your thorough review! I like this method better and modified the PR to use it.

I also tested on Torizon OS 6 and it works with `--print-metadata-key`.

tcbuilder/cli/deploy.py
LeoBelmont: I will make sure to do that in future contributions.

tcbuilder/backend/deploy.py
rborn-tx: Awesome. Thanks!

rborn-tx (approved):
LGTM

PR #71, opened 2025-07-14
Fix local tests using get_tezi_images script

flavioaraujodex:
Last change broke the local tests. Also the CONFIGME section is mentioned in the current documentation, but was remove in the same commit.

This change add the CONFIGME section again and also a check with verbose messages regarding the required environment variables to run the script get_tezi_images.sh.


flavioaraujodex:
Has a commit not verified, accordingly with CONTRIBUTING all commits should be verified

PR #72, opened 2025-07-14
Fix local tests using get_tezi_images script

flavioaraujodex:
Last change broke the local tests. Also the CONFIGME section is mentioned in the current documentation, but was remove in the same commit.

This change add the CONFIGME section again and also a check with verbose messages regarding the required environment variables to run the script get_tezi_images.sh.

### DIFF
--- a/tests/integration/README
+++ b/tests/integration/README
@@ -34,12 +34,14 @@ $ touch $PWD/workdir/images/.images_downloaded
 Another option is to use get_tezi_images.sh script to automatically
 download the images based on a specific Torizon OS version.
 
-To use the script, open it and set the parameters to download the Tezi
-images (search for CONFIGME in the script).
-
-And then run the script:
+To use the script, you can run the script with the default values:
 
 $ ./get_tezi_images.sh
+
+If you want to change build type, machine or the yocto branch please set 
+the respectively enviroment variables, an example would be:
+
+$ TARGET_BUILD_TYPE=release YOCTO_BRANCH=kirkstone-6.x.y TCB_MACHINE=verdin-am62 ./get_tezi_images.sh
 
 Download Common Torizon OS WIC/raw images
 =====================================
@@ -57,7 +59,9 @@ $ touch $PWD/workdir/images/.raw_images_downloaded
 
 There's also the option to use get_raw_images.sh to automatically
 download specific Common Torizon versions according to user-defined
-parameters (CONFIGME options).
+parameters. To have this please set the enviroment variable TCVERSION:
+
+$ TCVERSION=6.4.0-common ./get_raw_images.sh
 
 Install a fresh Torizon OS image in a device
 ============================================
--- a/tests/integration/get_raw_images.sh
+++ b/tests/integration/get_raw_images.sh
@@ -13,8 +13,8 @@ OUTDIR="$PWD/workdir/images"
 TMPDIR="$OUTDIR/tmp"
 STAMP="$OUTDIR/.raw_images_downloaded"
 
-# CONFIGME: version
-TCVERSION="6.6.0-common"
+# default values
+TCVERSION="${TCVERSION:-6.6.0-common}"
 
 prepare() {
     mkdir -p $OUTDIR
--- a/tests/integration/get_tezi_images.sh
+++ b/tests/integration/get_tezi_images.sh
@@ -4,6 +4,11 @@
 OUTDIR="$PWD/workdir/images"
 STAMP="$OUTDIR/.images_downloaded"
 
+# default values
+TARGET_BUILD_TYPE="${TARGET_BUILD_TYPE:-nightly}"
+YOCTO_BRANCH="${YOCTO_BRANCH:-scarthgap-7.x.y}"
+TCB_MACHINE="${TCB_MACHINE:-colibri-imx6}"
+
 prepare() {
     rm -rf workdir/images
     mkdir -p "$OUTDIR"


jsrc27 (approved):
LGTM

lucas-akira (commented):
Changes look good to me. One small suggestion I'd make is to prefix both commit messages with `tests` to explicitly indicate that they affect a test-related file e.g.:

```
tests: Add CONFIGME section to help define local enviroment variables
``` 
```
tests: Add required enviroment variables check.
```

tests/integration/get_tezi_images.sh
rborn-tx, line 1: Honestly, the changes here look odd to me.

I don't think we want developers to modify any file before they can run the tests. Checking the README, I see that it says users can simply run `get_tezi_images.sh` without passing variables which was true in the past but changed to implicitly require a couple variables without even checking if those variables were set. In my understanding, that was the main mistake, introduced by this commit:

(0a79fe6) tests/integration: Refactor get_tezi_images to download latest versioned images

To solve it we could simply:

1. Force users to pass the variables, as it was done in this PR.
1. Set default values in case we don't want to force users to always pass the required variables. For this, we could add this to the beginning of the script:
   ```
   TARGET_BUILD_TYPE="${TARGET_BUILD_TYPE:-nightly}"
   YOCTO_BRANCH="${YOCTO_BRANCH:-scarthgap-7.x.y}"
   TCB_MACHINE="${TCB_MACHINE:-colibri-imx6}"
   ```

Either way, I think **we should update the README** to tell users that they can/must pass the above variables when running the script.

Now the question becomes what option we prefer from the two above (Cc @jsrc27, @lucas-akira).

tests/integration/get_tezi_images.sh
```diff
@@ -4,9 +4,29 @@
...
+#TCB_MACHINE="colibri-imx6"
+
 prepare() {
     rm -rf workdir/images
     mkdir -p "$OUTDIR"
+
+    if [ -z "$TARGET_BUILD_TYPE" ]; then
+        echo "TARGET_BUILD_TYPE is empty. Check the CONFIGME section in the script."
+        exit 1
+    fi
```
rborn-tx: If we go with option (1), we could be more user friendly and concise by doing:

Define a "usage" function:

```
usage() {
    echo "The following variables must be passed:"
    echo "..."
    exit 1
}
```

and use it inside the conditional:

```
if [ -z "$TARGET_BUILD_TYPE" ] || [ -z "$YOCTO_BRANCH" ] || [ -z "$TCB_MACHINE" ]; then
    usage
fi
```

If we go with option (2), these conditionals could even be dropped.

rborn-tx (changes requested):
I'm suggesting some changes whose details will depend on maintainers opinions...

rborn-tx:
@flavioaraujodex As a general recommendation, please always try to add a description to your commits (with what and why something was done, but not how) and follow the commit title format for the project.

tests/integration/get_tezi_images.sh
rborn-tx: Hmm, we should probably discuss this on a meeting since I see the CONFIGME text is used by `get_raw_images.sh` as well.

lucas-akira: I don't have a very strong preference, but I'd say option (1) as it forces the user right from the beginning to learn about the variables and how they affect the downloaded image. Ideally the test README would be updated with examples on possible values for each variable.

> Hmm, we should probably discuss this on a meeting since I see the CONFIGME text is used by get_raw_images.sh as well.

Yeah, `get_raw_images.sh` was heavily based on what `get_tezi_images.sh` looked like at the time, and that was before https://github.com/torizon/torizoncore-builder/commit/0a79fe62c64568242c2fc72fd7de729e49752a71 . It can be changed as well to be aligned with our decision here, though it probably needs a deeper review as a whole, possibly a complete rewrite.

jsrc27: For me personally I actually never really use the `get_tezi_images.sh` script and just download the images myself when testing. So I don't have any preference personally.

Honestly both options seem okay to me or at least neither really seems more suitable than the other.

rborn-tx: Just to leave my opinion here: I think we should go with (2) and update the documentation to tell users how to select a different machine/branch/build-type if they don't want the default.

The same approach could be followed on the other script that fetches the raw images, so that we could get rid of the CONFIGME blocks completely.

flavioaraujodex: I agree as well. So I'll perform the changes using the option (2) and also update the README accordingly.

The usage will be something as: TARGET_BUILD_TYPE=<some_value> YOCTO_BRANCH=<some_value> TCB_MACHINE=<some_value> script.sh.

Besides that keep a default configuration as well.

jsrc27 (approved):

tests/integration/README
```diff
@@ -34,13 +34,15 @@ $ touch $PWD/workdir/images/.images_downloaded
...
-
-And then run the script:
+To use the script, you can run the script with the default values:
 
 $ ./get_tezi_images.sh
 
+If you want to change build type, machine or the yocto branch please set 
+the respectively enviroment variables, an example would be:
+
+$ TARGET_BUILD_TYPE=release YOCTO_BRANCH=dunfell-5.x.y TCB_MACHINE=verdin-am62 ./get_tezi_images.sh
```
lucas-akira: Specifically this example wouldn't work because there isn't a Torizon OS/TorizonCore 5 (dunfell) version for the Verdin AM62. You could change the example to be something like:

```
$ TARGET_BUILD_TYPE=release YOCTO_BRANCH=kirkstone-6.x.y TCB_MACHINE=verdin-am62 ./get_tezi_images.sh
```

lucas-akira (commented):
One thing I noticed is that all changes introduced in the first two commits are reverted by the last one you added.

In this case, wouldn't it be better if you drop the first two commits and only keep the last one?

tests/integration/get_raw_images.sh
```diff
@@ -13,8 +13,9 @@ OUTDIR="$PWD/workdir/images"
 TMPDIR="$OUTDIR/tmp"
 STAMP="$OUTDIR/.raw_images_downloaded"
 
-# CONFIGME: version
-TCVERSION="6.6.0-common"
+# default values
+TCVERSION="${TCVERSION:-6.6.0-common}"
+TCB_MACHINE="${TCB_MACHINE:-colibri-imx6}"
```
lucas-akira: I just noticed this now, but the `get_raw_images.sh` script doesn't use `TCB_MACHINE`; rather, it downloads all images declared in the `IMAGES` variable, which is already set at the start of the script.

So I think you can just remove the `TCB_MACHINE` declaration line here, as it isn't used.

lucas-akira (approved):
Newest revision looks good to me

PR #73, opened 2025-07-14
Use all cores to store container bundle

oresk:
By adding the -T0 option to the xz compression command I was able to get my build time of the whole image down from 8 minutes to 4 minutes. 

Signed-off-by: Lovro Oreskovic <lovro@oreskovic.me>

### DIFF
--- a/tcbuilder/backend/bundle.py
+++ b/tcbuilder/backend/bundle.py
@@ -39,7 +39,7 @@ def get_compression_command(output_file):
     output_file_tar = output_file
     if output_file.endswith(".xz"):
         output_file_tar = output_file[:-3]
-        command = ["xz", "-3", "-z", output_file_tar]
+        command = ["xz", "-T0", "-3", "-z", output_file_tar]
     elif output_file.endswith(".gz"):
         output_file_tar = output_file[:-3]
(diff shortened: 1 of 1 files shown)


lucas-akira (approved):
Hi @oresk ,

Currently the TCB container is based on Debian Bullseye, which has xz 5.2.5.

According to the [xz man page](https://man.archlinux.org/man/xz.1#T), newer versions of xz have `-T0` enabled by default, while in xz 5.4.x and older the default is `-T1`, so that would explain the gain in performance you experienced.

The addition looks good to me, thanks for the PR.

oresk:
Thank you for the prompt response.

jsrc27 (approved):
LGTM

rborn-tx (approved):
LGTM

rborn-tx (changes requested):
@oresk Let me just ask you to write the commit message according to the pattern we follow in this repository. This could be just:

```
bundle: Use all cores to store container bundle

<your sign-off>
```

Since this is very simple, a body is not needed.

oresk:
This should do it. 
Let me know if any other changes are needed.

rborn-tx (approved):

PR #74, opened 2025-07-25
tests/integration: Fail job if any test fails

lucas-p-bernardes:
After JUnit integration in TCB, the test behavior changed: if the tests reach the end, the job is considered successful.
As a result, even if some TCB tests fail, the pipeline may still pass.

This can be confusing for maintainers and makes it unclear when a test has actually failed. This commit changes the behavior so that the job fails if any TCB test fails.


jsrc27 (approved):
LGTM

lucas-akira (approved):
LGTM

PR #76, opened 2025-07-25
gitlab-ci: Disable linter jobs on Jenkins pipelines

lucas-p-bernardes:
The pipelines triggered by Jenkins are purely for integration purposes, aiming to identify new bugs in the torizoncore-builder when new versions of Torizon are released. They do not contain any changes to the torizoncore-builder code. Therefore, since there are no code changes, it doesn't make sense to run the linter in these pipelines, as it would be just a waste of resources. This commit adds a rule that disable linter pipelines when the source is a trigger token.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -72,6 +72,10 @@ stages:
 
 lint-dockerfiles:
   stage: pre-build
+  rules:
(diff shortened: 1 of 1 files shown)


jsrc27 (approved):
LGTM

lucas-akira (approved):
LGTM

PR #77, opened 2025-07-25
tests/integration: Run the possible jobs in cloud runners

lucas-p-bernardes:
This commit updates the pipeline configuration to run jobs that don't require internal resources in cloud runners.

Jobs related to "torizoncore" and "aval" continue to run on internal runners:
- "torizoncore" jobs require access to the internal Artifactory server.
- "aval" jobs establish connections to modules over the internal network.

Also, add WSL tests using AVAL to run together with Linux tests in push pipelines. This was done because it wasn’t necessary to revalidate all the jobs on WSL, and several jobs ended up being duplicated in the previous approach of selecting a runner tag via environment variable and running the entire pipeline.
Moreover, this method of dynamically defining the runner tag through an environment variable is not supported by GitLab, which led to several issues when it became necessary to make the code more complex to support the Cloud runners that use specific tags.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,10 +10,6 @@ variables:
   RELEASE_TYPE:
     value: "official"
     description: "Type of pipeline to run. Allowed values: 'official', 'early-access' or 'internal'."
-  # Variables defined in schedule pipelines has more precedente than this defined in gitlab-ci.yml. So,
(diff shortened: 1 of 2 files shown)


jsrc27 (approved):

lucas-akira (approved):
LGTM

PR #85, opened 2025-08-21
gitlab-ci: Use customized GitLab SaaS runners for CI jobs

lucas-p-bernardes:
This change to the GitLab runners is a request from the Automation team, which is now adopting the GitLab runners as the default runners to be used by the project. A study was carried out to determine which GitLab runner is the best fit for TCB. In this study, a full AVAL pipeline was executed using each of GitLab’s SaaS-linux runners, from small up to 2xlarge.

The use of these customized SaaS runners aims to minimize the time of each stage, up to the point where the cost becomes higher than the performance gain. These changes are being made in this commit:

- default: use saas-linux-small-amd64, as it has the lowest cost factor.

- lint-python and build-amd64 jobs: use saas-linux-large-amd64, because the job time decreased significantly up to the large runner. Beyond the large runner, the costs increased significantly and did not justify the gain in performance.

Related-to: TCB-733

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -51,7 +51,7 @@ variables:
 
 default:
   tags:
-    - aws-toradex-large
+    - saas-linux-small-amd64
 
 services:
   - name: docker:dind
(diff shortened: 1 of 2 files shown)


jsrc27 (approved):
LGTM

lucas-p-bernardes:
Hey @lucas-akira @jsrc27 @rborn-tx, can we have a round of review here?

lucas-akira (approved):
LGTM. Nice details on the commit message!

PR #87, opened 2025-08-26
tcbuild.schema: Add machines to schema

LeoBelmont:
Related-to: TIE-1066, TIE-1068, TIE-1070


jsrc27 (approved):
LGTM

lucas-p-bernardes:
I added commits related to tests and XRay on top of Bornia’s commit, since his commit was a prerequisite for mine.
Pipeline on GitLab: https://gitlab.com/toradex/rd/torizon-core/torizoncore-builder/-/pipelines/2005295219

Related-to: TCB-725

tests/integration/run_all.sh
```diff
@@ -104,8 +104,9 @@ check_device_connection() {
         export DEVICE_HOSTNAME="$hostname"
         export DEVICE_MAC="$macAddress"
         export TCB_DEVICE="$localIpV4"
+        export TCB_MACHINE=$(echo "$hostname" | sed -E 's/-[0-9]+$//')
```
lucas-akira: In `run.sh` there's an implementation to get the machine name by connecting to the device and getting the value of `/etc/profile.d/machine.sh`, which I think is more robust than getting the machine name from the hostname. This is only executed if `TCB_DEVICE` is set, and it appears to override the value of `TCB_MACHINE` set by the user.

We could change `run.sh` so that users don't need to pass `TCB_MACHINE` when connecting to a device, as the above implementation would take care of that.

lucas-p-bernardes: Makes sense!
However, I don’t want to delays @LeoBelmont work right now, and that’s not the initial purpose of this Merge Request. For this reason, I decided to drop the commit, and we can create another MR for it in the future.

@lucas-akira To don’t lose this, could you please create a ticket explaining it?
This is the new pipeline after the drop in that commit: https://gitlab.com/toradex/rd/torizon-core/torizoncore-builder/-/pipelines/2009972395

lucas-akira: Sure, I'll create one.

rborn-tx (approved):
LGTM

PR #88, opened 2025-09-01
Maintenance in integration pipeline for smarc and verdin imx95 devices

lucas-p-bernardes:
Pipeline: https://gitlab.com/toradex/rd/torizon-core/torizoncore-builder/-/pipelines/2015216937


tests/integration/aval/aval-tests.yml
```diff
@@ -402,6 +438,15 @@ aval-verdin-imx8mpq-scarthgap-nightly:
...
 
+aval-verdin-imx95h-scarthgap-release:
+  extends: .aval-template-linux-all-machines
+  variables:
+    SOC_UDT: "verdin-imx95h"
+    YOCTO_BRANCH: "scarthgap-7.x.y"
+    TARGET_BUILD_TYPE: "release"
+    DELEGATION_CONFIG: "./aval/delegation_config_scarthgap.toml"
+    TCB_MACHINE: "verdin-imx95"
+
```
lucas-p-bernardes, line 449: I'm aware that this will fail until we have a release for verdin-imx95. When the release is published, this job will then start to run.
However, I prefer this to fail until the release happens rather than creating another MR to add it.
This is not a problem for the `early-access` tag release because the AVAL jobs can fail and the release of this tag will still proceed.

If it’s preferable to remove this, I can simply drop the commit.

lucas-p-bernardes: The expected error occurred in the pipeline: 
```
ERROR - An error occurred while processing device 356d0a93-2add-4fc7-80b8-744b419bbc7c: Couldn't parse the latest build for release with verdin-imx95
```
AVAL couldn't parse the latest release build because it doesn't exist yet.

jsrc27 (approved):
LGTM

tests/integration/aval/aval-tests.yml
lucas-akira: If this doesn't affect the weekly early-access release then I see no problems in including it in the pipeline.

lucas-akira (approved):
LGTM

rborn-tx (approved):
LGTM

PR #89, opened 2025-09-09
Refactor 'secboot' commands and add sign feature to 'build' command

lucas-akira:
Refactor sign-kernel and sign-bootloader-hab to affect the current unpacked image instead of asking for an input image. This way the secboot commands would follow the stateful workflow applied for other standalone commands, which is similar to this:

`images unpack` -> `secboot` commands -> `union` -> `deploy`

This workflow also guarantees that the re-signed kernel FIT is included in an OSTree commit and is properly tracked, which wasn't the case with the previous implementation of the sign-kernel command.

Also add the sign feature of the secboot commands above to the build command, covering the stateless workflow as well.

CI tests for the secboot commands were adapted for the refactoring, and new tests covering the new signing feature of the build command were added.

Related-to: TCB-731

### DIFF
--- a/tcbuilder/backend/deploy.py
+++ b/tcbuilder/backend/deploy.py
@@ -20,8 +20,10 @@ from gi.repository import Gio, OSTree
 
 from tcbuilder.backend import ostree
 from tcbuilder.backend.common import (get_rootfs_tarball, resolve_remote_host,
-                                      run_with_loading_animation, REMOTE_CMD_TIMEOUT)
+                                      run_with_loading_animation, REMOTE_CMD_TIMEOUT,
+                                      SECBOOT_ARTIFACTS_DIR)
 from tcbuilder.backend.rforward import reverse_forward_tunnel, request_port_forward
+from tcbuilder.backend.secboot import FUSE_CMD_TXT_NAME
 from tcbuilder.errors import TorizonCoreBuilderError, InvalidDataError
 from tezi.utils import find_rootfs_content
 # pylint: enable=wrong-import-position
@@ -279,6 +281,8 @@ def deploy_ostree_local(src_sysroot_dir, src_ostree_archive_dir,
     log.info("Copy files not under OSTree control from original deployment.")
     src_sysroot = ostree.load_sysroot(src_sysroot_dir)
     copy_files_from_old_sysroot(src_sysroot, sysroot)
+
+    return csumdeploy
 # pylint: enable=too-many-locals
 
 
@@ -289,12 +293,18 @@ def deploy_tezi_image(tezi_dir, src_sysroot_dir, src_ostree_archive_dir,
     Creates a new Toradex Easy Installer image with a OSTree deployment of the
     given OSTree reference.
     """
-    deploy_ostree_local(src_sysroot_dir, src_ostree_archive_dir, dst_sysroot_dir, ref)
+    commit = deploy_ostree_local(src_sysroot_dir, src_ostree_archive_dir, dst_sysroot_dir, ref)
 
     log.info("Packing rootfs...")
     copy_tezi_image(tezi_dir, output_dir)
     pack_rootfs_for_tezi(dst_sysroot_dir, output_dir)
     log.info("Packing rootfs done.")
+
+    # Check if there is a signed bootloader to be copied for this ref
+    commit_dir = os.path.join(SECBOOT_ARTIFACTS_DIR, commit)
+    if os.path.isdir(commit_dir):
+        log.info(f"Copying signed artifacts linked with {commit} to output image.")
+        copy_signed_artifacts(commit_dir, output_dir)
 
 
 def write_rootfs_to_raw_image(base_raw_img, output_raw_img, base_rootfs_partition, rootfs_label,
@@ -584,3 +594,28 @@ def deploy_ostree_remote(remote_host, remote_username, remote_password, remote_p
 
     ostree.serve_ostree_stop(http_server_thread)
 # pylint: enable=too-many-locals
+
+
+def copy_signed_artifacts(src_commit_dir, tezi_dir):
+    """
+    Copy artifacts signed by TCB to the Torizon OS image in tezi_dir, overwriting
+    existing content. These include the bootloader container, the text file with
+    fusing instructions, and the tarball with the signing files.
+
+    :param src_commit_dir: Commit directory where the artifacts are located
+    :param tezi_dir: Path of Torizon OS image in Toradex Easy Installer format (directory)
+    """
+
+    # Add the signed artifacts to the tezi image, overwriting existing files
+    shutil.copytree(src_commit_dir, tezi_dir, dirs_exist_ok=True)
+
+    fuse_cmd_txt = os.path.join(tezi_dir, FUSE_CMD_TXT_NAME)
+    if os.path.isfile(fuse_cmd_txt):
+        with open(fuse_cmd_txt, 'r') as fuse_file:
+            print()
+            print("# Fusing instructions to be executed in U-Boot:")
+            print()
+            print(fuse_file.read())
+            print(f"# You can recheck the instructions given above in {FUSE_CMD_TXT_NAME} "
+                  f"created in directory '{os.path.basename(tezi_dir)}'.")
+            print()
--- a/tcbuilder/backend/images.py
+++ b/tcbuilder/backend/images.py
@@ -25,8 +25,9 @@ import fabric
 from tcbuilder.backend.common import (get_rootfs_tarball, get_tar_compress_program_options,
                                       set_output_ownership, run_with_loading_animation,
                                       get_tezi_image_version, DEFAULT_RAW_ROOTFS_LABEL,
-                                      RAW_PROP_TO_ARGNAME)
+                                      RAW_PROP_TO_ARGNAME, SECBOOT_ARTIFACTS_DIR)
 from tcbuilder.backend import ostree
+from tcbuilder.backend.secboot import DEFAULT_TCB_SIGNING_FILES_TARNAME, BOOTLOADER_CONTAINER_NAME
 from tcbuilder.errors import (TorizonCoreBuilderError, InvalidArgumentError, InvalidStateError)
 from tezi.image import ImageConfig, DEFAULT_IMAGE_JSON_FILENAME
 from tezi.errors import TeziError
@@ -386,6 +387,7 @@ def import_local_image(image_dir_or_file, tezi_dir, src_sysroot_dir, src_ostree_
     metadata, _, _ = ostree.get_metadata_from_checksum(src_sysroot.repo(), csum)
 
     if os.path.exists(tezi_dir):
+        track_tezi_signed_files(tezi_dir, csum, metadata.get("oe.machine"))
         log.info("Unpacked OSTree from Toradex Easy Installer image:")
     else:
         log.info("Unpacked OSTree from WIC/raw image:")
@@ -564,4 +566,38 @@ def provision(input_dir, output_dir, shared_data, online_data, hibernated=False,
         raise
 
 
+def track_tezi_signed_files(tezi_dir, commit_hash, machine):
+    """
+    Check if input TEZI image has a tarfile with signing files and if so,
+    associate it and the bootloader container with the current OSTree commit of the image.
+    Otherwise, do nothing.
+
+    :param tezi_dir: Path of directory containing input TEZI image.
+    :param commit_hash: Current OSTree commit hash of the TEZI image.
+    :machine: Machine name compatible with the TEZI image
+    """
+
+    if not machine:
+        log.warning("Warning: Could not determine compatible machine for the image.")
+        log.warning("Secboot commands will not be usable with this image.")
+        log.warning("Proceeding anyway.")
+        return
+
+    tcb_signing_files_tar = os.path.join(tezi_dir, DEFAULT_TCB_SIGNING_FILES_TARNAME)
+    if os.path.isfile(tcb_signing_files_tar):
+
+        # Assume the signing feature for the machine is not supported if TCB doesn't have its
+        # entry in BOOTLOADER_CONTAINER_NAME
+        if machine not in BOOTLOADER_CONTAINER_NAME:
+            log.warning("Warning: secboot commands are not supported for this machine.")
+            return
+
+        log.info(f"Found {DEFAULT_TCB_SIGNING_FILES_TARNAME}. Linking it to {commit_hash}.")
+        commit_dir = os.path.join(SECBOOT_ARTIFACTS_DIR, commit_hash)
+        os.makedirs(commit_dir, exist_ok=True)
+        shutil.copy2(tcb_signing_files_tar, commit_dir)
+
+        log.info(f"Linking bootloader container to {commit_hash}.")
+        shutil.copy2(os.path.join(tezi_dir, BOOTLOADER_CONTAINER_NAME[machine]),
+                     commit_dir)
 # EOF
--- a/tcbuilder/backend/union.py
+++ b/tcbuilder/backend/union.py
@@ -1,9 +1,12 @@
 import datetime
 import logging
 import os
+import shutil
 
 from tcbuilder.backend import ostree
 from tcbuilder.backend.ostree import OSTREE_WHITEOUT_PREFIX, OSTREE_OPAQUE_WHITEOUT_NAME
+from tcbuilder.backend.secboot import SIGNED_BOOTLOADER_ARTIFACTS_DIR
+from tcbuilder.backend.common import SECBOOT_ARTIFACTS_DIR
 from tcbuilder.errors import TorizonCoreBuilderError
 
 # pylint: disable=wrong-import-order,wrong-import-position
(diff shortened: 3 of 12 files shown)


jsrc27:
Some questions/comments on the UX.

* Only the create local directory version of `deploy` will fully work correct? If someone were to use the remote version of `deploy` to send these changes via network, only the signed FIT image would be deployed since that's in the OSTree. Meaning any HAB signed bootloader won't be deployed remotely.
* When TCB unpacks an image it assigns the default OSTree commit with the ref "base". Using this users can actually deploy "base" if they for some reason want the input OSTree/image outputted. I noticed when signing the bootloader the signed binary gets deployed to the tezi directory in storage overwriting the original bootloader binary. This would mean if someone does like `images unpack` -> `secboot sign-bootloader-hab` -> `deploy --output-directory output base`. In this case wouldn't the output now have a signed bootloader instead of being identical to the input? Not saying this is incorrect, but it does change a behavior.

lucas-akira:
Those are valid points to make, I haven't thought of them.

> Only the create local directory version of deploy will fully work correct? If someone were to use the remote version of deploy to send these changes via network, only the signed FIT image would be deployed since that's in the OSTree. Meaning any HAB signed bootloader won't be deployed remotely.

Correct, since the bootloader container isn't being tracked by OSTree, an usual SSH deploy wouldn't send the updated bootloader to the device. So maybe it would be better to restrict re-signing features to a local directory deploy only.

> When TCB unpacks an image it assigns the default OSTree commit with the ref "base". Using this users can actually deploy "base" if they for some reason want the input OSTree/image outputted. I noticed when signing the bootloader the signed binary gets deployed to the tezi directory in storage overwriting the original bootloader binary. This would mean if someone does like images unpack -> secboot sign-bootloader-hab -> deploy --output-directory output base. In this case wouldn't the output now have a signed bootloader instead of being identical to the input? Not saying this is incorrect, but it does change a behavior.

That does look like a problem to me, as the changes done in `sign-bootloader-hab` should only be associated with a new OSTree commit created with `union`, if the command is to follow the same stateful workflow other commands do.

I can remove the bootloader container overwriting function call of the `sign-bootloader-hab` command, and only add it when doing a deployment to a local directory, provided that the correct OSTree ref is passed to `deploy`.

So in summary TCB would have to internally track the bootloader when running `union` (if applicable) and check if it needs to be replaced during a local `deploy` command.

rborn-tx:
> > Only the create local directory version of deploy will fully work correct? If someone were to use the remote version of deploy to send these changes via network, only the signed FIT image would be deployed since that's in the OSTree. Meaning any HAB signed bootloader won't be deployed remotely.

Idea for the future :slightly_smiling_face:: Break down `deploy` into `images deploy-{tdx,wic}` and an `ostree deploy-ssh` to make their goals clear and the parameters each one takes very specific.

> > In this case wouldn't the output now have a signed bootloader instead of being identical to the input? Not saying this is incorrect, but it does change a behavior.
> 
> That does look like a problem to me, as the changes done in `sign-bootloader-hab` should only be associated with a new OSTree commit created with `union`, if the command is to follow the same stateful workflow other commands do.
> 
> I can remove the bootloader container overwriting function call of the `sign-bootloader-hab` command, and only add it when doing a deployment to a local directory, provided that the correct OSTree ref is passed to `deploy`.
> 
> So in summary TCB would have to internally track the bootloader when running `union` (if applicable) and check if it needs to be replaced during a local `deploy` command.

Idea for a solution: since we are trying to pretend that the bootloader is part of OSTree, we could consider doing something like this:

- As part of `images unpack`, move the bootloader binary into a state directory representing the bootloader(s); to allow for multiple bootloaders, create numbered sub-directories (say `/storage/bootloader/<index>/`); so, the original bootloader binary would be moved into `/storage/bootloader/0/`); this could be done specifically for signed images.
- Whenever a new bootloader binary is produced by `sign-bootloader-hab` store it into the next index, `/storage/bootloader/{1,2,...}/`.
- When `union` is executed, store the latest index into a file, say `/bootloader/version-map.json`. This json file would contain a map `commit-hash -> bootloader index`. We could populate this file with index 0 when doing the initial `images unpack` as well.
- When `deploy` is executed for a given commit, grab the bootloader from the index corresponding to the commit (from the map file) and stow it into the output tezi image.

Notice that there might be many commits, each one relating to a different bootloader, thus the need for keeping the multiple versions of bootloaders somewhere (since they were not actually part of the commit, but we are pretending they were).

Does that make sense? Of course this is not so straightforward to implement :weary:.

lucas-akira:
> Idea for a solution: since we are trying to pretend that the bootloader is part of OSTree, we could consider doing > something like this:
>
>    - As part of images unpack, move the bootloader binary into a state directory representing the bootloader(s); to allow for multiple bootloaders, create numbered sub-directories (say /storage/bootloader/\<index\>/); so, the original bootloader binary would be moved into /storage/bootloader/0/); this could be done specifically for signed images.
>   - Whenever a new bootloader binary is produced by sign-bootloader-hab store it into the next index, /storage/bootloader/{1,2,...}/.
>   - When union is executed, store the latest index into a file, say /bootloader/version-map.json. This json file would contain a map commit-hash -> bootloader index. We could populate this file with index 0 when doing the initial images unpack as well.
>   - When deploy is executed for a given commit, grab the bootloader from the index corresponding to the commit (from the map file) and stow it into the output tezi image.
>
> Notice that there might be many commits, each one relating to a different bootloader, thus the need for keeping the multiple versions of bootloaders somewhere (since they were not actually part of the commit, but we are pretending they were).

I thought on a similar solution, but storing the bootloader container and the fuse commands text file in a directory named after the union branch or commit hash e.g. `/storage/bootloader/<union branch>` when running `union`. The `deploy` command would then look if there is a directory named after the OSTree ref and if so replace the bootloader of the output image.

@rborn-tx as we discussed in an earlier sync: 

- It's probably better to use the commit hash for the directory name instead of the branch, as we're sure the hashes uniquely identify each commit;
- TCB should start tracking the bootloader right after running `images unpack`;
- As the contents of the tarball with the signing files are updated when running the signing commands, the tarball should be tracked with OSTree commits just like we're trying to do with the bootloader.

lucas-akira:
Alright, I've made the suggested changes mentioned above.

Also `sign-kernel` now gets the kernel FIT directly from the deployed rootfs of the unpacked image (specifically in `/ostree/deploy/torizon/deploy/*/usr/lib/modules/*/vmlinuz`) instead of expecting it to be inside `tcb_signing_files.tar.gz`, which makes the command not dependent on this tarball.

On the topic of a remote/SSH deploy not sending the bootloader changes to a device, I'm thinking if we should just clarify during execution that it only applies changes on components tracked by OSTree.

Tagging @jsrc27 and @rborn-tx to do a new review.

tcbuilder/backend/images.py
```diff
@@ -566,4 +568,38 @@ def provision(input_dir, output_dir, shared_data, online_data, hibernated=False,
...
+    Otherwise, do nothing.
+
+    :param tezi_dir: Path of directory containing input TEZI image.
+    :param commit_hash: Current OSTree commit hash of the TEZI image.
+    :machine: Machine name compatible with the TEZI image
+    """
+
+    if not machine:
+        log.warning("Could not determine compatible machine for the image.")
+        log.warning("Assuming image doesn't support secure boot features.")
```
jsrc27: These are non-critical warning messages. Maybe they could be reworded to sound less "problematic". For example if a user knows nothing about secure boot and sees these messages they might thing something is abnormal or such.

Maybe something along the lines of "Secboot commands will not be usable with this image" or something to that nature. Basically just to be clear that nothing is immediately wrong except you won't be able to use X feature.

tcbuilder/backend/images.py
```diff
@@ -566,4 +568,38 @@ def provision(input_dir, output_dir, shared_data, online_data, hibernated=False,
...
+        return
+
+    tcb_signing_files_tar = os.path.join(tezi_dir, DEFAULT_TCB_SIGNING_FILES_TARNAME)
+    if os.path.isfile(tcb_signing_files_tar):
+
+        # Assume the signing feature for the machine is not supported if TCB doesn't have its
+        # entry in BOOTLOADER_CONTAINER_NAME
+        if machine not in BOOTLOADER_CONTAINER_NAME:
+            log.warning(f"Found {DEFAULT_TCB_SIGNING_FILES_TARNAME}, but TorizonCore Builder "
+                        f" doesn't support signing the bootloader of images for \"{machine}\".")
```
jsrc27: Same suggestion as the previous comment.

jsrc27 (commented):
Looks good overall, just some minor comments.

I think we still need to handle the case where someone does the remote SSH version of `deploy` right? Could just be a log message like if a signed bootloader is detected n storage and the user is trying to do an SSH deploy. Then TCB prints a message like "bootloader can't be deployed remotely, and must be flashed with Easy Installer" or something like that.

rborn-tx:
> I think we still need to handle the case where someone does the remote SSH version of `deploy` right? Could just be a log message like if a signed bootloader is detected n storage and the user is trying to do an SSH deploy. Then TCB prints a message like "bootloader can't be deployed remotely, and must be flashed with Easy Installer" or something like that.

@jsrc27 In theory this concern is valid for anything that is outside the control of OSTree, right? That would include the bootloader whether it's (re)signed or not, Docker images, provisioning data, fuse data (or the u-boot environment in general) and anything else that relies on the installer itself for copying stuff to the device in the future. Considering that, maybe we could have a generic message always shown saying that only OSTree-managed artifacts will be deployed? How about that?

jsrc27:
> In theory this concern is valid for anything that is outside the control of OSTree, right

Yes it is, I'm just highlighting it here because we have part of the `secboot` command that will work as expected with the SSH `deploy` and another part that will not. For the average user it will not be obvious at all why one works, and the other does not. Which could lead to increased confusion/questions compared to the other cases where SSH `deploy` does not work.

I'd be okay with any approach that helps set/clarify user expectations. Especially when the result is not what an average user would expect.

That said I'm not sure if an average user would know what TCB actions are "OSTree-managed" and which are not. Especially since it's a little opaque what TCB commands are leveraging OSTree under the hood.

rborn-tx:
> Yes it is, I'm just highlighting it here because we have part of the `secboot` command that will work as expected with the SSH `deploy` and another part that will not...

Sure, that's a valid concern.
 
> I'd be okay with any approach that helps set/clarify user expectations. Especially when the result is not what an average user would expect.
> 
> That said I'm not sure if an average user would know what TCB actions are "OSTree-managed" and which are not. Especially since it's a little opaque what TCB commands are leveraging OSTree under the hood.

We could think of a message that clarifies what that means. How about something along these lines:

```
WARNING: Beware that artifacts not managed by OSTree (e.g. bootloader, container images, platform provisioning data, fuse values, U-Boot environment) will not be deployed by this operation.
```

tcbuilder/backend/deploy.py
```diff
@@ -595,3 +606,28 @@ def deploy_ostree_remote(remote_host, remote_username, remote_password, remote_p
...
+    fuse_cmd_txt = os.path.join(tezi_dir, FUSE_CMD_TXT_NAME)
+    if os.path.isfile(fuse_cmd_txt):
+        with open(fuse_cmd_txt, 'r') as fuse_file:
+            print()
+            log.info("# Fusing instructions to be executed in U-Boot:")
+            print()
+            log.info(fuse_file.read())
+            log.info(f"# You can recheck the instructions given above in {FUSE_CMD_TXT_NAME} "
+                     f"created in directory '{os.path.basename(tezi_dir)}'.")
+            print()
```
rborn-tx, line 621: I'd suggest not mixing log messages (which may or may not go to the stdout depending on log level) with `print` (which goes to stdout).

If you want to go with the logging system, I'd suggest collapsing the newlines with the log message rather than calling `log.info("")` since this tends to be understood as a since log "entry", e.g.:

```
log.info("\n# Fusing instructions to be executed in U-Boot:\n")
```

Generally speaking it's not a great idea to add newlines to log messages, but we are already (ab)using the logging system in other places by assuming that the messages will be shown on the console, in which case the newlines make sense for readability.

tcbuilder/backend/deploy.py
```diff
@@ -289,13 +293,20 @@ def deploy_tezi_image(tezi_dir, src_sysroot_dir, src_ostree_archive_dir,
...
     copy_tezi_image(tezi_dir, output_dir)
     pack_rootfs_for_tezi(dst_sysroot_dir, output_dir)
     log.info("Packing rootfs done.")
 
+    # Check if there is a signed bootloader to be copied for this ref
+    commit_dir = os.path.join(SECBOOT_ARTIFACTS_DIR, commit)
+    if os.path.isdir(commit_dir):
+        log.info(f"Copying signed artifacts linked with {commit} to output image.")
+        copy_signed_artifacts(commit_dir, output_dir)
+
```
rborn-tx, line 308: Normally we use 2 empty lines between functions.

tcbuilder/backend/union.py
```diff
@@ -165,4 +170,44 @@ def union_changes(changes_dir, ostree_archive_dir, union_branch,
...
+    determined by BOOTLOADER_CONTAINER_NAME.
+
+    :param commit_hash: OSTree commit hash to be associated with the signed bootloader
+    :param ostree_repo: OSTree repo object
+    """
+
+    # Check if there's a signed bootloader in storage, and if so store it to be
+    # deployed with the new branch
+    src_bootloader_path = os.path.join(FLASH_BIN_SIGNING_DIR, FLASH_BIN)
+    if os.path.isfile(src_bootloader_path):
```
rborn-tx: @lucas-akira Notice here we are copying a machine specific file to the "hash-linked directory". Given that, let me toss out an idea for your consideration:

What if the `sign-bootloader-hab` command produced a directory in storage, say `/storage/bootloader_artifacts/` containing the files to be tracked by union? Then here, when handling `union` we would simply need to blindly copy that directory contents into `/storage/secboot_tracked_artifacts/<hash>`. And then, as already being done, `deploy` would again blindly copy the hash-linked directory into the output image.

Would that be possible? Or is there anything preventing it?

The advantage with that approach is that we would concentrate machine specific code all within the signing command(s) (which is/are already machine specific). This should help us later when adding support for other machines.

What do you think?

tcbuilder/cli/secboot.py
```diff
@@ -146,33 +162,15 @@ def init_parser(subparsers):
...
-        help="Force program output, overwriting any existing file/directory.")
-
-    subparser.add_argument(
-        "--kernel-key-dir", dest="kernel_key_dir",
+        dest="kernel_key_dir",
+        metavar="KERNEL_KEY_DIR",
         help="Kernel fitImage key directory path. This directory must contain a private key in "
-             ".key format",
-        required=True)
+             ".key format")
```
rborn-tx: Maybe we should be a bit more accurate here, e.g.:

```
This directory must contain a private key file in PEM format having the .key extension.
```

tcbuilder/cli/secboot.py
```diff
@@ -196,22 +194,9 @@ def init_parser(subparsers):
...
-
-    subparser.add_argument(
-        "--force", dest="force",
-        default=False, action="store_true",
-        help="Force program output, overwriting any existing file/directory.")
+        dest="cst_dir",
+        metavar="CST_DIR",
+        help="CST directory path.")
 
     subparser.add_argument(
```
rborn-tx, line 201: Same comment about file format/extension. Actually, here we could say that both the key and certificate would be in PEM format.

rborn-tx (changes requested):
@lucas-akira Great work so far!

I've added a couple of general ideas for you to consider and suggested some minor changes.

tcbuilder/backend/images.py
lucas-akira: Sure, I can reword the messages to be like that. But should the first one (the 'could not determine compatible machine' one) be changed as well? It tells the user what caused the warning, and I think the rewording could still work with it, like this:

```
        log.warning("Could not determine compatible machine for the image.")
        log.warning("Secboot commands will not be usable with this image.")
```

These messages would only appear if TCB doesn't find the `oe.machine` metadata in the OSTree commit of the input image, which I'd say is not likely to happen, unless something very significant changes in our Yocto builds.

tcbuilder/backend/deploy.py
lucas-akira: Yeah, I ended up putting 3 lines there by mistake. Thanks, I'll correct it.

tcbuilder/backend/union.py
lucas-akira: I think it's possible to do that, yeah. I thought about `deploy` being machine-agnostic by blindly copying the hash-linked directory, but it makes sense to do the same for the `union` command as well and leave all the machine-specific logic to `sign-bootloader-hab`. I'll apply your suggestion.

tcbuilder/cli/secboot.py
lucas-akira: I wasn't exactly sure how to phrase it, but your suggestion makes more sense. Thanks, I'll apply it.

tcbuilder/backend/deploy.py
lucas-akira: I added the `print()` statements to improve readability while avoiding having log messages with empty lines. 

Maybe these lines in particular would be better suited for a regular `print` rather than having them logged, as they're just instructions on what to do outside TCB and not something strictly related to TCB itself.

rborn-tx: Yes, since those are instructions to the user, it's probably better not to make them filter-able by the logging system. A regular print is probably the best choice.

tcbuilder/backend/images.py
jsrc27: Keeping the first line should be okay. Maybe add something like "Proceeding anyways" or something to that effect. Basically just to communicate "X was found but proceeding without error" kind of message.

tcbuilder/backend/images.py
lucas-akira: Here I was thinking that TCB could just inform the user that the `secboot` commands are not supported for the machine, and that it will skip the secure boot setup:

```
            log.warning("Warning: secboot commands are not supported for this machine.")
            log.warning("Skipping secure boot setup for image.")
```

Maybe I can add the 'skipping secure boot setup' message instead of the 'proceeding anyway' one for this and the previous message?

jsrc27: Hmm I'm trying to envision what will cause users the least amount of confusion/worry. My concern is that in the past users were presented with warning logs that weren't actual problematic in any way. But, they got spooked by the log message and questioned our support about the meaning of the logs, without actually confirming whether the operation they were executing worked or not.

"Skipping secure boot setup" I could imagine someone might see this as abnormal and they may think that something is wrong.

Maybe just keep the one line "Warning: secboot commands are not supported for this machine." I think this is plenty clear enough on it's own. It states specifically what will not work and doesn't give the impression that there is an issue with the current command.

lucas-akira: > Maybe just keep the one line "Warning: secboot commands are not supported for this machine." I think this is plenty clear enough on it's own. It states specifically what will not work and doesn't give the impression that there is an issue with the current command.

Right, the first line should be enough to get the point across. I'll apply it.

tcbuilder/backend/images.py
lucas-akira: OK, I'll add the "proceeding anyway" message as well.

jsrc27 (approved):
Latest revision looks okay on my side.

tcbuilder/backend/union.py
rborn-tx: @lucas-akira This is ongoing, right?

lucas-akira: Latest revision has this change implemented. Sorry, I forgot to update this thread.

I ended up calling the directory for the signed bootloader artifacts `signed_bootloader_artifacts`, located in `/storage`.

rborn-tx (approved):
Awesome work!

rborn-tx:
@lucas-akira It seems this will need a rebase...

lucas-akira:
There were some minor conflicts with some import statements in `backend/common.py` and `backend/deploy.py`. I've cleared them up.

PR #92, opened 2025-09-12
aval-tests: Run only one AVAL job for the same machine at the same time

lucas-p-bernardes:
Previously, all AVAL jobs ran in parallel. For each machine, there are usually three jobs:
- Kirkstone, release
- Scarthgap, release
- Scarthgap, nightly

However, in practice, there is usually only one machine of each type in AVAL. As a result, one of these three jobs will run while the other two wait to acquire the machine, retrying and occupying runners, which can even result in a timeout if the machine cannot be locked. . This is the reason for the error "Wasn't able to lock the device <uuid> after 80 attempts".

This commit create a dependency chain between these jobs so that each job starts only after the previous one finishes, ensuring the machine is not locked for other job in the same pipeline.

Related-to: TCB-737

### DIFF
--- a/tests/integration/aval/aval-tests.yml
+++ b/tests/integration/aval/aval-tests.yml
@@ -107,6 +107,7 @@ build-aval-docker:
 
 aval-apalis-imx6q-kirkstone-release:
   extends: .aval-template-linux-all-machines
+  resource_group: apalis-imx6q
   variables:
     SOC_UDT: "apalis-imx6q"
     YOCTO_BRANCH: "kirkstone-6.x.y"
@@ -116,6 +117,7 @@ aval-apalis-imx6q-kirkstone-release:
 
 aval-apalis-imx6q-scarthgap-release:
   extends: .aval-template-linux-default-machines
+  resource_group: apalis-imx6q
   variables:
     SOC_UDT: "apalis-imx6q"
     YOCTO_BRANCH: "scarthgap-7.x.y"
@@ -125,6 +127,7 @@ aval-apalis-imx6q-scarthgap-release:
 
 aval-apalis-imx6q-scarthgap-nightly:
   extends: .aval-template-linux-all-machines
+  resource_group: apalis-imx6q
   variables:
     SOC_UDT: "apalis-imx6q"
     YOCTO_BRANCH: "scarthgap-7.x.y"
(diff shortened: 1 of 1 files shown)


lucas-p-bernardes:
These pushs was me improving the commit message, but I didn't change the code.

lucas-p-bernardes:
Pipeline triggered by GitLab: https://gitlab.com/toradex/rd/torizon-core/torizoncore-builder/-/pipelines/2037097863
Mock of pipeline triggered by Jenkins: https://gitlab.com/toradex/rd/torizon-core/torizoncore-builder/-/pipelines/2037099815

jsrc27 (approved):
LGTM

rborn-tx:
@lucasbernardestoradex Have you considered using https://docs.gitlab.com/ci/yaml/#resource_group. It seems to fit perfectly with this use case and it avoids the need for manually defining the job order...

lucas-p-bernardes:
> @lucasbernardestoradex Have you considered using https://docs.gitlab.com/ci/yaml/#resource_group. It seems to fit perfectly with this use case and it avoids the need for manually defining the job order...

No, I never used that before. Indeed seems to fit exactly to how AVAL works, because this feature block the resource even in different pipelines. I implemented that in the last push I did.
New pipeline: https://gitlab.com/toradex/rd/torizon-core/torizoncore-builder/-/pipelines/2043618313
Thank you for the idea @rborn-tx .

For the future: There's an open issue in GitLab to have this feature across projects in the same group, in our case the group is `rd`. This would be perfect for AVAL.
https://gitlab.com/gitlab-org/gitlab/-/issues/122010

rborn-tx:
> No, I never used that before. Indeed seems to fit exactly to how AVAL works, because this feature block the resource even in different pipelines. I implemented that in the last push I did. New pipeline: https://gitlab.com/toradex/rd/torizon-core/torizoncore-builder/-/pipelines/2043618313 Thank you for the idea @rborn-tx .

Looks a lot cleaner!

> For the future: There's an open issue in GitLab to have this feature across projects in the same group, in our case the group is `rd`. This would be perfect for AVAL. https://gitlab.com/gitlab-org/gitlab/-/issues/122010

That would be a nice feature!

rborn-tx (approved):

lucas-akira (approved):
LGTM as well

PR #95, opened 2025-09-19
tests/integration: Fix platform tests failing due to pipeline parallelism

lucas-p-bernardes:
According to commit 08ed2e049255d9fb588b8f3493121169fffff1b6, when multiple executions of the "platform tests" occurred at roughly the same time — specifically during the push of a generic package to the Cloud servers — one execution would succeed while the other failed, seemingly at random. That commit introduced a dependency chain between jobs using GitLab CI’s "needs:" keyword. However, this approach does not prevent jobs from different pipelines from running simultaneously.

This commit switches to using GitLab’s "resource_group", which ensures that these jobs never run at the same time, even across different pipelines, since the "resource_group" is enforced at the project level.

Related-to: TCB-754

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -203,6 +203,7 @@ build-arm64:
 
 test-host-tc6-platform-release:
   extends: .test-base
+  resource_group: platform
   stage: test
   variables:
     YOCTO_BRANCH: "kirkstone-6.x.y"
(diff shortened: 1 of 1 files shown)


jsrc27 (approved):
LGTM

lucas-akira (approved):
I agree that using `resource_group` is a better alternative than creating a dependency chain for the jobs. LGTM.

PR #96, opened 2025-09-23
isolate, deploy: Use Fabric library to run shell commands remotely

lucas-akira:
This fixes an IO desynchronization issue when sending remote sudo commands that could happen if the device has Torizon OS 7.3.0 or higher installed: the password would be sent to the device before sudo is ready to receive it, which would cause sudo to hang waiting for the password until its default timeout (5 minutes) gets reached and fail.

This wasn't reproducible on previous Torizon OS 7 versions. 7.3.0 introduced a newer version of sudo (1.9.17p1), which could explain this difference in behavior: probably this version takes a bit longer to show the password prompt, just enough for the issue to show up.

One way avoid this problem is to send shell commands remotely using a library designed to do this task, like Fabric. It can detect the sudo password prompt and auto-respond to it accordingly, which is a more robust method than the previous one.

-----

Also included is a fix to `isolate` when the device `/etc` has files or directories with leading or trailing whitespaces in their names. This was found in our CI pipeline while testing the fix above.

Closes TCB-729

### DIFF
--- a/requirements_debian.txt
+++ b/requirements_debian.txt
@@ -5,3 +5,4 @@ python-dateutil==2.8.2
 pyyaml==5.3.1
 requests-oauthlib==1.3.0
 debugpy==1.8.7
+fabric==3.2.2
--- a/tcbuilder/backend/common.py
+++ b/tcbuilder/backend/common.py
@@ -65,6 +65,8 @@ RAW_PROP_DEFAULTS = {
     "raw_rootfs_label" : DEFAULT_RAW_ROOTFS_LABEL
 }
 
+REMOTE_CMD_TIMEOUT = 30
+
 # Based on this solution: https://stackoverflow.com/a/50690347
 # Usage of Event object to stop thread was based on:
 # https://www.pythontutorial.net/python-concurrency/python-stop-thread/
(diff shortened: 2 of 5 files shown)


jsrc27:
Looks good on my side for the most part.

Just one question. Have you tested this with older versions of Torizon OS? From before we had the new `sudo` version, especially on Torizon 6.

I assume there would be no issue since this seems pretty agnostic from those details, but just wanted to be sure.

lucas-akira:
I just tested both `isolate` and `deploy` with Torizon OS 6.8.0 and 7.2.0, and they worked as expected.

jsrc27 (approved):

rborn-tx:
@lucas-akira Nice work!

Question:

Have you considered changing `backend/images.py::get_device_info()` to also use Fabric? Since you already did lots of work, converting that function seems relatively simple. If we did that, we could drop `paramiko` rather than have two libraries being used with the same purpose, which would be good for maintenance and consistency.

lucas-akira:
Yeah, I considered replacing Paramiko with Fabric in `backend/images.py` as well, but since it isn't explicitly sending any shell commands to be executed on the device side like with `isolate` and `deploy`, and there are no reported issues with the current implementation, I thought that altering it may be out of scope for this PR.

But you've raised a good point: it would be better for code maintenance and consistency using only one library to handle SSH connections to remote devices.

Fabric does use Paramiko under the hood, so TorizonCore Builder would still technically use it, but this doesn't really matter for code maintenance/consistency.

@rborn-tx I'm inclined to follow your suggestion. Any thoughts @jsrc27 ?

jsrc27:
> Any thoughts

I don't have any strong opinions on this as long as nothing changes for the user.

rborn-tx (approved):
Awesome!

PR #97, opened 2025-09-30
Throw error when running unsupported features for kernel in FIT format

lucas-akira:
When running certain commands/features, TorizonCore Builder will now first check if the unpacked image has the kernel in FIT format, and abort execution if the current implementation doesn't support it.

This applies to dt, dto, kernel and splash commands, as well as their respective features in the build command.

Related-to: TCB-755

Also, increase the timeout for remote commands sent via SSH to 90 seconds. Some tests in recent pipeline executions were failing early due to the command timeout being reached.

### DIFF
--- a/tcbuilder/backend/common.py
+++ b/tcbuilder/backend/common.py
@@ -844,38 +844,6 @@ def check_if_file_exists(filename, directory):
     return file_list[0]
 
 
-def get_unpacked_uenv_txt_vars(storage_dir, var_list):
-    """
-    Get the values of specific U-Boot variables declared in uEnv.txt in a
-    Torizon OS image unpacked in storage_dir
-
-    :param storage_dir: Path of the unpacked image. The rootfs dir is assumed to be named 'sysroot'
-    :param var_list: List of variables to get the values
-    :returns: Dictionary with the values for each variable in var_list. If an entry is not in
-              uEnv.txt, its value will be None
-    """
-
-    result = {}
-    rootfs_dir = os.path.join(storage_dir, "sysroot")
-    uenv_txt_path = glob.glob(os.path.join(rootfs_dir, "boot/loader*/uEnv.txt"))[0]
-
-    if not os.path.isfile(uenv_txt_path):
-        raise FileContentMissing("Couldn't find uEnv.txt in unpacked image rootfs! Aborting.")
-
-    with open(uenv_txt_path, 'r') as uenv_txt_file:
-        uenv_txt_str = uenv_txt_file.read()
-
-    for var in var_list:
-        match = re.search(r"^" + var + r"=(.*)", uenv_txt_str, re.MULTILINE)
-        if match:
-            result[var] = str(match.group(1))
-        else:
-            log.warning(f'"{var}=" variable not found in uEnv.txt.')
-            result[var] = None
-
-    return result
-
-
 def is_file_type_dtb(file_path):
     """
     Check if the file located in file_path is in DTB format by looking at the first bytes
--- a/tcbuilder/backend/common.py
+++ b/tcbuilder/backend/common.py
@@ -63,6 +63,14 @@ RAW_PROP_DEFAULTS = {
     "raw_rootfs_label" : DEFAULT_RAW_ROOTFS_LABEL
 }
 
+#  Hex value taken from the Devicetree Specification available at:
+#  https://devicetree-specification.readthedocs.io/en/stable/flattened-format.html
+FDT_HEADER_MAGIC = "d00dfeed"
+
+OSTREE_KERNEL_FILENAME = "vmlinuz"
+
+OSTREE_KERNEL_DEPLOY_PATH = "ostree/deploy/torizon/deploy/{csum}/usr/lib/modules/{kver}/"
+
 REMOTE_CMD_TIMEOUT = 30
 
 SECBOOT_ARTIFACTS_DIR = "/storage/secboot_tracked_artifacts"
@@ -866,3 +874,96 @@ def get_unpacked_uenv_txt_vars(storage_dir, var_list):
             result[var] = None
 
     return result
+
+
+def is_file_type_dtb(file_path):
+    """
+    Check if the file located in file_path is in DTB format by looking at the first bytes
+    of the file, which must be the value FDT_HEADER_MAGIC.
+    This can be used to quickly check if a binary is a Device Tree Blob or a Flattened Image
+    Tree (FIT) file.
+
+    :param file_path: Path of the file to be checked.
+    :returns: True if the first bytes of the file match FDT_HEADER_MAGIC, False otherwise.
+    """
+
+    expected_value = bytes.fromhex(FDT_HEADER_MAGIC)
+    value_length = len(expected_value)
+
+    with open(file_path, "rb") as _file:
+        found_value = _file.read(value_length)
+
+    return found_value == expected_value
+
+
+def is_file_type_fit(file_path):
+    """
+    Check if the binary located in file_path is a valid Flattened Image Tree (FIT) file by looking
+    if it has the required nodes and properties at the root node. Raises an error if the file is in
+    FDT format but doesn't have the required FIT properties or nodes.
+
+    :param file_path: Path of the file to be checked.
+    :returns: True if the file has the required properties and nodes,
+              False if the file is not in FDT format.
+    """
+
+    # Values taken from 'Flattened Image Tree (FIT) Format' in the U-Boot v2024.07 Documentation:
+    # https://docs.u-boot.org/en/v2024.07/usage/fit/source_file_format.html
+    required_props = ["timestamp"]
+    required_nodes = ["images", "configurations"]
+
+    if not is_file_type_dtb(file_path):
+        return False
+
+    try:
+        output = subprocess.run(["fdtget", file_path, "/", "-p"],
+                                text=True, check=True, stdout=subprocess.PIPE)
+
+        # Check if file has the required properties in the root node
+        if not all(prop in output.stdout for prop in required_props):
+            raise InvalidDataError("File is in FDT format but doesn't have the required FIT "
+                                   "properties. Aborting.")
+
+        output = subprocess.run(["fdtget", file_path, "/", "-l"],
+                                text=True, check=True, stdout=subprocess.PIPE)
+
+        # Check if file has the required nodes in the root node
+        if not all(node in output.stdout for node in required_nodes):
+            raise InvalidDataError("File is in FDT format but doesn't have the required FIT "
+                                   "nodes. Aborting.")
+
+    except subprocess.CalledProcessError as exc:
+        raise TorizonCoreBuilderError(f"Error running fdtget: {exc.output.strip()}")
+
+    return True
+
+
+def find_kernel_in_sysroot(storage_dir):
+    """
+    Find kernel binary path in the unpacked image sysroot. Raises an error if it cannot find it.
+
+    :param storage_dir: Path of the unpacked image. The rootfs dir is assumed to be named 'sysroot'
+    :returns: String with absolute path of the kernel binary inside sysroot.
+    """
+
+    sysroot_path = os.path.join(storage_dir, "sysroot")
+
+    # As the kernel path is composed of directories named after the deployment commit and the
+    # kernel version, get them via OSTree metadata.
+
+    sysroot_obj = ostree.load_sysroot(sysroot_path)
+
+    csum, _ = ostree.get_deployment_info_from_sysroot(sysroot_obj)
+    kernel_version = ostree.get_kernel_version(sysroot_obj.repo(), csum)
+
+    # Add deployment index part to checksum string, which is 0 for a new deployment
+    csum += ".0"
+
+    kernel_path = OSTREE_KERNEL_DEPLOY_PATH.format(csum=csum, kver=kernel_version)
+
+    kernel_path = os.path.join(sysroot_path, kernel_path, OSTREE_KERNEL_FILENAME)
+
+    if not os.path.isfile(kernel_path):
+        raise PathNotExistError("Kernel not found in unpacked rootfs. Aborting.")
+
+    return kernel_path
--- a/tcbuilder/backend/common.py
+++ b/tcbuilder/backend/common.py
@@ -71,7 +71,7 @@ OSTREE_KERNEL_FILENAME = "vmlinuz"
 
 OSTREE_KERNEL_DEPLOY_PATH = "ostree/deploy/torizon/deploy/{csum}/usr/lib/modules/{kver}/"
 
-REMOTE_CMD_TIMEOUT = 30
+REMOTE_CMD_TIMEOUT = 90
 
 SECBOOT_ARTIFACTS_DIR = "/storage/secboot_tracked_artifacts"
 
--- a/tcbuilder/backend/secboot.py
+++ b/tcbuilder/backend/secboot.py
@@ -11,8 +11,8 @@ import subprocess
 
 from tcbuilder.backend.common import \
     (set_output_ownership, get_tar_compress_program_options, is_tcb_container_64bit,
-     check_if_file_exists, get_unpacked_uenv_txt_vars, get_tezi_image_version,
-     OSTREE_KERNEL_DEPLOY_PATH, OSTREE_KERNEL_FILENAME)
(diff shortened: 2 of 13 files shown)


tcbuilder/backend/common.py
```diff
@@ -866,3 +874,61 @@ def get_unpacked_uenv_txt_vars(storage_dir, var_list):
...
+    if not is_file_type_dtb(file_path):
+        return False
+
+    try:
+        output = subprocess.run(["fdtget", file_path, "/", "-p"],
+                                text=True, check=True, stdout=subprocess.PIPE)
+
+        # Check if file has the required properties in the root node
+        if not all(prop in output.stdout for prop in required_props):
+            return False
```
rborn-tx: Shouldn't this throw an error instead? I mean, at this point we know the file is an FDT, but it could be an actual device-tree blob (DTB), an image-tree blob (ITB) or something unrecognized. The DTB and the unrecognized cases should probably be considered as errors rather than handled as non-FIT by the caller, right?

tcbuilder/cli/dt.py
```diff
@@ -29,8 +34,14 @@ def do_dt_status(args):
...
     if unpacked_image_type(args.storage_directory) == "raw":
-        raise InvalidDataError("dt commands are not supported for WIC/raw images. "
-                               "Aborting.")
+        raise InvalidDataError("Command not supported for WIC/raw images. Aborting.")
+
+    unpacked_kernel_path = check_if_file_exists(OSTREE_KERNEL_FILENAME,
+                                                os.path.join(args.storage_directory, "sysroot",
+                                                             OSTREE_KERNEL_DEPLOY_PATH))
+    if is_file_type_fit(unpacked_kernel_path):
+        raise FeatureNotImplementedError("Command not supported for kernel in FIT image format. "
```
rborn-tx: Since FIT = "Flattened **Image** Tree", how about writing "FIT format" rather than "FIT image format" (here and in the other occurrences)?

tcbuilder/cli/dto.py
```diff
@@ -50,8 +52,15 @@ def dto_apply(dtos_path, dtb_path, include_dirs, storage_dir,
...
     images_unpack_executed(storage_dir)
     if unpacked_image_type(storage_dir) == "raw":
-        raise InvalidDataError("dto commands are not supported for WIC/raw images. "
-                               "Aborting.")
+        raise InvalidDataError("Device tree overlay customization is not supported for WIC/raw "
+                               "images. Aborting.")
+
+    unpacked_kernel_path = check_if_file_exists(OSTREE_KERNEL_FILENAME,
+                                                os.path.join(storage_dir, "sysroot",
+                                                             OSTREE_KERNEL_DEPLOY_PATH))
```
rborn-tx: How about defining a `find_kernel_in_sysroot(storage_dir)` to encapsulate this logic?

Maybe we could even avoid the need for having a `check_if_file_exists()` at all since the globbing performed by that function seems a bit specific to this use case of finding the kernel.

tcbuilder/cli/kernel.py
```diff
@@ -224,9 +248,17 @@ def do_kernel_clear_custom_args(args):
 
     images_unpack_executed(args.storage_directory)
     if unpacked_image_type(args.storage_directory) == "raw":
-        raise InvalidDataError("Kernel commands are not supported for WIC/raw images. "
+        raise InvalidDataError("Kernel customization is not supported for WIC/raw images. "
                                "Aborting.")
 
+
```
rborn-tx: Extraneous empty line.

rborn-tx (commented):
@lucas-akira Raise some questions for your consideration.

tcbuilder/cli/dt.py
lucas-akira: Hmm sounds good. I'll change it.

tcbuilder/backend/common.py
```diff
@@ -63,7 +63,15 @@
...
-REMOTE_CMD_TIMEOUT = 30
+#  Hex value taken from the Devicetree Specification available at:
+#  https://devicetree-specification.readthedocs.io/en/stable/flattened-format.html
+FDT_HEADER_MAGIC = "d00dfeed"
+
+OSTREE_KERNEL_FILENAME = "vmlinuz"
+
+OSTREE_KERNEL_DEPLOY_PATH = "ostree/deploy/torizon/deploy/*/usr/lib/modules/*/"
+
+REMOTE_CMD_TIMEOUT = 90
```
jsrc27, line 74: What is the reason for the increased timeout? Was the previous one causing issues by being too short?

tcbuilder/cli/dt.py
jsrc27: This is with regards to all the `FeatureNotImplementedError` additions in this PR.

Wouldn't it be more efficent to just have the same generic `Command not supported for kernel in FIT image format. Aborting.` message for everything here.

Instead of having slightly different messages spread across a dozen points in the code.

jsrc27 (commented):
Just left some minor comments.

tcbuilder/backend/common.py
lucas-akira: Yeah, some device tests in our pipeline executions were failing due to the command timeout being reached, so I increased it.

tcbuilder/cli/dt.py
rborn-tx: Let me chime in to ask: if the message is generic, in case the user is using the "build" command, would it be clear what exactly is not supported?

lucas-akira: I changed the wording according to whether the feature could used in the `build` command as well.: specifically I avoided using the word `command` in this case.

But I can see that it could be easier for maintenance if it had the same generic message for all cases. Maybe it could say 'feature' instead of 'command' like `Feature not supported for kernel in FIT format. Aborting.`

rborn-tx: Let me confirm: if a user tries to set a device-tree through the "build" command, would they be able to understand from the command output that setting the device-tree is not yet supported?

Let's keep in mind that these error messages are just stopgap measures for the short-term, until we implement the features. So, maintenance shouldn't be much of a concern, at least in my view.

lucas-akira: Well, the `build` command has some info logs before entering each subsection that more or less explain what it will do. Assuming the generic error messages are being used, in some cases the `build` logs are clear on what is not supported, like with the splash screen:

```
$ torizoncore-builder build
Building image as per configuration file 'tcbuild.yaml'...

=>> Handling input section
Unpacking Toradex Easy Installer image.
[...]
=>> Handling customization section

=> Setting splash screen
Error: Feature not supported for kernel in FIT format. Aborting.
```

Setting a custom device tree is also clear I'd say:

```
$ torizoncore-builder build
Building image as per configuration file 'tcbuild.yaml'...

=>> Handling input section
Unpacking Toradex Easy Installer image.
[...]
=>> Handling customization section

=> Handling device-tree subsection

=> Selecting custom device-tree 'device-trees/dts-arm32/imx6ull-colibri-eval-v3.dts'
Error: Feature not supported for kernel in FIT format. Aborting.
```

But there are cases where it isn't clear, like when trying to remove an overlay, where it stops at the `device-tree` subsection without explicitly saying that it is about a feature related to overlays:

```
$ torizoncore-builder build
Building image as per configuration file 'tcbuild.yaml'...

=>> Handling input section
Unpacking Toradex Easy Installer image.
[...]
=>> Handling customization section

=> Handling device-tree subsection
Error: Feature not supported for kernel in FIT format. Aborting.
```

tcbuilder/backend/common.py
lucas-akira: Right, it makes sense. Now that you said it, I can't imagine a use case for this function if the binary being checked is in FDT format but isn't a FIT file.

This function wouldn't be used at all if checking a non-FIT FDT file anyway. I'll follow your suggestion.

tcbuilder/cli/dt.py
rborn-tx: @lucas-akira Thanks for checking. Well, in that case I'd prefer more specific messages as you did originally.

@jsrc27 What do you think?

jsrc27: I don't have a strong opinion. I just pointed it out cause it looked odd that all the messages were 95% the same.

tcbuilder/cli/dto.py
lucas-akira: Right, having the function would be a cleaner implementation.

> Maybe we could even avoid the need for having a check_if_file_exists() at all since the globbing performed by that function seems a bit specific to this use case of finding the kernel.

Thinking now, it's probably better to form the kernel path in sysroot by getting the commit and the kernel version from the OSTree metadata of the unpacked image. I'll use this logic for the new function.

tcbuilder/cli/kernel.py
lucas-akira: Thanks, fixed.

tcbuilder/cli/dt.py
lucas-akira: Done!

lucas-akira:
FYI I still need to update some tests to consider the case where the kernel in FIT format is unsupported. After that I'll ask for a new round of review @jsrc27 @rborn-tx .

lucas-akira:
Alright, I've updated the tests and I've updated the kernel FIT check in `secboot` to use the `is_file_type_fit` function

I've also removed the dependency of the `secboot` tests to a separately provided signed OS image: the test script will now check beforehand if the provided input image has its kernel in FIT format and if it has signing artifacts, and skip or run/test cases accordingly.

tcbuilder/cli/dt.py
lucas-akira: I'll keep the messages as is then.

tcbuilder/backend/common.py
```diff
@@ -836,33 +844,94 @@ def check_if_file_exists(filename, directory):
...
+
+    csum, _ = ostree.get_deployment_info_from_sysroot(sysroot_obj)
+    kernel_version = ostree.get_kernel_version(sysroot_obj.repo(), csum)
+
+    # Add deployment index part to checksum string, which is 0 for a new deployment
+    csum += ".0"
 
-    if not os.path.isfile(uenv_txt_path):
-        raise FileContentMissing("Couldn't find uEnv.txt in unpacked image rootfs! Aborting.")
+    kernel_path = OSTREE_KERNEL_DEPLOY_PATH.replace('*', csum, 1).replace('*', kernel_version, 1)
```
rborn-tx: A way to simplify this statement and make it more robust (avoiding depending on the order of the parameters):

```
OSTREE_KERNEL_DEPLOY_PATH = "ostree/deploy/torizon/deploy/{csum}/usr/lib/modules/{kver}/"
kernel_path = OSTREE_KERNEL_DEPLOY_PATH.format(csum=csum, kver=kernel_version)
```

tcbuilder/backend/secboot.py
```diff
@@ -510,17 +511,14 @@ def check_basic_signing_prerequisites(storage_dir):
...
         raise InvalidArgumentError("Feature only supported on Torizon OS 7 images. Aborting.")
 
-    log.info("Checking if the unpacked image has the kernel in fitImage format...")
-    uenv_var = get_unpacked_uenv_txt_vars(storage_dir, ["kernel_image_type"])
-
-    if uenv_var["kernel_image_type"]:
-        log.info(f"Detected kernel image format: {uenv_var['kernel_image_type']}")
-    else:
-        log.info("Could not determine the kernel image format")
+    log.info("Checking if the unpacked image has the kernel in FIT format...")
```
rborn-tx: Perhaps this could be made a debug message instead?

tests/integration/run.sh
```diff
@@ -228,6 +223,20 @@ rm -rf $SAMPLES_DIR && cp -a ../$SAMPLES_DIR .
 mkdir -p $REPORT_DIR
 echo -e "Starting integration tests...\n"
 
+# If using TEZI image, check if it has a kernel in FIT format as some tests need this info
+if [ "$IS_WIC" != "1" ] && [ -n "$DEFAULT_TEZI_IMAGE" ]; then
+    torizoncore-builder images --remove-storage unpack "$DEFAULT_TEZI_IMAGE"
+    export IS_DEFAULT_TEZI_IMAGE_FIT="$(set-unpacked-kernel-fit-flag)"
+    if [ "$IS_DEFAULT_TEZI_IMAGE_FIT" = "1" ]; then
```
rborn-tx: A more conventional way of doing this would be:

```
export IS_DEFAULT_TEZI_IMAGE_FIT="0"
if unpacked-kernel-in-fit-format; then
    IS_DEFAULT_TEZI_IMAGE_FIT="1"
fi
```

i.e., the exit/return code of the function would indicate the result rather than the text sent to stdout. `unpacked-kernel-in-fit-format` (or whatever you want to call it) could return 0 for FIT, 1 for non-FIT and maybe 2 for invalid FIT (though the condition here would handle an invalid FIT as non-FIT).

rborn-tx (commented):
I've added some minor suggestions. Good improvements from the previous version, btw!

tcbuilder/backend/common.py
lucas-akira: Nice suggestion. I'll apply it, thanks!

tcbuilder/backend/secboot.py
lucas-akira: Either way would be fine by me, but I have put it as an info log given that a FIT kernel is a strict requirement to run the `secboot` commands, so users would know it always checks this condition beforehand.

tests/integration/run.sh
lucas-akira: We could handle an invalid FIT as non-FIT, or maybe we could check the status code of the function in case it doesn't return 0, and separate non-FIT cases from any errors like invalid FIT, kernel not found, etc.:

```
export IS_DEFAULT_TEZI_IMAGE_FIT=""
if unpacked-kernel-in-fit-format; then
    IS_DEFAULT_TEZI_IMAGE_FIT="1" # vaild FIT kernel
else
    status=$?
    if [ $status -eq 1 ]; then
        IS_DEFAULT_TEZI_IMAGE_FIT="0" # non-FIT kernel
    else
        # The function would provide the specific error message
        exit $status
    fi
fi
```
What do you think?

tcbuilder/backend/secboot.py
jsrc27: Hmm just to add my view. Right after this an error gets printed if the image is not FIT right? So it's kinda redundant information in that sense.

Like just to imagine:

* Image is FIT they see this log, which can be "ignored".
* Image is not FIT and they see this log and then the error log afterwards reiterating that the image is not FIT.

It's not a big deal to me if this remains as is. But I could see the argument of it not being needed on the default log level.

rborn-tx: Just to clarify I also thought it was a bit redundant, but the reason was the print right after the error check. In the FIT case you'd see:

```
Checking if the unpacked image has the kernel in FIT format...
Found kernel in FIT format.
``` 

while in the non-FIT error case:

```
Checking if the unpacked image has the kernel in FIT format...
Unpacked image does not have the kernel in FIT format. Aborting.
```

tests/integration/run.sh
rborn-tx: Sounds good to me.

tcbuilder/backend/secboot.py
lucas-akira: No problem, I'll put it as a debug log message then.

jsrc27 (approved):

tests/integration/run.sh
```diff
@@ -223,11 +218,34 @@ fi
...
 rm -rf $SAMPLES_DIR && cp -a ../$SAMPLES_DIR .
 mkdir -p $REPORT_DIR
 echo -e "Starting integration tests...\n"
 
+# If using TEZI image, check if it has a kernel in FIT format as some tests need this info
+if [ "$IS_WIC" != "1" ] && [ -n "$DEFAULT_TEZI_IMAGE" ]; then
+    torizoncore-builder images --remove-storage unpack "$DEFAULT_TEZI_IMAGE"
+    if unpacked-kernel-in-fit-format; then
+        echo "Image has kernel in FIT format"
+        IS_DEFAULT_TEZI_IMAGE_FIT="1" # vaild FIT kernel
```
rborn-tx: vaild -> valid

rborn-tx (commented):
Found a typo

tests/integration/run.sh
lucas-akira: Fixed, thanks!

rborn-tx (approved):

PR #102, opened 2025-10-17
Attempt to fix issues preventing the release

rborn-tx:
Please check the commit messages for details.

Related-to: TCB-767


.gitlab-ci.yml
```diff
@@ -453,9 +461,14 @@ aggregate-reports:
...
-    # This avoid to run this step in development branches
-    - if: '$ENABLE_XRAY =~ /^true$/i'
+    # On early-access releases (and Xray enabled): always runs this job so we take a picture of the
+    # current situation and store it into Xray.
+    - if: '$RELEASE_TYPE == "early-access" && $ENABLE_XRAY =~ /^true$/i'
       when: always
+    # On official releases (and Xray enabled): run this job manually (give opportunity for us
+    # fixing/repeating previous jobs even before posting the report).
+    - if: '$RELEASE_TYPE == "official" && $ENABLE_XRAY =~ /^true$/i'
+      when: manual
```
lucas-p-bernardes, line 471: Hello @rborn-tx , in `xray-junit-uploader-wrapper.sh` there is this part:

```
for report_path in tests/integration/workdir/reports/report-*-nightly.xml; do
  if [ ! -e "$report_path" ]; then
    echo "Warning: Report path '$report_path' does not exist, skipping."
    continue
  fi
```
Let’s suppose the job `test-host-tos7-platform-nightly` was retriggered 3 times — do you know if there will be 3 XML report files to be used in this script, or if only the last one will exist?

rborn-tx: My understanding is this:

- Each time a test job runs the resulting artifacts (the reports) are uploaded to the Gitlab server and the name of the uploaded file depends solely on the job name (the file is named `report-${CI_JOB_NAME}.xml`). Therefore, every time a job runs it overwrites the previous report kept on the server with the same job name (for that pipeline).

- Then, when the aggregate reports job runs it takes the latest reports from the server, one report for each job previously executed **for that pipeline** (since the artifacts are kept per pipeline).

So, answering your question, there should be only one XML per job, from the perspective of the aggregation job.

rborn-tx: Extra information: looking at the `Build > Artifacts` page of a project I see artifacts being listed by job name with a link to a pipeline which I believe kind of confirms that GitLab keeps artifacts keyed by this pair (pipeline + job). But I think it means the artifacts would be replaced even if we didn't use the job name in the name of the report. For example, if we set the report file name as `report-<datetime>`, the aggregation job would only see the report for the last execution of each job.

rborn-tx:
I'll take the liberty of merging this to try and go ahead with the release.

.gitlab-ci.yml
lucas-p-bernardes: Perfect, I think that’s the expected behavior then.

PR #103, opened 2025-11-05
cli/platform: Properly handle case when credentials file is not found

lucas-akira:
With the exception of `platform push-bootloader`, the `platform` commands (`push`, `static-delta`, `provisioning-data`, and `lockbox`) didn't check if the provided credentials filename in the `--credentials` argument actually refers to an existing file before attempting to access it, which would result in an untreated error message with a traceback in case the file can't be found.

This adds the credentials file check for these commands before any access attempt, and aborts the program if the verification returns false.

Related-to: TCB-392


rborn-tx (approved):
LGTM

jsrc27 (approved):
LGTM

PR #105, opened 2025-11-17
Introduce support for DT operations (apply, status) when the kernel is in the FIT format

rborn-tx:
Summary:

- Make some refactoring to allow reuse in the code being introduced.
- Introduce the infrastructure required for querying and manipulating a kernel FIT image.
- Use that infrastructure to implement the DT (Device Tree) operations (`dt status`, `dt apply`.
- Adapt the secboot sign-kernel command to allow re-signing a kernel image whose DT has been modified.
- Modify tests to perform more detailed checks on the DT application.

### DIFF
--- a/tcbuilder/backend/dt.py
+++ b/tcbuilder/backend/dt.py
@@ -21,13 +21,23 @@ def get_dt_changes_dir(storage_dir):
 
 def get_current_uenv_txt_path(storage_dir):
     '''Get the path to the currently applied uEnv.txt, the bootloader environment file.'''
+    # Check for file in the changes directory.
     path = os.path.join(get_dt_changes_dir(storage_dir), "usr", "lib", "ostree-boot", "uEnv.txt")
     if os.path.exists(path):
-        # Found a recently applied (but not yet deployed) uEnv.txt.
+        log.debug("Found uEnv.txt in changes dir: '{path}'")
         return path
-    # Fallback to uEnv.txt from the base image.
-    path = os.path.join(storage_dir, "sysroot", "boot", "loader", "uEnv.txt")
-    assert os.path.exists(path), "panic: missing uEnv.txt in base image!"
+
+    # Check for the ostree-managed version of the file in the deployment; this
+    # finds the commited version of the file rather the modified copy produced
+    # by libostree in the boot directory when a deployment is created.
+    path = subprocess.check_output(
+        ["find", f"{storage_dir}/sysroot/ostree/deploy",
+         "-wholename", "*/usr/lib/ostree-boot/uEnv.txt",
+         "-print", "-quit"],
+        shell=False, text=True).strip()
+    assert path and os.path.exists(path), "panic: missing uEnv.txt in base image!"
+    log.debug("Found uEnv.txt in deployment: '{path}'")
+
     return path
 
 
--- a/tcbuilder/backend/dt.py
+++ b/tcbuilder/backend/dt.py
@@ -11,6 +11,9 @@ import sys
 import io
 import re
 
+from tcbuilder.errors import InvalidStateError
+from tcbuilder.backend.kernel import get_kernel_changes_dir
+
 log = logging.getLogger("torizon." + __name__)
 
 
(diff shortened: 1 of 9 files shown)


tcbuilder/backend/dt.py
```diff
@@ -21,21 +24,42 @@ def get_dt_changes_dir(storage_dir):
...
+    # Look for the file in two changes directories; the former is used with
+    # non-FIT whereas the latter with FIT images.
+    dchangesdir = get_dt_changes_dir(storage_dir)
+    dpath = os.path.join(dchangesdir, "usr", "lib", "ostree-boot", "uEnv.txt")
+    kchangesdir = get_kernel_changes_dir(storage_dir)
+    kpath = os.path.join(kchangesdir, "usr", "lib", "ostree-boot", "uEnv.txt")
+
+    if os.path.exists(dpath) and os.path.exists(kpath):
+        raise InvalidStateError(
+            "Bad state: uEnv.txt was found both in '{dpath}' and '{kpath}'")
```
jsrc27, line 37: Should we also maybe give a suggestion to fix the issue in this error. Since the average user may not understand what it means when they get this error.

Do we just suggest the user to clear storage and try again, or is there a better suggestion?

tcbuilder/backend/kernelfit.py
```diff
@@ -0,0 +1,689 @@
...
+    try:
+        node_off = fdt.path_offset(node_path, quiet=QUIET_NOTFOUND)
+    except libfdt.FdtException as exc:
+        raise KernelFitException(
+            f"Error obtaining offset of '{node_path}' within FDT.") \
+            from exc
+    return node_off >= 0
+
+
+# AI-generated:
```
jsrc27: Do you intend to keep this comment in the final code?

tcbuilder/backend/kernelfit.py
```diff
@@ -0,0 +1,689 @@
...
+        size += _estimate_node_size(fdt, child_off)
+        child_off = fdt.next_subnode(child_off, QUIET_NOTFOUND)
+
+    # End node tag
+    size += 4
+
+    return size
+
+
+# AI-generated:
```
jsrc27: Same here

tcbuilder/backend/kernelfit.py
```diff
@@ -0,0 +1,689 @@
...
+           list[dict]: A list of dictionaries, each with the fields:
+               - conf_name (str): Name of the configuration node.
+               - image_name (str): Name of the image referenced by the configuration
+                     node.
+               - file_name (str): This is expected to represent the name of the file
+                     that originated the entry in the FIT image.
+        """
+        sufre = suf.replace('.', '\\.')
+        names = self._list_nodes(CONFIG_NODE_PATH, f"{pref}.*{sufre}")
+        # log.debug("get_dtb_entries_helper: names=%s", str(names))
```
jsrc27: Did you want to keep this commented out, or maybe just remove it?

jsrc27 (commented):
Left some minor comments but looks good for the most part.

Just some questions/clarifications on the UI/UX.

If a user has an OS with a FIT image they can just customize the device-tree basically the same way they would with a non-FIT image. But to re-sign they need to explicitly run the `secboot` commands, or have the equivalent fields in their build yaml. 

Basically the workflows would be:

* Signed FIT image:

`images unpack` -> `dt` commands -> `secboot` commands -> `union` -> `deploy`

* Or, if the user has a not-signed FIT image for some reason:

`images unpack` -> `dt` commands  -> `union` -> `deploy`
(Which is basically the same flow I guess for a non-FIT image as well.)

So as a user I would just need to know whether my image was originally signed or not. As that's really the only difference in the process here. Is that all correct?

tcbuilder/backend/dt.py
rborn-tx: The only way I can imagine for this error to be triggered is:

- User runs `images unpack <UNSIGNED image>` followed by DT customization.
- User runs `images unpack <SIGNED image>` **without** clearing the storage followed by an attempt to customize the DT.

So, yes, in theory the error would happen when the user deliberately chooses not to clear the storage - something I'd expect only an advanced user to attempt. And if such a user sees a message indicating the state is invalid I'd expect them to reasonably conclude that it was caused by not clearing the state during their previous `images unpack` and try differently on the next attempt.

Given the above, in my view the current message is enough and no particular suggestion is needed. But if you prefer for us to to add a suggestion (which I'm not against) I'd change the code to throw a more specific exception and catch it at the `cli` layer to show an appropriate message with the suggestion.

The reason I don't want to simply change the message here is that this stretch of code is at the `backend` layer which shouldn't concern itself with usage details (those are responsibility of the `cli` layer).

tcbuilder/backend/kernelfit.py
rborn-tx: Good question. Here I was being clear on how that stretch of code was produced. Since I reviewed and even changed parts of it, it's probably best to get rid of that comment.

tcbuilder/backend/kernelfit.py
rborn-tx: It can be safely removed at this point.

tcbuilder/backend/kernelfit.py
rborn-tx: Done

tcbuilder/backend/kernelfit.py
rborn-tx: Done

tcbuilder/backend/kernelfit.py
rborn-tx: Done

rborn-tx:
> If a user has an OS with a FIT image they can just customize the device-tree basically the same way they would with a non-FIT image. But to re-sign they need to explicitly run the `secboot` commands, or have the equivalent fields in their build yaml.
> 
> Basically the workflows would be:
> 
> Signed FIT image:
> 
> 
> `images unpack` -> `dt` commands -> `secboot` commands -> `union` -> `deploy`
> 
> Or, if the user has a not-signed FIT image for some reason:
> 
> 
> `images unpack` -> `dt` commands -> `union` -> `deploy` (Which is basically the same flow I guess for a non-FIT image as well.)
> 
> So as a user I would just need to know whether my image was originally signed or not. As that's really the only difference in the process here. Is that all correct?

Your understanding is correct. In general, for signed images users would just need to run re-signing commands right before creating the commit with `union`. In theory, for BCoT images this would be strictly needed only if they made changes impacting the kernel, which in practice means most of the customization types provided by TCB (splash, dt, dto, kernel) with only plain filesystem (`/etc`) changes excluded.

tcbuilder/backend/dt.py
rborn-tx: Please let me know your opinion and thanks for the feedback so far.

jsrc27: I don't have strong opinions about this, so I'm okay with leaving it as is.

jsrc27 (approved):

PR #106, opened 2025-12-01
Introduce support for DTO operations (apply, list, status, remove) when the kernel is in the FIT format

rborn-tx:
Summary:

- Make quite a bit of refactoring to allow reuse of the code with the commands being introduced and to drop duplicated code.
- Modify existing tests to perform more detailed checks on the DTO application.
- Add tests for performing checks specific to the FIT case.

Related-to: TCB-762


rborn-tx:
NOTE: moved to draft to check failure with Torizon OS 5.

rborn-tx:
Ready for review now.

tcbuilder/cli/dto.py
```diff
@@ -264,52 +485,110 @@ def do_dto_list(args):
...
+    dtob_basenames = dto.get_applied_overlay_names(storage_dir)
+    dtob_basenames.remove(dtob_basename)
+    dto.set_applied_overlay_names(dtob_basenames, changes_dir, storage_dir)
+    # Remove the overlay blob if it's not deployed.
+    dtob_path = dto.find_path_to_overlay(storage_dir, dtob_basename)
+    if dtob_path.startswith(changes_dir):
+        os.remove(dtob_path)
+
+
+def _dto_remove_single_fit(dtob_basename, changes_dir, storage_dir, restore=True):
```
jsrc27, line 546: I see we default the `restore` parameter to `True`. But I don't currently see anywhere where this function is called and `restore` is set to anything else. Is there a reason for this? Maybe internal debugging/testing, or even a future use-case?

tcbuilder/cli/dto.py
```diff
@@ -151,58 +250,172 @@ def do_dto_apply(args):
...
+    This handles the FIT case where the DTB would reside inside the kernel
+    FIT image. To be able to return a path, this function will extract the
+    DTB into a temporary directory.
+    """
+
+    if dtb_name and "/" in dtb_name:
+        # In the future the presence of slashes may be used to allow passing a
+        # DTB path outside the storage.
+        log.error("error: the device tree name (%s) must not contain slashes.", dtb_name)
+        sys.exit(1)
```
jsrc27, line 325: Huh, interesting were we not checking for slashes in the name of the dtb before? Was it not problematic before? Or it was and we just never noticed?

Also do you think it's important to add a test-case for slashes now that we explicitly check for this?

jsrc27 (commented):
Gave it an initial review. Mostly focusing on the new feature additions/changes. I did also look at the refactors but they seemed fine. Mostly just renaming, moving, reducing code which is nice.

By the way, one small thought (not necessarily for this PR but in general going forward). Do you think we should start to document the architecture of the storage directory? It just came to my mind since such details become more important with how FIT images change where specific artifacts are stored in the storage directory. It's not super vital/urgent, but could help with keeping things clear in the future, and for external contributors of course.

tcbuilder/cli/dto.py
rborn-tx: Before, if the user passed a path to a dtb, that path would be relative the dtb/ directory in the image (which in turn is under the version-specific kernel directory). Since the dtb/ directory does not have sub-directories, if the user passed one, TCB would fail to access the file with some not very user-friendly message. That's why I added this check.

As a note, if the user passed a path to a dts, then the path would be relative to the work directory which, I think, is sensible. But it feels a bit inconsistent. So, as I wrote in the comment, we could check for the presence of a path which, when present, could represent a path relative to the work directory for consistency with the dts case.

As for the test, I tested locally but I didn't create a test because that didn't feel like a important use case. But, it does not hurt to have one. I'll do that...

tcbuilder/cli/dto.py
rborn-tx: The default is True to try and mimic the behavior of the non-FIT case, where the DTs/DTOs are never really removed (they are removed from the changes directory but that doesn't cause them to be removed from the final image). I considered having a behavior as close as possible to the non-FIT case, as a general goal for this initial work of introducing support for FIT.

Now, I added this `restore` parameter to allow for a different behavior in the future in case we want to provide some FIT-specific behavior to users. Of course, this would require some new switch to the command that I didn't want to add right now.

Having the switch might be useful because it would allow users to really remove DTOs (and the same could be done for DTs) to make the kernel binary smaller and there might be good security reasons as well to drop stock DTs/DTOs from the FIT.

rborn-tx:
@jsrc27 Regarding:

> By the way, one small thought (not necessarily for this PR but in general going forward). Do you think we should start to document the architecture of the storage directory? It just came to my mind since such details become more important with how FIT images change where specific artifacts are stored in the storage directory. It's not super vital/urgent, but could help with keeping things clear in the future, and for external contributors of course.

We definitely could. I think today we are missing even more basic docs such as explanations about the code design/structure that we should probably also have.

tcbuilder/cli/dto.py
jsrc27: Okay so the parameter is for potential future use then. But, for the time being it's always true as you said. Should we add a small comment for this parameter that explains this or do you think it's fine as is?

tcbuilder/cli/dto.py
jsrc27: Makes sense.

tcbuilder/cli/dto.py
rborn-tx: I've added function comments explaining all parameters including the `restore` one, plus a TODO for us to consider providing users with the corresponding functionality.

tcbuilder/cli/dto.py
rborn-tx: Test cases added.

tcbuilder/cli/dto.py
```diff
@@ -8,132 +8,230 @@
...
+    log.info(f"Overlay {dtob_name} can successfully modify the device tree "
+             f"{os.path.basename(dtb_path)}.")
+
+    return True
+
+
+def _test_apply_overlay_fit(*,
+                            dtob_path, dtob_name, dtb_name,
+                            overlay_names, changes_dir, storage_dir):
+    """Try to apply an overlay on top of a device-tree (non-FIT case).
```
lucas-akira: Here the description should say that it's the 'FIT case'

lucas-akira (commented):
Did an initial review, focusing on the new feature and the refactoring. I had some catch-up to do as this PR modifies code merged from a previous PR that I didn't review, mostly related to the new `backend/kernelfit.py` contents. I didn't look at the tests yet.

Overall great additions, it makes sense to add the DT/DTO modifications in the `kernel` changes directory for the FIT kernel.

About the `dto remove` command, I agree that it would be nice to actually remove the DTOs from the FIT to reduce the binary size, though I wonder if we should also make this option available for the non-FIT case as well to keep feature parity between the two cases.

tcbuilder/cli/dto.py
rborn-tx: Good catch. Done!

rborn-tx: Regarding:

> About the `dto remove` command, I agree that it would be nice to actually remove the DTOs from the FIT to reduce the binary size, though I wonder if we should also make this option available for the non-FIT case as well to keep feature parity between the two cases.

Good point. Keeping feature parity would probably be the best option from the user perspective. We should definitely consider that if/when we decide to add to the tool this ability of fully deleting the DTB/DTBO from the image (which of course would require new switches to commands and new properties in `tcbuild.yaml`).

jsrc27 (approved):
Did another quick look. I don't see anything further to comment on.

lucas-akira (approved):
I took a look at the tests, and did a quick re-review on the rest; everything looks good to me.

PR #108, opened 2025-12-10
Introduce support for kernel bootargs customization (set, get, clear) when the kernel is in the FIT format

rborn-tx:
Add support for setting/getting/clearing custom kernel bootargs when the kernel is in the FIT format. This includes handling the **custom-kargs** overlay inside the FIT image, when present; this is the case with Secure Boot where the full kernel command line (not only the user supplied additions) must be statically defined to satisfy the hardening feature known as the kernel command-line protection.

Related-to: TCB-763

### DIFF
--- a/tcbuilder/backend/dt.py
+++ b/tcbuilder/backend/dt.py
@@ -18,12 +18,15 @@ log = logging.getLogger("torizon." + __name__)
 
 DTB_PREFIX_RE = re.compile(r'bootm[^#]*#conf-([^$]*)\$')
 
+BACKSLASH_SPC_RE = re.compile(r"\\\s*$")
+
 
 def get_dt_changes_dir(storage_dir):
     """Returns the directory that contains external device tree related changes."""
     return os.path.join(storage_dir, "dt")
 
 
+# TODO: Consider moving the various "uenv" functions to a separate module (other than common).
 def get_current_uenv_txt_path(storage_dir):
     """Get the path to the currently applied uEnv.txt, the bootloader environment file."""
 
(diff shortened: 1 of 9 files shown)


tcbuilder/backend/dt.py
```diff
@@ -57,6 +60,111 @@ def get_current_uenv_txt_path(storage_dir):
     return dpath
 
 
+def set_uenv_txt_variable(var_name, var_value, *,
+                          changes_dir, storage_dir, append=False):
```
jsrc27, line 64: Could we maybe document the usage of `var_value` being `None` as a way to mean that the variable is being deleted/cleared. It wasn't apparent to me on first look that `None` was an acceptable value and what it meant.

tcbuilder/cli/kernel.py
```diff
@@ -41,19 +44,36 @@
...
+
+SECBOOT_NODE_PATH = "/fragment@0/__overlay__/chosen/toradex,secure-boot"
+SECBOOT_REQ_BOOTARGS_PROPNAME = "required-bootargs"
+SECBOOT_REQ_BOOTARGS_ORG_PROPNAME = "required-bootargs-org"
+
+
+def _on_custom_kargs_not_supported():
+    raise UnsupportedImageFeature(
+        "Error: the Torizon OS image you are customizing does not support "
+        "custom kernel arguments; please upgrade to a newer OS image version.")
```
jsrc27, line 68: I know the new karg method isn't in an official release yet but it should be in 7.5.0. I believe correct? Should we specify this in the error message?

tcbuilder/cli/kernel.py
```diff
@@ -143,146 +326,236 @@ def kernel_set_custom_args(kernel_args, storage_dir):
...
+    kargs_methods = kernel.get_supported_bootargs_methods(storage_dir)
 
-    # Format string to become file contents.
-    dts_contents = KERNEL_SET_CUSTOM_ARGS_DTS.format(kernel_args=kargs)
+    # Check image requirements:
+    if kernel_is_fit and "uenv" not in kargs_methods:
+        raise UnsupportedImageFeature(
+            "Error: the Torizon OS image you are customizing has a kernel in FIT "
+            "format but it does not support the uEnv method for passing kernel "
+            "arguments; please upgrade to a newer OS image version.")
```
jsrc27: Same thing here.

tcbuilder/cli/kernel.py
```diff
@@ -143,146 +326,236 @@ def kernel_set_custom_args(kernel_args, storage_dir):
...
+    log.debug(f"kernel_clear_custom_args: kernel_is_fit={kernel_is_fit}")
+
+    kargs_methods = kernel.get_supported_bootargs_methods(storage_dir)
+
+    # Check image requirements:
+    if kernel_is_fit and "uenv" not in kargs_methods:
+        raise UnsupportedImageFeature(
+            "Error: the Torizon OS image you are customizing has a kernel in FIT "
+            "format but it does not support the uEnv method for passing kernel "
+            "arguments; please upgrade to a newer OS image version.")
```
jsrc27: Same thing here.

tcbuilder/backend/dt.py
jsrc27, line 64: It seems the `append` does not get used in this implementation yet. I assume this is something for the future possibly?

tcbuilder/cli/kernel.py
```diff
@@ -143,146 +326,236 @@ def kernel_set_custom_args(kernel_args, storage_dir):
...
+    # +----------+----------+------+
+    # | uEnv.txt | Yes      | Yes  |
+    # +----------+----------+------+
+    #
+    if kernel_is_fit and "uenv" not in kargs_methods:
+        log.info(
+            "Notice: this image has a kernel in FIT format but it does not support "
+            "the uEnv method for passing kernel arguments; this means you will not "
+            "be able to set its custom kernel arguments. This can be solved by "
+            "upgrading to the latest Torizon OS image.")
```
jsrc27: Same here about specifying the version.

jsrc27 (commented):
Looks good for the most part, just left some minor comments/questions.

tcbuilder/cli/kernel.py
rborn-tx: This particular message would be shown only if the image did not support custom kernel arguments at all (not even with the previous (overlay-based) method). This would only happen to images older than April/2021.

So, here it wouldn't be correct to show version 7.5.0. But on the other cases, yes it would make sense to show that version.

tcbuilder/cli/kernel.py
rborn-tx: Done

tcbuilder/cli/kernel.py
rborn-tx: Done

tcbuilder/cli/kernel.py
rborn-tx: Done

tcbuilder/backend/dt.py
rborn-tx: My idea is to make this new function the only responsible for updating `uEnv.txt` now and in the future and I imagined someone might want to make additions to that file at the end rather than at the beginning for example to mimic what ostree does when creating a deployment. But, yeah, that's for possible future implementations.

rborn-tx: Added a function comment explaining all parameters.

tcbuilder/backend/kernel.py
```diff
@@ -235,7 +243,9 @@ def find_kernel_in_sysroot(storage_dir):
     kernel_path = OSTREE_KERNEL_DEPLOY_PATH.format(csum=csum, kver=kernel_version)
     kernel_path = os.path.join(sysroot_path, kernel_path, OSTREE_KERNEL_FILENAME)
 
-    if not os.path.isfile(kernel_path):
+    if os.path.exists(kernel_path):
+        log.debug("Kernel found at '%s'", kernel_path)
+    else:
         raise PathNotExistError("Kernel not found in unpacked rootfs. Aborting.")
```
lucas-akira, line 249: Is there a specific reason to replace `os.path.isfile()` with the more generic `os.path.exists()`, since we're always looking for a regular file in this case?

lucas-akira (commented):
Looks good to me, nice work. Only have one small question.

jsrc27 (approved):

tcbuilder/backend/kernel.py
rborn-tx: Oh, I'm not sure how I managed to do that, but it wasn't intentional. Thanks for catching it.

rborn-tx: Fixed

lucas-akira (approved):

PR #109, opened 2025-12-18
Create new "splash-config" command

jsrc27:
Create new `splash-config` command that can customize the Plymouth configuration file for our splash screen on Torizon OS. Also did some refactoring on the `splash` command so that it can work more smoothly with `splash-config` and any future features that are also related to splash-screen configuration.

### DIFF
--- a/tcbuilder/cli/splashconfig.py
+++ b/tcbuilder/cli/splashconfig.py
@@ -1,0 +1,164 @@
+"""
+CLI for the splash-config command
+"""
+
+import logging
+import os
+import shutil
+
+from tcbuilder.backend import splashconfig as splashconfig_be
+
+from tcbuilder.backend.common import \
+    (images_unpack_executed, set_output_ownership,
+     is_file_type_fit)
+from tcbuilder.errors import \
+    (InvalidStateError, PathNotExistError)
+from tcbuilder.backend.initramfs import UnpackedInitramfs
+from tcbuilder.backend.kernel import \
+    (find_kernel_in_sysroot, get_kernel_changes_dir)
+from tcbuilder.backend.splash import \
+    (get_splash_changes_dir, PLYMOUTH_THEME_PATH)
+from tcbuilder.backend.splashconfig import PLYMOUTH_CONFIG
+
+log = logging.getLogger("torizon." + __name__)
+
+
+def splash_config_set(config):
+    """Main handler for 'splash-config set' command.
+
+    :param config: Path to provided configuration file.
+    """
+
+    config_file = os.path.abspath(config)
+    if not os.path.exists(config_file):
+        raise PathNotExistError(f"Unable to find configuration file {config}.")
+
+    with UnpackedInitramfs(source="auto") as rdm:
+        splashconfig_be.add_plymouth_splash_config(rdm.get_path(), config_file)
+
+    unpacked_kernel_path = find_kernel_in_sysroot()
+    kernel_is_fit = is_file_type_fit(unpacked_kernel_path)
+    log.debug(f"splash: kernel_is_fit={kernel_is_fit}")
+
+    if kernel_is_fit:
+        # FIT case: all changes go to the kernel-changes directory.
+        changes_dir = get_kernel_changes_dir()
+    else:
+        # non-FIT case: changes go to the splash-changes directory.
+        changes_dir = get_splash_changes_dir()
+
+    splashconfig_be.add_plymouth_splash_config(changes_dir, config_file)
+
+    log.info("Splash screen configuration updated")
+
+
+def do_splash_config_set(args):
+    """Run 'splash-config set' command."""
+
+    images_unpack_executed()
+
+    splash_config_set(args.splash_config)
+
+
+def create_config(output_file, config, force):
+    """Prints or writes the current active Plymouth Configuration file.
+
+    :param output_file: Filename for where to write the conent. If 'None'
+                        then prints the content to console instead.
+    :param config: Path to currently active config file.
+    :param force: Whether to overwrite output_file if it arleady exists.
+    """
+
+    # Dump to stdout
+    if not output_file:
+        print()
+        with open(config, 'r') as file:
+            for line in file:
+                print(line, end='')
+    else:
+        if os.path.exists(output_file) and not force:
+            raise InvalidStateError(f"File '{output_file}' already exists; aborting.")
+
+        log.info(f"Creating configuration file '{output_file}'")
+        shutil.copy(config, output_file)
+        set_output_ownership(output_file)
+
+
+def splash_config_dump(output_file, force):
+    """Main handler for splash-config dump.
+
+    :param output_file: Filename to write splash config content to.
+    :param force: Whether to overwrite output_file if it already exists.
+    """
+
+    unpacked_kernel_path = find_kernel_in_sysroot()
+    kernel_is_fit = is_file_type_fit(unpacked_kernel_path)
+    log.debug(f"splash: kernel_is_fit={kernel_is_fit}")
+
+    if kernel_is_fit:
+        # FIT case: all changes go to the kernel-changes directory.
+        changes_dir = get_kernel_changes_dir()
+    else:
+        # non-FIT case: changes go to the splash-changes directory.
+        changes_dir = get_splash_changes_dir()
+
+    config_custom = os.path.join(changes_dir, PLYMOUTH_THEME_PATH, PLYMOUTH_CONFIG)
+
+    # First determine whether a customized config is present,
+    # otherwise we use the one from the unpacked image.
+    if os.path.exists(config_custom):
+        config_current = config_custom
+        log.info("Found customized configuration file.")
+    else:
+        config_current = splashconfig_be.get_default_config()
+        log.info("Found default configuration file.")
+
+    create_config(output_file, config_current, force)
+
+
+def do_splash_config_dump(args):
+    """Run 'splash-config dump' command."""
+
+    images_unpack_executed()
+
+    splash_config_dump(args.output_file, args.force)
+
+
+def init_parser(subparsers):
+    """Initialize 'splash-config' subcommands command line interface."""
+
+    parser = subparsers.add_parser(
+        "splash-config",
+        help=("Perform customizations that affect the configuration of the splash "
+              "screen."),
+        allow_abbrev=False)
+    subparsers = parser.add_subparsers(title='Commands', required=True, dest='cmd')
+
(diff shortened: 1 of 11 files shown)


lucas-akira (commented):
New feature looks good to me. One question: Should we add an option to reset the splash config back to the default one from the unpacked image, in a similar way we have the `kernel clear_custom_args` command?

jsrc27:
> New feature looks good to me. One question: Should we add an option to reset the splash config back to the default one from the unpacked image, in a similar way we have the kernel clear_custom_args command?

I thought about it. But I think it's best to use unpacking the image as a more standard way to globally reset the image of any customization.

tcbuilder/cli/union.py
```diff
@@ -195,6 +195,12 @@ def union(changes_dirs, storage_dir, union_branch,
 
     storage_dir_ = os.path.abspath(storage_dir)
 
+    # Create initramfs for splash changes if needed
+    splash_dir = os.path.join(storage_dir_, "splash")
+    if os.path.exists(splash_dir):
+        src_ostree_archive_dir = os.path.join(storage_dir, "ostree-archive")
+        ub.create_splash_initramfs(splash_dir, src_ostree_archive_dir)
```
rborn-tx: From what I can gather, this implementation is introducing a new approach regarding the management of the tool state, right? I mean, today all stateful commands create or update "changes" directories which are ready to be consumed by OSTree via the `union` command. With this new approach, we need to preprocess some intermediate directory in order to produce the changes directory before OSTree can consume it.

This approach looks sensible at first sight and it's probably better in terms of performance if we had followed it with all commands. Unfortunately that's not the case and introducing a new approach for a single command raises concerns about the impact it could have on both new and existing commands.

The new approach would basically force new commands operating in artifacts directly or indirectly produced from the intermediate ones to follow the same approach, that is, to postpone their work to when `union` is run. This means, errors related to them might also be detected later (during `union`) which would likely become a source of confusion to users of the stateful workflow.

Thinking a bit about existing commands:

Today we already have the command that signs the kernel and we are about to change the `splash` command to add support for FIT images. The kernel signing command takes the kernel as input and produces a new, signed one. With the new approach, we'd need to move the signing to the `union` command too (because only then we'd create a new ramdisk and use it to create a new kernel FIT image). But that would be strange to users, as the signing wouldn't happen during the `secboot sign-kernel` invocation but later. We can imagine cases where the user is keeping their keys in a PKCS#11 device and they'd expect the keys to be accessed during that invocation, which would make things even stranger.

@jsrc27 Given the above, I believe introducing this new approach at this point is not a good idea for us or for users of TorizonCore Builder. But, let me know your thoughts.

Anyway, since we're working with the `splash` command for FIT, we should probably wait a bit before continuing the work here to avoid hard-to-solve conflicts...

rborn-tx (commented):
I've stopped my review preemptively because the concern I've raised will likely change a lot the implementation.

tcbuilder/cli/union.py
jsrc27: Well for secure boot/FIT images I did not take any of this into consideration as it was not in the scope of the ticket I was working on here.

In general, the only reason I went with this approach was because now we would have two commands that optionally customize the initramfs, that can be ran by the user in any order. Currently the `splash` command creates the customized initramfs file to be consumed later. If this approach were to be followed for `splash-config` there would be a weird ordering issue. Meaning whichever command was ran second would need to check if there was already a customization done, then recreate the customized initramfs file again to include the additional changes from whatever command was ran second.

It could be done, but for obvious reasons this approach seemed inefficient and redundant. Therefore I decided to move the creation of the initramfs to later in the process. At a point where TCB can know that all splash related commands have been run (or not run). That way any customizations can be considered at the same, meaning the initramfs only needs to be generated once if needed. It also avoids the issue of `splash` and `splash-config`  needing to "know" or check whether the other command has been ran first . Since ultimately these commands should be independent of one another.

> The new approach would basically force new commands operating in artifacts directly or indirectly produced from the intermediate ones to follow the same approach, that is, to postpone their work to when union is run. This means, errors related to them might also be detected later (during union) which would likely become a source of confusion to users of the stateful workflow.

Not sure if I understand the concern here. We only have splash screen related operations that create a customized initramfs currently. Currently we only have the one command that does so, meaning this command dictated things. Are we not allowed to change the precedent set by this one command that was created years ago?

rborn-tx: Yes, the performance benefits are clear as I mentioned.

But, how did you intend to solve the issues I described above (which come from the fact that with secure boot images the ramdisk is inside the kernel FIT image)? I'm talking about this paragraph:

> Today we already have the command that signs the kernel and we are about to change the splash command to add support for FIT images. The kernel signing command takes the kernel as input and produces a new, signed one. With the new approach, we'd need to move the signing to the union command too (because only then we'd create a new ramdisk and use it to create a new kernel FIT image). But that would be strange to users, as the signing wouldn't happen during the secboot sign-kernel invocation but later. We can imagine cases where the user is keeping their keys in a PKCS#11 device and they'd expect the keys to be accessed during that invocation, which would make things even stranger.

You could think that this is out of scope because relates to Secure Boot, but if we don't think ahead, we could introduce a change that breaks the Secure Boot workflow which would lead to complicated discussions in the near future (notice that the support for the `splash` command with FIT-based images is in progress right now).

lucas-akira: I haven't thought about the creation of the updated initramfs on the `union` command affecting the re-signing workflow. I think both points on keeping/re-working the changes are valid concerns:

- We could keep things in the PR as they are now, but that would require later moving the kernel signing step to the `union` command as well, which I agree that it would be strange, as `secboot sign-kernel` wouldn't actually sign anything.

- We could recreate the entire initramfs binary each time the splash/splash-config command is invoked, each time checking if a customization to initramfs was made beforehand, but that wouldn't be very efficient.

From what I understand currently for the `splash` command TCB creates a zipped CPIO file that only has the PNG image then appends it to what is effectively a copy of the unmodified initramfs file, which is also a zipped CPIO file. The end result is that the provided PNG image ends up overwriting the existing `watermark.png`.

Since `splash` and `splash-config` modify different files present in initramfs, their appending steps shouldn't interfere with one another, right?

If we remove the entire initramfs creation step in `splash` and `splash-config` and only append their additions to an existing initramfs file in the changes directory, wouldn't that solve the ordering issue? The unmodified initramfs binary could be copied to the changes directory during `images unpack`, so `splash`/`splash-config` would always have it available from the start.

jsrc27: > The unmodified initramfs binary could be copied to the changes directory during images unpack, so splash/splash-config would always have it available from the start.

So we'd always have an initramfs file in changes from the start no matter what. What if no initramfs customizations were done? Would `union` just "commit" this into the the new OSTree despite it not actually changing? I guess in theory for the user this wouldn't matter, though from a technical perspective it would be weird to commit something that didn't really change.

I guess `union` could check for the existence of other artifacts to see if an initramfs customization was done (i.e if `usr/share/plymouth/themes/spinner/` exists). Then, `union` could decide whether to include the initramfs in the commit or not. Though this creates an annoying precedent for future commands that may customize initramfs. As `union` would need to check for some indication if any of these commands were ran prior as well. In which case do we just commit the initramfs found in changes regardless no matter what every time?

Furthermore, another less than favorable thing is that each initramfs customizing command would still need to create their own zipped CPIO file to be appended. Rather than just doing it once for all initramfs customizations.

jsrc27:
Rebased with the latest changes.

tcbuilder/cli/splashconfig.py
```diff
@@ -0,0 +1,148 @@
...
+    # First determine whether a customized config is present,
+    # otherwise we use the one from the unpacked image.
+    if os.path.exists(config_custom):
+        config_current = config_custom
+        log.info("Found customized configuration file.")
+    else:
+        config_current = splashconfig_be.get_default_config()
+        log.info("Found default configuration file.")
+
+    create_config(args.output_file, config_current)
```
rborn-tx: Would it make sense to factor this out into a `splash_config_dump()` function?

That would keep us aligned with the pattern we've been using, where `do_xxx()` focuses on connecting argument parsing (`ArgParse`) with the internal (CLI-level) API. It also makes reuse easier and clarifies what data the API actually needs, given that CLI arguments are wrapped in an args object with non-obvious properties.

tcbuilder/cli/splashconfig.py
```diff
@@ -0,0 +1,148 @@
...
+        "dump",
+        help=("Prints the current Plymouth configuration."),
+        allow_abbrev=False)
+    subparser.add_argument(
+        "--file",
+        dest="output_file",
+        help=("Writes current Plymouth configuration to the provided file name."),
+        required=False,
+        default=None)
+    subparser.set_defaults(func=do_splash_config_dump)
```
rborn-tx, line 164: I just thought: Would it be too much to have a `--force` flag here to cause the output file to be overwritten if it existed?

rborn-tx (commented):
Looks great! I've added some suggestions/questions nonetheless.

Btw, I see that the new implementation of `splash` made the addition of this new command much simpler compared to the previous attempt. What do you think?

tcbuilder/cli/splashconfig.py
```diff
@@ -0,0 +1,148 @@
...
+
+    images_unpack_executed()
+
+    splash_config_set(args.splash_config)
+
+
+def create_config(output_file, config):
+    """Prints or writes the current active Plymouth Configuration file.
+
+    :param config_file: Filename for where to write the conent. If 'None'
```
lucas-akira: I guess the parameter here should now be `output_file`.

lucas-akira (approved):
Code looks good to me. I only have one comment.

tcbuilder/cli/splashconfig.py
jsrc27: Done in V2

tcbuilder/cli/splashconfig.py
jsrc27: Done in V2.

tcbuilder/cli/splashconfig.py
jsrc27: Nice catch, done in V2.

rborn-tx:
@jsrc27 Once this is merged, please remember to cherry-pick it to the new `trixie` branch.

rborn-tx (approved):
LGTM

rborn-tx:
@jsrc27 I've already created a PR updating trixie with the changes in this PR: https://github.com/torizon/torizoncore-builder/pull/124.

PR #110, opened 2025-12-18
Introduce support for kernel module building when the kernel is in the FIT format

rborn-tx:
Add support for building kernel modules via `kernel build_module` when kernel is in the FIT format. Besides that, solve some relate bugs found when making the implementation, namely:

- When a module was auto-loaded (via switch --autoload passed to `kernel build_module`) the various module names were accumulating into the `tcb.conf` file.
- The prefix to some error messages were duplicated when running the build command because the command itself added the prefix in some cases.
- The build command was signing the kernel before other customization was performed which resulted in a kernel with not all required nodes being signed (ultimately preventing the boot of a device).

Related-to: TCB-763

### DIFF
--- a/tcbuilder/backend/kernel.py
+++ b/tcbuilder/backend/kernel.py
@@ -38,11 +38,11 @@ SET_BOOTARGS_TORIZON_RE = r'^\s*set_bootargs_torizon='
 
 
 def get_kernel_changes_dir(storage_dir):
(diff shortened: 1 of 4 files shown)


jsrc27 (approved):
No major issues or comments from my side.

lucas-akira (approved):
LGTM

PR #112, opened 2025-12-22
kernel: Prepend custom kargs with secboot images

rborn-tx:
When detecting that the image has secboot features, we now prepend (rather than append) the custom kargs string to the existing bootargs. This creates a limitation in that users won't be able to override bootargs explicitly set in the base image, but it makes the change in bootargs compatible with the kernel command-line protection feature (which is part of the Toradex hardening features in U-Boot). This relies on the new variables torizon_bootargs_{l,r} provided by the latest boot script (uEnv.txt).

If the image is not a secboot one, the kargs are still appended as usual. This is to keep the behavior exactly as before for non-secboot users, avoiding a breaking change from their perspective.

NOTES:

1. This will need to be rebased after the following PR is merged (DONE):
   * https://github.com/torizon/torizoncore-builder/pull/110
2. The changes will work with secure boot images built after the following PR is merged (DONE):
   * https://github.com/torizon/meta-toradex-torizon/pull/408

~(leaving as draft until the above happens)~

### DIFF
--- a/tcbuilder/cli/kernel.py
+++ b/tcbuilder/cli/kernel.py
@@ -43,9 +43,12 @@ KERNEL_SET_CUSTOM_ARGS_DTS = """
 # XXX: Always keep in sync with uEnv.txt.in in recipe `u-boot-distro-boot`.
 KERNEL_SET_CUSTOM_ARGS_PROPERTY = 'bootargs_custom'
 
-# Name of the property defining the kernel args in uEnv.txt.
+# Name of the variables in uEnv.txt defining the kernel args; one variable
+# is added to the left (prepended) and the other to the right (appended) of
+# the original bootargs string.
 # XXX: Always keep in sync with uEnv.txt.in in recipe `u-boot-distro-boot`.
-UENV_CUSTOM_BOOTARGS_VAR = "torizon_boot_args"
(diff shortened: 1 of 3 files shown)


tcbuilder/cli/kernel.py
```diff
@@ -370,8 +390,21 @@ def kernel_set_custom_args(kernel_args, storage_dir):
...
+    # use the presence of the overlay to decide whether the passed bootargs should
+    # be prepended or appended to the original secboot bootargs string.
+    sb_kargs_ovl_pres = False
+    if kernel_is_fit:
+        sb_kargs_ovl_pres, _ = _secboot_kargs_ovl_present(storage_dir)
+
     if "uenv" in kargs_methods:
-        _set_custom_kargs_uenv(kargs, changes_dir, storage_dir)
+        if sb_kargs_ovl_pres:
+            log.info(
```
lucas-akira, line 403: Could it be a `log.warning` here, given the nature of the message?

lucas-akira (commented):
Overall looks good to me.

tcbuilder/cli/kernel.py
rborn-tx: Good point. Done!

PR #113, opened 2025-12-22
backend/deploy: Create output raw image from a copy of the base one

lucas-akira:
The generated output disk image is now based on a copy of the base image instead of it being initially an empty file that is then filled with the corresponding partitions.

This makes the output image have all partitions not listed in the partition table that the base image has, which may hold critical data like bootloader binaries.

The final output image size criteria has also been changed to take into account the entire base disk image size instead of only its rootfs size.

Related-to: TCB-800


rborn-tx (approved):
LGTM

PR #114, opened 2025-12-30
Fix DTO test on TCB images with 'early-access' tag

lucas-akira:
Fix a test that only fails when running a TorizonCore Builder image with the 'early-access' tag.

This specific tag causes an additional message to be printed to `stderr` when running every command informing that the image is an early-access version. This extra message made the test fail as it only expected the one line of output from the 'dto status' command itself.

Also increase the verbosity level for a WIC build test to facilitate debugging, and add a new `COMMON_TORIZON_ALLOW_FAILURE` boolean variable (`false` by default) that allows the pipeline to continue even if Common Torizon tests fail.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,6 +13,9 @@ variables:
   ENABLE_XRAY:
     value: "false"
     description: "Whether to send test results to JIRA using XRay"
+  COMMON_TORIZON_ALLOW_FAILURE:
+    value: "false"
+    description: "Whether to allow the pipeline to continue upon failures in Common Torizon tests."
 
   DEBIAN_RELEASE: "bullseye-slim"
   # Pipeline Settings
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -376,6 +379,19 @@ test-torizoncore-common-intel:
...
+      when: never
+    # The following condition allows the pipeline to continue even if tests fail.
+    # In such cases, subsequent steps let the developer decide, at their discretion,
+    # whether to proceed with the release.
+    - if: '$COMMON_TORIZON_ALLOW_FAILURE =~ /^true$/i'
+      when: on_success
+      allow_failure: true
+    # In all other cases the pipeline continues only if there no failures.
+    - when: on_success
+      allow_failure: false
```
rborn-tx, line 401: - How about creating a `.test-base-common` to serve as base for Common Torizon jobs?

- Don't we need to have the following rules from `.test-base` here as well?

```
    # Following condition prevents execution with custom TorizonCore images (triggered by Jenkins).
    - if: $BUILD_MACHINE
      when: never
    # The following condition allows the pipeline to continue for official releases even if tests
    # fail. In such cases, subsequent steps let the developer decide, at their discretion, whether
    # to proceed with the release.
    - if: '$RELEASE_TYPE == "official"'
      when: on_success
      allow_failure: true
```

rborn-tx (commented):
I've raised some questions.

.gitlab-ci.yml
lucas-akira: Both points make sense to me. I'll apply them, thanks!

PR #115, opened 2026-01-02
tests/integration: Allow absence of overlays on dto tests

rborn-tx:
Minor changes to allow testing with latest machines:

- Change one of the "dto list" tests to not require the listing of overlays; this prevents failures when the committed device-trees tarball does not have any overlays for the machine being tested.
- Add more degbug output messages to help identify the cause of the error when testing "dto list".

Related-to: TCB-810


lucas-akira (approved):
LGTM

PR #116, opened 2026-01-06
 Introduce support for changing the splash screen when the kernel is in the FIT format

lucas-akira:
Similarly to previous implementations to support FIT kernel images for other commands like "dt" and "dto", TCB applies the corresponding modifications to the "kernel" changes directory if it detects that the unpacked image has the kernel in FIT format, including related artifacts like the PNG image path.

These modifications would be applied in a "union" command just like the current implementation for non-FIT images. The general workflow to use the "splash" command with FIT images should be basically identical to non-FIT ones, as well as its corresponding entry in tcbuild.yaml when using the "build" command.

Related-to: TCB-757

---

~~Basic implementation is done but I still need to update the tests, so I'm leaving this as a draft PR for now.~~

I've updated the tests, so the PR is now ready for review.

### DIFF
--- a/tcbuilder/backend/initramfs.py
+++ b/tcbuilder/backend/initramfs.py
@@ -1,0 +1,348 @@
+"""
+This module offers the `UnpackedInitramfs` class to manipulate an unpacked initramfs file.
+"""
+
+import logging
+import os
+import subprocess
+import shutil
+import shlex
+
+from tcbuilder.errors import \
+    (InvalidArgumentError, InvalidDataError, PathNotExistError, TorizonCoreBuilderError)
+from tcbuilder.backend.common import \
+    (get_storage_dir, is_file_type_fit, OSTREE_ROOT_DEPLOY_PATH)
+from tcbuilder.backend import kernel as kernel_be
+from tcbuilder.backend import splash as splash_be
+from tcbuilder.backend import ostree
+from tcbuilder.backend.kernelfit import KernelFit
+
+log = logging.getLogger("torizon." + __name__)
+
+MAX_INITRAMFS_FILE_SIZE = 64*1024*1024
+# Initramfs binary has the same root subdir path as the kernel
+OSTREE_INITRAMFS_SUBDIR_PATH = kernel_be.OSTREE_KERNEL_SUBDIR_PATH
+OSTREE_INITRAMFS_FILENAME = "initramfs.img"
+
+UNPACKED_INITRAMFS_DIR = "initramfs_{suffix}"
+
+
+def check_initramfs_data_size(data):
+    """Check if data is above the maximum defined initramfs file size."""
+
+    if len(data) > MAX_INITRAMFS_FILE_SIZE:
+        raise InvalidDataError(
+            f"New initramfs file size of {data/1024:.2f} kiB would exceed "
+            f"limit of {MAX_INITRAMFS_FILE_SIZE/1024:.2f} kiB. Aborting.")
+
+
+def get_initramfs_subdir():
+    """Get the versioned initramfs path."""
+
+    storage_dir = get_storage_dir()
+    src_sysroot_dir = os.path.join(storage_dir, "sysroot")
+    src_sysroot = ostree.load_sysroot(src_sysroot_dir)
+    csum, _ = ostree.get_deployment_info_from_sysroot(src_sysroot)
+    kernel_version = ostree.get_kernel_version(src_sysroot.repo(), csum)
+    return OSTREE_INITRAMFS_SUBDIR_PATH.format(kver=kernel_version)
+
+
+def find_initramfs_in_sysroot():
+    """
+    Find initramfs binary path in the unpacked image sysroot. Raises an error if
+    it cannot find it.
+
+    :returns: String with absolute path of the initramfs binary inside sysroot.
+    """
+
+    storage_dir = get_storage_dir()
+    sysroot_path = os.path.join(storage_dir, "sysroot")
+
+    # As the initramfs path is composed of directories named after the deployment commit and the
+    # kernel version, get them via OSTree metadata.
+    sysroot_obj = ostree.load_sysroot(sysroot_path)
+    csum, _ = ostree.get_deployment_info_from_sysroot(sysroot_obj)
+    kernel_version = ostree.get_kernel_version(sysroot_obj.repo(), csum)
+    # Add deployment index part to checksum string, which is 0 for a new deployment
+    csum += ".0"
+
+    initramfs_path = os.path.join(
+        OSTREE_ROOT_DEPLOY_PATH.format(csum=csum),
+        OSTREE_INITRAMFS_SUBDIR_PATH.format(kver=kernel_version)
+    )
+    initramfs_path = os.path.join(sysroot_path, initramfs_path, OSTREE_INITRAMFS_FILENAME)
+
+    if os.path.isfile(initramfs_path):
+        log.debug(f"Initramfs found at '{initramfs_path}'")
+    else:
+        raise PathNotExistError("Initramfs not found in unpacked sysroot. Aborting.")
+
+    return initramfs_path
+
+
+def find_initramfs_in_changes_dir(changes_dir, basename=None):
+    """Find path of initramfs binary in the given changes directory."""
+
+    initramfs_subdir = get_initramfs_subdir()
+    initramfs_src_dir = os.path.join(changes_dir, initramfs_subdir)
+    initramfs_src_path = os.path.join(initramfs_src_dir, basename or OSTREE_INITRAMFS_FILENAME)
+    if os.path.isfile(initramfs_src_path):
+        log.debug(f"Initramfs found at '{initramfs_src_path}'", )
+    else:
+        log.debug(f"Initramfs not found at '{initramfs_src_path}'")
+        return None
+
+    return initramfs_src_path
+
+
+def _get_initramfs_path(source):
+    """Get initramfs file path according to source"""
+
+    if source == "sysroot":
+        return find_initramfs_in_sysroot(), source
+
+    changes_dir = splash_be.get_splash_changes_dir()
+    changes_path = find_initramfs_in_changes_dir(changes_dir)
+    if source == "auto":
+        if changes_path:
+            return changes_path, "changes"
+        return find_initramfs_in_sysroot(), "sysroot"
+
+    if source == "changes":
+        return changes_path, source
+    raise InvalidArgumentError(f"_get_initramfs_path(): invalid initramfs source '{source}'")
+
+
+def _get_kernel_path(source):
+    """Get kernel file path according to source"""
+
+    if source == "sysroot":
+        return kernel_be.find_kernel_in_sysroot(), source
+
+    changes_dir = kernel_be.get_kernel_changes_dir()
+    changes_path = kernel_be.find_kernel_in_changes_dir(changes_dir)
+    if source == "auto":
+        if changes_path:
+            return changes_path, "changes"
+        return kernel_be.find_kernel_in_sysroot(), "sysroot"
+
+    if source == "changes":
+        return changes_path, source
+    raise InvalidArgumentError(f"_get_kernel_path(): invalid kernel source '{source}'")
+
+
+def extract_initramfs_from_fit(fit):
+    """Extract initramfs data from a FIT object and save it to a file."""
+
+    initramfs_name, initramfs_data = fit.extract_default_ramdisk()
+
+    changes_dir = kernel_be.get_kernel_changes_dir()
+    os.makedirs(changes_dir, exist_ok=True)
+    initramfs_path = os.path.join(changes_dir, initramfs_name)
+    with open(initramfs_path, "wb") as fhandle:
+        fhandle.write(initramfs_data)
+    log.debug(f"Extracted initramfs node '{initramfs_name}' inside FIT to {initramfs_path}")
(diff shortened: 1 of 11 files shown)


tcbuilder/backend/kernelfit.py
```diff
@@ -629,6 +642,81 @@ def extract_dtb(self, file_name: str, *, overlay=False) -> bytearray:
         img_data = bytes(img_data)
         return img_data
 
+    def update_ramdisk(self, ramdisk_name: str, data: Union[bytes, bytearray]) -> None:
+
```
rborn-tx, line 653: Could we have a short description here?

rborn-tx (commented):
@lucas-akira This implementation looks sensible to me. Nevertheless, I was thinking a bit ahead and wanted to propose something for your consideration. I'll create a thread with my proposal soon, okay?

tcbuilder/backend/kernelfit.py
lucas-akira: Forgot to put one there, good catch.

rborn-tx:
### Status

Today we have the `splash` command that performs its work by creating a directory structure containing the watermark file, builds a cpio file from that structure and creates a new ramdisk by concatenating the original one with the built cpio file; the new ramdisk is then placed into a changes directory to be consumed by the `union` command.

The original ramdisk is taken from the OSTree base commit directly (i.e. from the OSTree repo) rather than the deployment directory present in the sysroot; it should be noted that this behavior differs from all other commands.

The concatenation (splicing) operation also uses OSTree-specific operations.

### Problems

1. The output ramdisk will contain the original splash image plus the new one; customizing an already customized image would add another one, and so one. So, splash images could pile up into the ramdisk file which is a file that has a relatively small size limit.
2. The implementation is too tied to OSTree operations making it harder to generalize and reuse.
3. The handling of the non-FIT and FIT cases has little in common.

### Proposal

First of all I wanted to point out that we can already foresee some new functionalities in TCB that will impact the contents of the ramdisk. For example, we know some users need to make changes to the `plymouth` configuration for different purposes such as hiding or displacing the spinner icon; this recently led us to the idea of creating a `splash-config` command whose work is ongoing. With Secure Boot, particularly with ECoT images, we'll need to update the key inside the ramdisk responsible for validating the composefs file when mounting the root filesystem. Also, for generating the ECoT "transitional" image, a configuration file inside the ramdisk will require changes.

Given the above, I believe it would be very beneficial for us to try and make changes to the ramdisk easier to handle internally in the code.

For that I would propose structuring the `splash` command like this:

```
rd_path = unpack_ramdisk(source="auto")
change_splash_screen(rd_path, ...)
repack_ramdisk(rd_path)
```

The idea is to put all non-FIT/FIT and OSTree-related logic into the `unpack_ramdisk()` and `repack_ramdisk()` functions. In the future, all commands impacting the ramdisk would follow the exact same pattern.

By doing so, the `change_splash_screen()` function would perform its operations (replacing a PNG file somewhere) directly into a `rd_path` directory and wouldn't need to care about anything related to FIT/non-FIT and even OSTree. Therefore, its implementation would become trivial. And, notice this approach would solve the problems that I listed before.

Unless I'm missing something, the implementation of `splash-config {set,dump}` would similarly be quite simple and if we wanted to extend `splash` to allow replacing the images in general, the same would be true. Notice that the approach we have today of concatenating cpio files, does not even allow removing files from the ramdisk. So the proposed approach would solve that as well.

Now, let me outline what `unpack_ramdisk(source="auto")` woud do:

- non-FIT case: (a) Find the ramdisk image in the `splash` changes directory or, if not found, in the deployment directory (sysroot) and (b) unpack it into a work directory in storage (say `/storage/ramdisk_workdir_<suffix>`).
- FIT case: (a) Find the kernel image in the `kernel` changes directory or, if not found, in the deployment directory (sysroot), (b) extract the ramdisk from it and and (c) unpack the ramdisk into a work directory in storage (again `/storage/ramdisk_workdir_<suffix>`).
- The returned value would be the path to the work directory.
- The parameter source would limit where the search for the ramdisk/kernel would occur; if set to "sysroot", the search would be only in the deployment dir in the sysroot; when set to "changes", only the appropriate changes directory would be checked.
- The `<suffix>` would indicate where the ramdisk/kernel came from to allow simultaneous unpacking of both ramdisk "versions" ("sysroot" and "changes"); this might be helpful if one wanted to compare the two ramdisks, for example.

And `repack_ramdisk(rd_path)` would:
- Recreate a ramdisk cpio file from the contents of the directory specified by `rd_path`.
- non-FIT case: (a) Copy the modified ramdisk into the `splash` changes directory.
- FIT case: (a) Find the kernel image in the `kernel` changes directory or, if not found, in the deployment directory (sysroot); (b) copy that file to the `kernel` changes directory; (c) update that file with the new ramdisk.
- Remove the work directory (defined by `rd_path`).

@lucas-akira @jsrc27 That's the proposal. So, let me ask:

- What do you guys think?
- Do you foresee issues that would prevent its implementation?
- Do you agree that it would simplify our next implementations? In particular, regarding the new `splash-config` command?

lucas-akira:
Considering that the new TCB features we're discussing will also modify the initramfs in various ways, I think it makes sense to unpack it and directly modify its contents. The ability to remove files from the ramdisk would also be good to have.

On my side we can go forward with this proposal.

jsrc27:
> The <suffix> would indicate where the ramdisk/kernel came from to allow simultaneous unpacking of both ramdisk "versions" ("sysroot" and "changes"); this might be helpful if one wanted to compare the two ramdisks, for example.

How would this point be used in practice exactly? I'm having trouble picturing it. If a ramdisk is found in changes then we want to unpack from the sysroot of course. But, if there is a ramdisk already present in changes, meaning at least one splash related command has already been ran prior, then we always want to unpack from changes. That way the customizations can be accumulated, correct?

Or is there a use-case where the ramdisk in sysroot would be desired even if a ramdisk exists in changes?

 I guess I ask because of this other point:

> Remove the work directory (defined by rd_path).

If the work directory gets removed at the end of this process then how can we accumulate more than one?

> non-FIT case: (a) Copy the modified ramdisk into the splash changes directory.

I assume if there is already a modified ramdisk in changes due to a previous command, we overwrite it with the latest one always, yes?

rborn-tx:
@jsrc27 Thanks for the questions! Here are the answers:

> > The  would indicate where the ramdisk/kernel came from to allow simultaneous unpacking of both ramdisk "versions" ("sysroot" and "changes"); this might be helpful if one wanted to compare the two ramdisks, for example.
> 
> How would this point be used in practice exactly? I'm having trouble picturing it. If a ramdisk is found in changes then we want to unpack from the sysroot of course. But, if there is a ramdisk already present in changes, meaning at least one splash related command has already been ran prior, then we always want to unpack from changes. That way the customizations can be accumulated, correct?
>
> Or is there a use-case where the ramdisk in sysroot would be desired even if a ramdisk exists in changes?
> 
> I guess I ask because of this other point:
> 
> > Remove the work directory (defined by rd_path).

For example, imagine a command that compares the original and the new plymouth file. It would:

```
rd_path_org = unpack_ramdisk(source="sysroot")
rd_path_new = unpack_ramdisk(source="changes")
# then compare `${rd_path_org}/.../spinner.plymouth` and `${rd_path_new}/.../spinner.plymouth`
# in this case repack_ramdisk() wouldn't be called because this command does not change anything
shutil.rmtree(rd_path_org)
shutil.rmtree(rd_path_new)
```

Another example: a command that resets the plymouth file to its original contents:

```
rd_path_org = unpack_ramdisk(source="sysroot")
rd_path_new = unpack_ramdisk(source="changes")
# then copy `${rd_path_org}/.../spinner.plymouth` over to `${rd_path_new}/.../spinner.plymouth`
repack_ramdisk(rd_path_new)
shutil.rmtree(rd_path_org)
```

In both examples, I'm assuming the ramdisk in the "changes" already exists. The actual code should handle the case where it doesn't.

> If the work directory gets removed at the end of this process then how can we accumulate more than one?
> 
> > non-FIT case: (a) Copy the modified ramdisk into the splash changes directory.
> 
> I assume if there is already a modified ramdisk in changes due to a previous command, we overwrite it with the latest one always, yes?

Yes. To be clear, what would happen in a sequence of standalone commands from `unpack` to `deploy` would look like this:

1. `images unpack`: The deployment/sysroot is populated and changes are cleared.
2. `splash`: In this first execution, the statement `rd_path = unpack_ramdisk("auto")` would find the ramdisk in the **deployment** and unpack it into directory `ramdisk_workdir_sysroot/`; the watermark is replaced; the call `repack_ramdisk(rd_path)` would generate the new ramdisk and place it into the changes dir.
3. `splash-config`: The statement `rd_path = unpack_ramdisk("auto")` would find the ramdisk in the **changes** directory and unpack it into directory `ramdisk_workdir_changes/`; the plymouth files gets replaced; the call `repack_ramdisk(rd_path)` would generate the new ramdisk and overwrite the one in changes dir.
4. `union` would make a commit containing the ramdisk (or the kernel in the FIT case) from the changes dir.
5. `deploy` generates the installer image.

rborn-tx:
@lucas-akira Since we've merged https://github.com/torizon/torizoncore-builder/pull/118, this PR got conflicts which need manual solution.

I've done that already: https://github.com/rborn-tx/torizoncore-builder/tree/TCB-757-rebased

Notice I left a separate commit with changes adapting the code to use the new approach of not passing the `storage_dir` around with the new code that you wrote as well.

tcbuilder/backend/initramfs.py
```diff
@@ -0,0 +1,315 @@
...
+
+        self.unpacked_initramfs_dir = os.path.join(
+            get_storage_dir(),
+            UNPACKED_INITRAMFS_DIR.format(suffix=suffix)
+        )
+
+    def __enter__(self):
+        """Unpack initramfs contents to storage."""
+
+        if not self.src_initramfs_file:
```
rborn-tx, line 274: I'd suggest a minor change to this interface that would make it more flexible:

- Rather than returning a path, we could return `self`; with this we could provide other methods related to the context.
- Provide a method to get the path currently returned by `__enter__()`.

Then the usage would become:

```
    with UnpackedInitramfs(source="auto") as rdm:
        rdm.get_path()  # in place of irfs_path
```

By doing that, we'd be able to add more functionality to the context manager when needed.

Then, we could already leverage that flexibility:

1. What if we had a method to indicate whether or not the repacking is needed? For example, if one has a command that may or may not modify the contents of the ramdisk they could do this:

```
    with UnpackedInitramfs(source="auto") as rdm:
        #...
        if no_changes_made_to_ramdisk:
            rdm.set_repack(False)
```

2. Related to this, we could also consider the case where one knows beforehand that no repack is needed (because the command only inspects the ramdisk). In that case, we could have a parameter in the context manager ctor:

```
    with UnpackedInitramfs(source="auto", repack=False) as rdm:
                                        # ^^^^^^^^^^^^
```

What do you think?

tcbuilder/backend/splash.py
```diff
@@ -5,62 +5,38 @@
...
 
-    # implement cat `ostree cat ref /usr/lib/modules/${kver}/initramfs.img`
-    # /storage/splash/initrmafs.splash > /storage/splash/usr/lib/modules/${kver}/initramfs.img
-    ret, root, _commit = repo.read_commit(ostree.OSTREE_BASE_REF)
-    if not ret:
-        raise TorizonCoreBuilderError(f"Error couldn't reat commit: {ostree.OSTREE_BASE_REF}")
+    plymouth_theme_abspath = os.path.join(
+        root_dir,
+        PLYMOUTH_THEME_PATH
+    )
```
rborn-tx: Nitpick: how about collapsing this into a single line?

rborn-tx (commented):
This looks great!

Nonetheless, I've added a suggestion regarding the interface of the context manager.

jsrc27 (approved):
No major additional comments from my side.

tcbuilder/backend/splash.py
lucas-akira: This was a larger statement at some point, but I condensed it and forgot to collapse it into a single line. Thanks, I'll fix it.

tcbuilder/backend/initramfs.py
lucas-akira: I haven't thought of the use case where initramfs contents could only be read and not modified, so having the option to skip the repacking step makes sense.

I'm also in favor of returning the object directly instead of the path if that makes things more flexible. I'll apply the suggestions then.

tcbuilder/backend/initramfs.py
```diff
@@ -0,0 +1,348 @@
...
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        """Repack initramfs contents in storage to changes directory."""
+
+        if not self.unpacked_initramfs_dir:
+            log.debug("__exit__(): Unpack not done, nothing to do.")
+            return
+        try:
+            if self.repack:
```
rborn-tx: We could avoid the extra nesting by doing:

```
if not self.repack:
    log.debug(f"__exit__(): Skipping repack of {self.unpacked_initramfs_dir}")
    return
```

rborn-tx (commented):
Did a pass on the changes and added a suggestion.

tcbuilder/backend/initramfs.py
lucas-akira: Yeah, that would be better. I'll change it.

rborn-tx (approved):
Great work! I'll go ahead and merge it.

PR #119, opened 2026-01-14
Refactor Dockerfile to follow some good practices

rborn-tx:
Minor improvements to the `torizoncore-builder.Dockerfile`:

- Put each package being installed via apt-get in a separate line; this makes it easier for developers to trace the package additions and removals over time.
- Sort the packages being installed alphabetically.
- Collapse multiple installs in a row into a single install command.

### DIFF
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -340,7 +340,9 @@ jobs:
     # Only run if not manually cancelled, test-setup succeeds,
     # and after amd64-build-test-torizoncore5 and amd64-build-test-torizoncore6,
     # no matter their results.
+    # TODO: Update the input image and re-enable.
     if: |
+      false &&
       (success() || failure()) &&
       contains(needs.test-setup.result, 'success')
     needs: [test-setup, amd64-build-test-torizoncore5, amd64-build-test-torizoncore6]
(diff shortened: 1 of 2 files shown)


jsrc27 (approved):
LGTM. With regards to the failed test in the pipeline though, that looks completely unrelated to what was done in this PR. Any ideas where that came from?

rborn-tx:
> LGTM. With regards to the failed test in the pipeline though, that looks completely unrelated to what was done in this PR. Any ideas where that came from?

I know these errors with the raspi4 images have been happening for quite some time now but honestly I haven't looked into them thus far. But now, looking at the logs I see the error is related to disk space. Perhaps we need to review the tests to avoid leaving behind large WIC images in the tests working directory to hopefully not hit some disk space limit...

lucas-akira:
The Raspberry Pi 4 tests are failing for some time now, but given that the tested image is an older version that we don't support, I wasn't actively looking into it.

Maybe we should remove this and other related tests until we update them to use newer, supported images.

lucas-akira (approved):

rborn-tx:
> The Raspberry Pi 4 tests are failing for some time now, but given that the tested image is an older version that we don't support, I wasn't actively looking into it.
> 
> Maybe we should remove this and other related tests until we update them to use newer, supported images.

Yeah, makes sense. I've added a commit causing the tests for the Raspberry Pi 4 to be skipped on GitHub.

rborn-tx:
I'll go ahead and merge this since my only addition after the approval was the disabling of the Raspberry Pi 4 test job.

PR #125, opened 2026-02-04
gitlab-ci: fix manual trigger for early-access and official pipelines

lucas-p-bernardes:
The desired state pipelines in GitLab CI was as follows:
- early-access: If all tests pass, it automatically deploys the early-access tag. If any test fails, the maintainers must still have the option to proceed with the tag deployment, since they may decide that the failure is not critical.

- official/internal: The maintainers must always manually decide whether to deploy the tag, regardless of the state of the tests.

However, GitLab CI’s "when: manual" keyword currently only allows manual execution if all previous stages have succeeded. This is incompatible with the desired behavior for both pipelines, where the maintainer must be able to choose to continue with the later stages even if earlier stages have failed.

Therefore, this commit ensures the desired behavior for these pipelines, allowing the maintainer to perform the tag deployments even if previous stages have failed.

Implementation details:
- Split jobs into "*-auto" and "*-manual" variants to handle different trigger logic.
- Implemented "needs: [<job>]" in manual jobs. This explicitly decouples them from the global stage order (specifically the test stage).

Related-to: TCB-751

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -192,15 +192,7 @@ build-arm64:
     # Following condition prevents execution with custom TorizonCore images (triggered by Jenkins).
     - if: $BUILD_MACHINE
       when: never
-    # The following condition allows the pipeline to continue for official releases even if tests
-    # fail. In such cases, subsequent steps let the developer decide, at their discretion, whether
-    # to proceed with the release.
-    - if: '$RELEASE_TYPE == "official"'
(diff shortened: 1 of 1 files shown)


jsrc27 (approved):
LGTM

rborn-tx (approved):
Nice job!

lucas-akira (approved):
LGTM

PR #127, opened 2026-02-10
support parallel downloads of root.json

bclouser:
especially useful for repositories that have over 1k root.jsons :)

### DIFF
--- a/tcbuilder/backend/platform.py
+++ b/tcbuilder/backend/platform.py
@@ -12,6 +12,7 @@ import re
 import sys
 import shutil
 import subprocess
+from concurrent.futures import ThreadPoolExecutor, as_completed
 
 from fnmatch import fnmatchcase
 from io import BytesIO, TextIOWrapper
@@ -106,6 +107,7 @@ COMMIT_CHECK_TIMEOUT = int(os.environ.get("COMMIT_CHECK_TIMEOUT", "60"))
 FETCH_READ_TIMEOUT = int(os.environ.get("FETCH_READ_TIMEOUT", "60"))
 UPDATE_DESC_TIMEOUT = int(os.environ.get("UPDATE_DESC_TIMEOUT", "60"))
 UPLOAD_DELTA_TIMEOUT = int(os.environ.get("UPLOAD_DELTA_TIMEOUT", "240"))
+PARALLEL_DOWNLOADS = int(os.environ.get("PARALLEL_DOWNLOADS", "10"))
 
 
 def load_metadata(fname, ftype=None, maxlen=DEFAULT_METADATA_MAXLEN):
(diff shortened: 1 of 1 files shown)


tcbuilder/backend/platform.py
```diff
@@ -1207,21 +1223,32 @@ def fetch_director_metadata(lockbox_name, director_url, dest_dir, access_token=N
...
+        for count, future in enumerate(as_completed(futures), 1):
+            fname = futures[future]
+            try:
+                future.result()
+            except FetchError as exc:
+                log.warning(str(exc))
+                # pylint: disable-next=raise-missing-from
+                raise TorizonCoreBuilderError(
+                    f"Error: Could not fetch metadata file '{fname}' from server")
+            log.info(f"Fetched director root metadata ({count} of {total})")
```
jsrc27, line 1251: This has nothing to really do with the changes, but more of a question of consistency now that I'm looking at the logic for fetching the root metadata.

Why was the logic slightly different for fetching image-repo root metadata vs director root metadata. For example, we have a try/except block in the director root metadata fetching logic, but not for image-repo. Even though they do essentially the same thing as far as I can tell. Along with other minor inconsistencies.

@rborn-tx or @lucas-akira, do you guys know why these differed in the past? I couldn't find any good reason. If not then I'd prefer the logic to be the same/consistent between the two. I guess there's no strong reason to, it just raises weird questions about why they're different at all.

jsrc27 (commented):
No major comments, just a small comment about an inconsistency I noticed. But this inconsistency was present before these changes as well.

tcbuilder/backend/platform.py
lucas-akira: From what I can tell by looking at the commit history, the fetching logic for all metadata was added at once in a single commit from years ago, so I'm assuming this was done this way without a particular reason.

rborn-tx: Actually, the two functions were introduced in separate commits:

- https://github.com/torizon/torizoncore-builder/commit/953be7bd7249f0a742a7e1164a3abf2dd8bcc5da
- https://github.com/torizon/torizoncore-builder/commit/e238283d9de8f6d25f289b08cbdde6515f4c4751

I honestly didn't recall the reason for the differences but I believe the commit message for the second commit above explains it. The situation described was temporary so we probably could change things now to handle both image-repo and director in the very same way...

rborn-tx (approved):
@bclouser LGTM. Nice addition, btw.

Since you mentioned having over 1k `root.json` files, one potential improvement to consider is the use of session objects (https://requests.readthedocs.io/en/latest/user/advanced/#session-objects). Of course, this is entirely at your discretion.

tcbuilder/backend/platform.py
lucas-akira: OK, so these functions go back to a previous source file before being moved to `backend/platform.py`, and back when the offline update feature was still being tested and implemented.

On my side I'm in favor of making things consistent now since it would improve organization and it's simple to do.

lucas-akira:
Hey @bclouser I just saw that the Pylint test in the pipeline is failing. Can you take a look at your additions and make the necessary changes?

tcbuilder/backend/platform.py
jsrc27: Alright thanks for the clarification everyone.

@bclouser I guess even though this is tangential to your changes. It would be appreciated if you could make the logic for fetching root metadata consistent between image-repo and director. Maybe even just factor it out to a common function.

That said it's not vital for the PR, and we can probably also just "fix" it later.

PR #128, opened 2026-02-17
Introduce support for customizing ECoT images

rborn-tx:
Add support for customizing and signing ECoT (Extended Chain-of-Trust) images, that is, those built with the root filesystem protection enabled.

**Summary of Changes**

1. **secboot Refactors and Improvements**
   - Reordered secboot code to group bootloader and kernel signing functions, matching CLI sub-command order.
   - Replaced references from "fitImage" to "FIT image" for consistency.
   - Refactored CLI layer in secboot for readability and reusability.
   - Changed kernel key directory to be an optional switch in CLI rather than a positional argument (**breaking** change).
   - Made CST directory a CLI switch rather than a positional argument (**breaking** change).
   - Improved messaging, help output, and consistency.
   - Updated order of `sign-kernel` and `sign-bootloader-hab` tests and processing.
   
2. **OSTree and composefs Support**
   - Built ostree from source to enable composefs and ed25519 signing.
   - Added support for signing OSTree commits (ECoT/composefs images).
   - Integrated OSTree signing and key update features into "build" command.
   - Added support for updating OSTree pubkey in initramfs via secboot.
   
3. **initramfs and kernelfit Enhancements**
   - Allowed users to specify fixed paths for initramfs operations.
   - Disabled automatic deletion of hash props in FIT image ramdisk update due to signing errors (left a TODO to determine root cause).
   
4. **Tests and Test Infrastructure**
   - Added/updated tests to support/skip cfs (composefs) and non-cfs images.
   - Made test titles clearer.
   - Refactored variable names and test code for clarity.
   - Adapted tests for cfs-enabled images and new CLI switches.
   - Improved logging and started registries during "build" tests, improved failure diagnostics.
   
5. **Union Command Updates**
   - Added support for signing OSTree commits in union module.
   - Dropped deprecated parameters (--extra-changes-directory, --union-branch).
   - Reindented union CLI help text for consistency.
   
6. **General Code Maintenance**
   - Refactored large functions for readability (e.g., import_local_image).
   - Removed redundant informational messages.
   - Made minor improvements, e.g., logging and string consistency.

Please see the individual commit messages for details.

Related-to: TCB-817  

P.S.: Since this is a somewhat big PR, I'd suggest reviewing the commits one-by-one.


tcbuilder/backend/union.py
```diff
@@ -61,9 +61,50 @@ def process_whiteouts(mtree, path="/"):
         process_whiteouts(submt, os.path.join(path, dirname))
 
 
+def _sign_commit(repo, commit, ostree_key):
+    """Sign a commit.
+
+    :param repo: OSTree repository (OSTree.Repo).
+    :param commit: Checksum of commit to be signed.
+    :param ostree_dir: Wrapper object for key information (OSTreeKey).
```
lucas-akira: Here it should be `ostree_key`.

tcbuilder/cli/union.py
```diff
@@ -189,18 +192,68 @@ def make_dirs_labels(changes_dirs, stor_pref, work_pref):
...
+                "support for the root filesystem protection. Aborting.")
+
+    if ostree_key is None:
+        return None
+
+    if ostree_key_dir is not None and not os.path.isdir(ostree_key_dir):
+        raise PathNotExistError(
+            f"Error: OSTree keys directory '{ostree_key_dir}' does not exist. Aborting.")
+
+    ostree_key_dir = ostree_key_dir or "."
```
jsrc27, line 223: So if the user does not pass `--ostree-key-dir` then TCB defaults to searching the current working directory for the keys?

Should we note this behavior in the help dialog for the switch itself?

tcbuilder/cli/secboot.py
```diff
@@ -62,7 +65,47 @@ def _parse_kernel_key_arg(kernel_key):
...
+                "support for the root filesystem protection. Aborting.")
+
+    if ostree_key is None:
+        return None
+
+    if ostree_key_dir is not None and not os.path.isdir(ostree_key_dir):
+        raise PathNotExistError(
+            f"Error: OSTree keys directory '{ostree_key_dir}' does not exist. Aborting.")
+
+    ostree_key_dir = ostree_key_dir or "."
```
jsrc27, line 172: Similar here, should we note the default behavior in the help dialog for the switch itself?

jsrc27 (commented):
Gave it an initial pass. Some minor comments, but overall did not notice anything major.

One thought though as I was going through this. Should we consider using the "Extended chain-of-trust"/ECoT terminology in some of the help/log messages in TCB? We do use this term frequently in our documentation/marketing on secure boot.

I just have a minor concern that users may not immediately connect root filesystem protection to ECoT, or vice versa. Especially with how many terms we throw around regarding secure-boot. Could cause some confusion.

tcbuilder/backend/ostree.py
```diff
@@ -32,22 +32,104 @@
...
+
+    org_conf = dst_repo.get_config()
+    new_conf = src_repo.copy_config()
+
+    if keep_core:
+        core_group = "core"
+        assert org_conf.has_group(core_group), \
+            f"Source OSTree repository has no group '{core_group}'."
+        assert new_conf.has_group(core_group), \
+            f"Destination OSTree repository has no group '{core_group}'."
```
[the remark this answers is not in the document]
rborn-tx: Nice catch!

tests/integration/functions.sh
```diff
@@ -358,3 +358,101 @@ signing-artifacts-in-unpacked-tezi-image() {
...
+    local repo="/storage/sysroot/ostree/repo/"
+    local prop="ex-integrity.composefs"
+    if status=$(torizoncore-builder-shell "ostree config --repo ${repo} get ${prop}"); then
+	if [[ "${status}" == @(true|yes|1) ]]; then
+	    # composefs is enabled
+	    return 0
+	fi
+	# composefs is not enabled
+	return 1
+    fi
```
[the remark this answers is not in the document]
rborn-tx: According to Bash's manual, in the compound command `[[`, the `==` operator works as if `extglob` is set by default.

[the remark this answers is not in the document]
rborn-tx: But indeed, `res` was unused.

tcbuilder/backend/union.py
```diff
@@ -107,45 +148,57 @@ def commit_changes(repo, ref, changes_dirs, branch_name, *,
...
-        subject = f"TorizonCore Builder union commit created at {isodatetime}"
-
     # GLib.Variant of type "a{sv}" (array of dictionaries), which is the
     # metadata obeject
     newmetadatavar = GLib.Variant.new_array(GLib.VariantType("{sv}"), newmetadata)
 
+    # Add composefs properties (esp. digest) to the commit metadata.
+    if ostree_key is not None:
+        newmetadatavar_dict = GLib.VariantDict.new(newmetadatavar)
+        assert repo.commit_add_composefs_metadata(0, newmetadatavar_dict, root, None)
```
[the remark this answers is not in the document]
rborn-tx: We're accepting that risk for now.

tcbuilder/cli/union.py
rborn-tx: Done in V2 along with the changes for Copilot.

tcbuilder/cli/secboot.py
rborn-tx: Done in V2 along with the changes for Copilot.

tcbuilder/backend/kernelfit.py
```diff
@@ -39,6 +39,13 @@
 
 DELETE_PROPS_ON_DTB_COPY = True
 
+# TODO: Understand why deleting the hash properties is causing "no-space" errors:
+# When running mkimage we get:
+#
+#   Can't set hash 'value' property for 'hash-1' node(FDT_ERR_NOSPACE)
```
lucas-akira, line 45: Does this issue happen only on images with composefs support, or more generally for images with a kernel in FIT format?

I did a quick test with this code and that didn't happen when I locally ran the `splash` tests on a pre-signed `tdx-signed` Torizon OS minimal image I got from our Artifactory server.

lucas-akira (commented):
Lots of code cleanup along with the new feature, nice. I have a few points but nothing very serious.

tcbuilder/backend/union.py
rborn-tx: Nice catch!

tcbuilder/backend/kernelfit.py
rborn-tx: I'm not 100% sure, but I think I tested with `tdx-signed` as well (i.e. no composefs) and got the same error. I tried adding padding to the FIT image when saving it but it didn't solve the problem. But I didn't want to spend much time investigating it, since it was a bit unrelated to what I was doing.

In your test, was the splash screen file bigger than the original one?

rborn-tx:
@jsrc27 @lucas-akira One thing I forgot to mention: I built fully customized images for `TC5` (yes, you read it right), `TC6`, `TOS7 x {unsigned, tdx-signed, torizon-signed}` and checked the customizations on an actual device, including watching the splash screen image on a display -- everything worked!

That was something that I wanted to do since the upgrade from `bullseye` to `trixie` in order to to increase our confidence in the latest version of TCB.

tcbuilder/backend/union.py
rborn-tx: Done

tcbuilder/backend/kernelfit.py
lucas-akira: I took a look at meta-toradex-torizon and the original splash screen has around 16 kiB in size: https://github.com/torizon/meta-toradex-torizon/tree/scarthgap-7.x.y/recipes-core/plymouth/files

I used the image from our test samples (https://github.com/torizon/torizoncore-builder/tree/trixie/tests/integration/samples/splash), which has around 30 kiB, so yes the new splash screen file size is larger than the original one.

lucas-akira: Well, if skipping this step doesn't cause any major issues when re-signing the kernel FIT, then I see no problems keeping this and doing an investigation later.

rborn-tx: As far as I can tell, it shouldn't cause any issues :crossed_fingers: (and the tests went well on a device).

rborn-tx:
@jsrc27 Regarding:

> One thought though as I was going through this. Should we consider using the "Extended chain-of-trust"/ECoT terminology in some of the help/log messages in TCB? We do use this term frequently in our documentation/marketing on secure boot.
>
> I just have a minor concern that users may not immediately connect root filesystem protection to ECoT, or vice versa. Especially with how many terms we throw around regarding secure-boot. Could cause some confusion.

Good point. I also had that question in my mind when writing the error messages...

I ended up using the expression "root filesystem protection" for technical accuracy. Because in theory each protection is independent, an image could potentially have the root filesystem protection while not having the other protections from the previous links (so, it wouldn't technically be an ECoT image). Also, notice that the code does not detect if the image has ECoT coverage - it only checks for the enabling of composefs from where it infers the presence of root filesystem protection.

But, I'm open to ideas...

jsrc27:
> But, I'm open to ideas...

On my side it was just a fleeting thought I had when going through the changes. I'm okay with just proceeding with the log and error messages as is. It's a fairly minor UX detail anyways, and can easily be changed in the future if we find the need to.

jsrc27 (approved):

lucas-akira (approved):

PR #129, opened 2026-02-19
bundle: Make bundle containers unique

jsrc27:
Prior the container names, volumes, and ports used by the bundle command were hard-coded. This could lead to situations where multiple instances on the same system and network could cause conflicts.

Improve this by using unique timestamp based names. As well as searching for random available ports to avoid possible conflicts.

Related-to: TCB-816

### DIFF
--- a/tcbuilder/backend/bundle.py
+++ b/tcbuilder/backend/bundle.py
@@ -4,13 +4,14 @@
 
 import logging
 import os
-import re
 import shlex
 import shutil
+import socket
 import subprocess
 import sys
 import time
 
+from contextlib import closing
 from datetime import datetime
 
 import docker
@@ -69,6 +70,16 @@ def get_compression_command(output_file):
         output_file_tar = f"{output_file}.tar"
 
     return (output_file_tar, command)
+
+
+def get_unused_port():
+    """Returns an unused network port for localhost."""
+
+    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
+        # Binding to port 0 asks the OS to find an available ephemeral port
+        s.bind(('127.0.0.1', 0))
+        # Get the port number the OS assigned
+        return s.getsockname()[1]
 
 
 class DockerManager:
@@ -203,7 +214,9 @@ class DindManager(DockerManager):
 
         self.docker_host = None
         self.dind_volume = None
+        self.volume_name = None
         self.dind_container = None
+        self.timestamp = None
 
     def _wait_certs(self):
         # Wait until TLS certificate is generated
@@ -243,32 +256,20 @@ class DindManager(DockerManager):
         log.info("\nStarting DIND container")
         dind_cmd = ["--storage-driver", "overlay2"]
         ports = None
+        port = get_unused_port()
 
         if network_name == "host":
-            # Choose a safe and high port to avoid conflict with already
-            # running docker instance...
-            port = 22376
+            self.docker_host = f"tcp://127.0.0.1:{port}"
             dind_cmd.append(f"--host=tcp://0.0.0.0:{port}")
-
-            if "DOCKER_HOST" in os.environ:
-                # In case we use a Docker host, also connect to that host to
-                # reach the DIND instance (Gitlab CI case)
-                docker_host = os.environ["DOCKER_HOST"]
-                results = re.findall(r"tcp?:\/\/(.*):(\d*)\/?.*", docker_host)
-                if not results or len(results) < 1:
-                    raise InvalidArgumentError(f"Regex does not match: {docker_host}")
-                host_ip = results[0][0]
-                self.docker_host = f"tcp://{host_ip}:{port}"
-            else:
-                self.docker_host = f"tcp://127.0.0.1:{port}"
             log.info(f"Using Docker host \"{self.docker_host}\"")
         else:
-            port = 22376
             ports = {f"{port}/tcp": port}
             dind_cmd.append(f"--host=tcp://0.0.0.0:{port}")
 
         # Create the volume to hold the /var/lib/docker data.
-        self.dind_volume = self.host_client.volumes.create(name=self.DIND_VOLUME_NAME)
+        self.timestamp = datetime.now().strftime("-%Y%m%d%H%M%S_%f")
+        self.volume_name = self.DIND_VOLUME_NAME + self.timestamp
+        self.dind_volume = self.host_client.volumes.create(name=self.volume_name)
 
         # The workdir below is for the DinD instance.
         environ = {
@@ -286,7 +287,7 @@ class DindManager(DockerManager):
                 read_only=False
             ),
             docker.types.Mount(
-                source=self.DIND_VOLUME_NAME,
+                source=self.volume_name,
                 type='volume',
                 target='/var/lib/docker/',
                 read_only=False
@@ -306,13 +307,14 @@ class DindManager(DockerManager):
         if dind_env is not None:
             environ.update(dind_env)
 
+        container_name = self.DIND_CONTAINER_NAME + self.timestamp
         run_kwargs = {
             "privileged": True,
             "environment": environ,
             "mounts": mounts,
             "ports": ports,
             "network": network_name,
-            "name": self.DIND_CONTAINER_NAME,
+            "name": container_name,
             "auto_remove": False,
             "detach": True,
             "command": dind_cmd
@@ -341,7 +343,7 @@ class DindManager(DockerManager):
             self.dind_container.reload()
             dind_ip = self.dind_container.attrs \
                 ["NetworkSettings"]["Networks"][network_name]["IPAddress"]
-            self.docker_host = "tcp://{}:22376".format(dind_ip)
+            self.docker_host = "tcp://{}:{}".format(dind_ip, port)
 
     def _stop_container(self):
         log.info("Stopping DIND container")
@@ -440,7 +442,7 @@ class DindManager(DockerManager):
                 read_only=False
             ),
             docker.types.Mount(
-                source=self.DIND_VOLUME_NAME,
+                source=self.volume_name,
                 type='volume',
                 target='/var/lib/docker/',
                 read_only=True
@@ -453,9 +455,10 @@ class DindManager(DockerManager):
 
         # Due to issues with WSL, we are running the container detached and
         # explicitly waiting it to stop.
+        tar_container_name = self.TAR_CONTAINER_NAME + self.timestamp
         _tar_container = self.host_client.containers.run(
             "debian:bullseye-slim",
-            name=self.TAR_CONTAINER_NAME,
+            name=tar_container_name,
             mounts=_mounts,
             command=_tar_command,
             detach=True)


jsrc27:
I did test this locally with and without `--net=host` on the TCB container and it worked fine in both cases. I was hoping to see the results of the pipeline to make sure this also works properly in a CI/CD environment. But it seems the `bundle` related tests here failed due to issues unrelated to my changes.

Seems like the Docker used in the Github pipelines are now using the new default storage-driver `overlayfs`, instead of the expected and supported by TCB `overlay2`.

tcbuilder/backend/bundle.py
```diff
@@ -71,6 +74,36 @@ def get_compression_command(output_file):
...
+            # Binding to port 0 asks the OS to find an available ephemeral port
+            s.bind(('127.0.0.1', 0))
+            # Get the port number the OS assigned
+            return s.getsockname()[1]
+    else:
+        # Can't use the same port 0 trick when the host is remote.
+        # Let's just try to connect to a random port until we find an open one.
+        open_port = ""
+
+        while not open_port:
```
lucas-akira: So here TCB would keep trying to connect to an available port number until it finds one.

Maybe we should add a maximum number of tries and a delay between attempts to avoid possibly overwhelming the host, triggering any connection limits or just hanging indefinitely if the host happens to have most if not all ports closed.

lucas-akira (commented):
Just have one concern about getting an available port number.

tcbuilder/backend/bundle.py
jsrc27: Hmm that's a good point. Since we're just randomly selecting a port here in a large range, the chances of needing more than 1 or 2 guesses would be highly unlikely.

Though in the extreme case something like what you described could happen. Does a 2s delay with up to 10 tries sound sensible? If we still haven't found an available port after 10 rolls of the device, probably should just give up at that point I would assume.

lucas-akira: 10 tries with a 2s delay sounds reasonable to me. Not sure if there's a standard criteria for these values for our use case, maybe it's worth researching a bit to see if there's something we can use as a starting point.

jsrc27: After thinking about it for a bit, 5 is probably a good enough amount of tries for what we're trying to do here. If TCB can't find an open port after 5 random attempts, there's likely something else the matter with the host we're trying to connect to.

lucas-akira: 5 attempts should be OK as well, considering that there's tens of thousands of ports to try.

lucas-akira (approved):
New revision looks good to me.

tcbuilder/backend/bundle.py
```diff
@@ -71,6 +74,44 @@ def get_compression_command(output_file):
...
+                open_port = s.getsockname()[1]
+                s.close()
+                if open_port:
+                    break
+
+            except Exception:    # pylint: disable=broad-except
+                s.close()
+        else:
+            raise OperationFailureError(
+                f"Unable to find free port at DOCKER_HOST: '{host}'")
```
rborn-tx: I'm not sure I understand how this works.

We are trying to find a free port on a certain host, right? If the port was free (but also if the port is protected by a firewall, for example), we wouldn't be able to connect to it (because no program would be listening to it), but here it seems we are considering the port is free if we were able to connect to it. Then we are taking the source port (which was randomly assigned by the OS at the origin) as the free port.

Could you please explain?

Note: [getsockname](https://docs.python.org/3/library/socket.html#socket.socket.getsockname) gets the socket's own address

jsrc27: > Note: [getsockname](https://docs.python.org/3/library/socket.html#socket.socket.getsockname) gets the socket's own addressNote: [getsockname](https://docs.python.org/3/library/socket.html#socket.socket.getsockname) gets the socket's own address

My mistake I assumed this behaved the same no matter the type of connection/bind on the socket. I copied it from the case where host is `localhost`. I didn't notice it in testing cause this did actually spit out seemingly "random" port numbers. But it seems this wasn't really related to the actual connection itself.

Looks like `getpeername` is the remote equivalent: https://docs.python.org/3/library/socket.html#socket.socket.getpeername

> We are trying to find a free port on a certain host, right? If the port was free (but also if the port is protected by a firewall, for example), we wouldn't be able to connect to it (because no program would be listening to it), but here it seems we are considering the port is free if we were able to connect to it.

Hmm I didn't question things at first because in my test it seemed to work. But on closer inspection I think the dummy Docker-in-Docker container I setup to act as the remote `DOCKER_HOST` was just accepting connections always, which made this part work. Though you are correct this isn't a realistic case always.

I looked at the original code and now I'm actually kinda confused. So the `else` chunk that you highlighted here was to handle the case where `DOCKER_HOST` is set in the environment. Looking at the original code this is the relevant chunk:
```
if "DOCKER_HOST" in os.environ:
                # In case we use a Docker host, also connect to that host to
                # reach the DIND instance (Gitlab CI case)
                docker_host = os.environ["DOCKER_HOST"]
                results = re.findall(r"tcp?:\/\/(.*):(\d*)\/?.*", docker_host)
                if not results or len(results) < 1:
                    raise InvalidArgumentError(f"Regex does not match: {docker_host}")
                host_ip = results[0][0]
                self.docker_host = f"tcp://{host_ip}:{port}"
```

Based on the in-line comment this was only added to support the Gitlab CI case. I blindly assumed based on the comment that we do this in our own CI. However looking at our Gitlab CI, I don't think that's the case. We do set `DOCKER_HOST` in the CI environment: https://github.com/torizon/torizoncore-builder/blob/trixie/.gitlab-ci.yml#L26

But this code is specifically for if `DOCKER_HOST` is set in the TCB container environment. Which it doesn't seem like it is.

So now my question is. Was this chunk of code ever being used in TCB? Maybe at one point a long time ago. If this chunk is useless, maybe I just remove the highlighted else block in my own changes then. Cause if that's true then this code was only added for our own internal usage, and it's not being used anymore (if it ever really was).

jsrc27: On even further thought this `DOCKER_HOST` use-case doesn't even really make sense.

If you set `DOCKER_HOST` in the TCB container, this only affects where the Docker-in-Docker (dind), and tar containers get executed. The original TCB container would still be executed on the localhost. But if the dind and tar containers are being executed in a different host than the TCB container, how would one even get the bundle files back to the original environment where the TCB container is being executed?

With the way bundle works it really only makes sense if the TCB container is executed in the same host/environment as the dind and tar containers that create the output. Which makes me even more confused on how this was supposed to work in the past, if at all.

I think I've convinced myself further that this `DOCKER_HOST` use-case should just be dropped and removed from the code entirely. Unless someone can think of a reason for it to remain that makes sense.

rborn-tx: Hmm, good points. And yeah, this whole code is really confusing...

Honestly, looking at the code I can't even tell for sure how things are working in CI :weary:.

For example, how would this piece of code in the `DindManager` ctor work if `DOCKER_HOST` is not set in the TCB environment?

```
self.host_client = docker.from_env()
```

Perhaps we should take the opportunity to better understand this whole topic? Without further analysis I can't really say what to do with the code handling `DOCKER_HOST`.

A legitimate use case that I can imagine for passing `DOCKER_HOST` to TCB is if the Docker daemon was running on a non-standard port. The IP could be anything as long as it belongs to the same host as you said (so in practice the IP wouldn't be much useful but the port would).

jsrc27: I did some testing and research and I think I have a better grasp now. It's still a bit confusing due to the multiple levels of Docker involved with the `bundle` command.

> For example, how would this piece of code in the DindManager ctor work if DOCKER_HOST is not set in the TCB environment?

So based on what I observed and saw here: https://docker-py.readthedocs.io/en/stable/client.html?highlight=from_env#docker.client.from_env

This will setup the client in TCB and use the listed Docker environment variables to further configure the client object. Now if `DOCKER_HOST` isn't set then the default behavior is just the usual Docker socket at `unix:///var/run/docker.sock`. The TCB container can access the Docker daemon outside of itself cause we typically bind-mount this socket into the TCB container.

Now the confusing part is the importing of the Docker environment variables. For example if `DOCKER_HOST` is set **outside** of the TCB container then it has no effect on the client object inside the TCB container. Since the environments in and outside the container are separate. This is a little confusing because while TCB is accessing the Docker daemon outside, but it's using the Docker environment variables inside it's own containerized environment it would seem.

> Honestly, looking at the code I can't even tell for sure how things are working in CI 😩.

In our Gitlab CI we set `DOCKER_HOST: tcp://docker:2375`. But again, this should only affect where the TCB container gets executed. We never pass any `DOCKER_HOST` into the TCB container environment.

This is supported by log messages in CI like:
```
Using Docker host "tcp://127.0.0.1:22376"
Connecting to Docker Daemon at "tcp://127.0.0.1:22376"
```

So the dind container started by the TCB container is using the localhost. The question being localhost where? In Gitlab CI we're basically executing `DOCKER_HOST=tcp://docker:2375 torizoncore-builder ....`. This will cause the TCB container to be executed at `tcp://docker:2375`. So in this context localhost is the localhost of `tcp://docker:2375`. This works just fine as expected since both TCB and the dind container are being ran in the same host environment of `tcp://docker:2375`. The TCB container starts the dind container via the usual Docker socket in `tcp://docker:2375`

I verified this by setting up a dind container and running TCB  with the correct `DOCKER_HOST`, so that it executes inside this dind container. Inside my dind container I could see TCB and the dind container (from `bundle`) running side by side. The command executed successfully with the bundle directory being produced inside my dind container.

But then, this really does mean that this `if "DOCKER_HOST" in os.environ:` section in the original code is not being used at all, since this is based on `DOCKER_HOST` in the TCB container environment, which is not set. So really as far as I can tell this section of code is useless with the way TCB is being used today. We also don't document this internal behavior, so it wasn't for user use.

> A legitimate use case that I can imagine for passing DOCKER_HOST to TCB is if the Docker daemon was running on a non-standard port.

Yeah it's a niche use-case but possible. This would mean someone has setup their Docker environment such that the daemon is only accessible on a specific TCP address, and not the default Docker socket.

Off the top of my head you'd probably need to set `DOCKER_HOST` for the TCB container. That way it gets executed using the right port. But then you'd also need to set `DOCKER_HOST` in the TCB alias to pass it into the TCB container as well. That way the dind container from `bundle` would get executed using the correct port as well. In theory this would already work today I believe. Since `self.host_client = docker.from_env()` would grab the `DOCKER_HOST` set in the TCB alias. This would then be used to execute the dind container at the designated TCP address.

rborn-tx: > But then, this really does mean that this if "DOCKER_HOST" in os.environ: section in the original code is not being used at all, since this is based on DOCKER_HOST in the TCB container environment, which is not set. So really as far as I can tell this section of code is useless with the way TCB is being used today. We also don't document this internal behavior, so it wasn't for user use.

It makes sense to drop it then to keep things as clean as possible and avoid confusion.

lucas-akira: > But then, this really does mean that this if "DOCKER_HOST" in os.environ: section in the original code is not being used at all, since this is based on DOCKER_HOST in the TCB container environment, which is not set. So really as far as I can tell this section of code is useless with the way TCB is being used today.

Looking at the commit message where this `if` statement and its respective comment were added, it looks like this part was really intended to be used for our GitLab CI tests: the logic there apparently expects `DOCKER_HOST` inside the TCB container environment to be set with the same value the as the environment where the TCB container was executed, which, as you pointed out, isn't the case.

> A legitimate use case that I can imagine for passing DOCKER_HOST to TCB is if the Docker daemon was running on a non-standard port.

That is the only use-case that I can imagine this being useful. Not sure how common this Docker configuration would be though.

> Off the top of my head you'd probably need to set DOCKER_HOST for the TCB container. That way it gets executed using the right port. But then you'd also need to set DOCKER_HOST in the TCB alias to pass it into the TCB container as well. That way the dind container from bundle would get executed using the correct port as well. In theory this would already work today I believe. Since self.host_client = docker.from_env() would grab the DOCKER_HOST set in the TCB alias. This would then be used to execute the dind container at the designated TCP address.

The `bundle` command has a `--dind-env` option, so I think that it should be used with the same value as the `DOCKER_HOST` in the TCB alias, right?

jsrc27: Okay as discussed I removed the logic for handling `DOCKER_HOST`.

Users can still execute TCB itself on a different host by prepending `DOCKER_HOST` to the TCB command, or setting in the host environment. But for the aforementioned reasons, it only makes sense for the dind container used by `bundle` to be executed on the same host as the TCB container. Which is automatically done as is.

rborn-tx (approved):
LGTM

lucas-akira (approved):

PR #130, opened 2026-02-19
bundle: Allow container bundling on overlayfs

rborn-tx:
Allow the bundle command to run if the storage driver of the base system (where DinD in running) is "overlayfs" (in addition to the previously supported "overlay2" driver).


lucas-akira (approved):
LGTM

jsrc27 (approved):

PR #133, opened 2026-03-06
bundle: Allow container bundling on overlayfs

rborn-tx:
Ported from trixie:

Allow the bundle command to run if the storage driver of the base system (where DinD in running) is "overlayfs" (in addition to the previously supported "overlay2" driver).

Related-to: TCB-841


lucas-akira (approved):
LGTM

PR #134, opened 2026-03-12
Expand raw disk images when combining large container images

lucas-akira:
Automatically expand the root filesystem of the output disk image if the required size of the uncompressed contents of the container bundle exceeds or almost exceeds the available root space. 

In order to expand the filesystem, its underlying partition and the disk image file itself need to be expanded first. These operations were implemented using the libguestfs `virt-resize` tool.

Also included are some code improvements related to raw disk image manipulation. See the commit messages for more details.

Related-to: TCB-840

### DIFF
--- a/tcbuilder/backend/combine.py
+++ b/tcbuilder/backend/combine.py
@@ -9,11 +9,12 @@ import re
 import datetime
 
 import fnmatch
+import subprocess
 
 from tezi.image import ImageConfig
 from tcbuilder.backend.common import \
     (set_output_ownership, check_licence_acceptance, run_with_loading_animation,
-     open_disk_image, DOCKER_BUNDLE_TARNAME)
+     open_disk_image, get_tar_compress_program_options, DOCKER_BUNDLE_TARNAME)
 from tcbuilder.errors import InvalidStateError, InvalidDataError, TorizonCoreBuilderError
 
 log = logging.getLogger("torizon." + __name__)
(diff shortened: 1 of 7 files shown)


tcbuilder/backend/combine.py
```diff
@@ -261,94 +300,188 @@ def combine_tezi_image(image_dir, bundle_dir, output_directory, tezi_props, forc
...
+        tmp_image = output_path + ".not_bundled"
+        log.debug("Copying '%s' -> '%s' for virt-resize.", output_path, tmp_image)
+        shutil.copyfile(output_path, tmp_image)
+        image_path = tmp_image
+
+    expand_disk_partition(image_path, root_partition, output_path)
+
+    if tmp_image and os.path.isfile(tmp_image):
+        log.debug("Deleting '%s'", tmp_image)
+        os.remove(tmp_image)
```
rborn-tx: How about putting the call to `expand_disk_partition()` inside a try-block and the image deletion in its `finally`?

rborn-tx (changes requested):
Good job! Nice code cleanup by using the context manager btw.

Overall looks good. I've added a minor improvement suggestion to the code though.

Regarding the commit messages, could you remove the empty line between the `Related-to` and your sign-off? This is a pattern we should follow with all repositories to allow the commit messages to be parsed as [trailers](https://git-scm.com/docs/git-interpret-trailers) by Git.

tcbuilder/backend/combine.py
lucas-akira: I agree, it makes sense to put the tmp_image removal  in a `finally` clause in case `virt-resize` fails for whatever reason. I'll apply the suggestion along with the commit message changes.

PR #141, opened 2026-04-16
.gitlab-ci.yml: switch to Automation-managed AWS runners

lucas-p-bernardes:


jsrc27 (approved):

lucas-p-bernardes:
While this MR was open, the Automation team made a change to the runner and tag infrastructure, so I updated the commits to align with that change.

lucas-akira (approved):
LGTM

rborn-tx:
@lucas-p-bernardes Would you mind to explain why it's okay to move to these tags containing only the architecture?

The `saas-linux-<size>-<arch>` are defined by Gitlab.com with hosted runners and checking the docs I see they still exist. This leads to the question: if we drop these tags, how would the appropriate runner be selected? Is that information somewhere else now?

lucas-p-bernardes:
> @lucas-p-bernardes Would you mind to explain why it's okay to move to these tags containing only the architecture?
> 
> The `saas-linux-<size>-<arch>` are defined by Gitlab.com with hosted runners and checking the docs I see they still exist.

@rborn-tx Recently, we ran into the issue of GitLab CI minutes being exhausted. Those `saas-linux-<size>-<arch>` runners do still exist, but the Automation team is moving away from using those CI minutes due to internal issues.

What they did was provision `amd64` and `arm64` runners on AWS and assign the `amd64` and `arm64` tags to those runners. They described  the available tags in this document: https://toradex.atlassian.net/wiki/spaces/AUT/pages/5086740482/GitLab+Runner+Tags

> if we drop these tags, how would the appropriate runner be selected? Is that information somewhere else now?

Addressing this issue more specifically,, what is being done here is changing the runners in use, moving away from GitLab-hosted runners to runners provisioned on AWS by the Automation team, because now these are the appropriate runners according to Automation.

lucas-p-bernardes:
@rborn-tx I changed the title and the commits to make it clear that this is a change to the runners being used, and not just to the tags.

rborn-tx:
> @rborn-tx I changed the title and the commits to make it clear that this is a change to the runners being used, and not just to the tags.

@lucas-p-bernardes Got it now. Thanks for the clarifications!

rborn-tx (approved):

PR #142, opened 2026-04-16
Add new --fleet option for images provision

jsrc27:
This adds a new option for `images provision` that allows users to provision their devices to fleets in Torizon Cloud with auto-provisioning. This adds the following:

* Standalone command support `images provision --fleet`
* Support for option in the `build` command
* Test cases for the new option

### DIFF
--- a/tcbuilder/backend/images.py
+++ b/tcbuilder/backend/images.py
@@ -15,6 +15,7 @@ import subprocess
 import sys
 import tempfile
 import urllib.request
+import uuid
 
 from zipfile import ZipFile
 from tempfile import TemporaryDirectory
@@ -415,7 +416,8 @@ def prov_check_provdata_presence(input_dir):
     return config.search_filelist(src=PROV_DATA_FILENAME) is not None
 
 
-def prov_gen_provdata_tarball(output_dir, shared_data, online_data, hibernated):
+def prov_gen_provdata_tarball(output_dir, shared_data, online_data, *,
+                              hibernated=False, fleets=None):
     """Generate tarball containing all provisioning data
 
     The tarball will be stored into the output directory; then it should be
@@ -459,6 +461,12 @@ def prov_gen_provdata_tarball(output_dir, shared_data, online_data, hibernated):
                         online_data_obj['hibernated'] = True
                         online_data_json = json.dumps(online_data_obj).encode("utf-8")
 
+                    # Add fleet UUIDs if present
+                    if fleets and isinstance(online_data_obj, dict):
+                        log.info("Adding fleet UUID(s).")
+                        online_data_obj['fleetids'] = fleets
+                        online_data_json = json.dumps(online_data_obj).encode("utf-8")
+
                 except (binascii.Error, json.decoder.JSONDecodeError) as exc:
                     raise TorizonCoreBuilderError(
                         "Failure decoding online data: aborting.") from exc
@@ -489,7 +497,7 @@ def prov_add_provdata_tarball(output_dir):
 
 
 def provision(input_dir, output_dir, shared_data, online_data, *,
-              hibernated=False, force=False):
+              hibernated=False, fleets=None, force=False):
     """Generate TEZI image with added provisioning data
 
     :param input_dir: Path of directory containing input image.
@@ -498,6 +506,7 @@ def provision(input_dir, output_dir, shared_data, online_data, *,
                         offline and online cases) provisioning data.
     :param online_data: Base-64 string containing online provisioning data.
     :param hibernated: Boolean that indicates to whether or not provision in hibernated mode
+    :param fleets: List of fleet UUID(s) to provision to
     :param force: Boolean indicating whether to remove output directory if it
                   already exists.
     """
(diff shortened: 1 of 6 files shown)


jsrc27:
I've marked this as a draft as this relies on this PR in the OS: https://github.com/torizon/meta-toradex-torizon/pull/489

Which is pending merging due to the feature freeze. Otherwise feel free to start reviewing this.

tcbuilder/backend/images.py
```diff
@@ -519,19 +527,40 @@ def provision(input_dir, output_dir, shared_data, online_data, *,
...
             # Torizon OS 6.7 or below doesn't support hibernated auto-provisioning
             # without changes to the auto-provisioning script.
             log.warning("Warning: Hibernated auto-provisioning is not supported on Torizon "
                         "OS 6.7 or older. Proceeding anyway.")
 
+    if online_data and fleets:
+        if (img_major == 7 and img_minor <= 6) or img_major < 7:
+            # Torizon OS 7.6 or below doesn't support fleet auto-provisioning
+            log.warning("Warning: Fleet auto-provisioning is not supported on Torizon "
+                        "OS 7.6 or older. Proceeding anyway.")
```
lucas-akira: The `img_major` and `img_minor` `if` checks here will raise a `TypeError` Exception if any of these variables are `None`, which is possible if `get_tezi_image_version()` can't determine a number from the version string.

I think the `if (online_data and hibernated) or (online_data and fleets)` check can be removed, and the hibernated/fleets version checks can be `elif` blocks, with the case where the version couldn't be determined be their corresponding `if` block:

```Python
    # Check unpacked version of Torizon OS
    img_major, img_minor, _ = get_tezi_image_version(input_dir)
    
    if img_major is None or img_minor is None:
        log.warning("Warning: Unable to determine image version in input directory. "
                    "Proceeding anyway.")
    elif online_data and hibernated:
        if (img_major == 6 and img_minor <= 7) or img_major < 6:
        ...
    elif online_data and fleets:
        if (img_major == 7 and img_minor <= 6) or img_major < 7:
        ...
```

jsrc27: Nice catch. I adjusted this in V2. Though I had to combine the inner ifs with the elifs like so:

```
    if img_major is None or img_minor is None:
        log.warning("Warning: Unable to determine image version in input directory. "
                    "Proceeding anyway.")
    elif ((online_data and hibernated) and
         ((img_major == 6 and img_minor <= 7) or img_major < 6)):
        ...
    elif ((online_data and fleets) and
         ((img_major == 7 and img_minor <= 6) or img_major < 7)):
        ...
```

In your suggestion if both fleets and hibernated were provided it would go into the elif block for hibernated but never do the check for fleets then. So I just made a single, slightly more complex check to cover this.

lucas-akira: > In your suggestion if both fleets and hibernated were provided it would go into the elif block for hibernated but never do the check for fleets then. So I just made a single, slightly more complex check to cover this.

Hmm you're right. But the new change would only show up the hibernated warning for versions 6.7.0 or lower, even if both hibernated and fleets options are enabled.

Looking back at what I've suggested before, the elifs on these two options don't make sense, as there are cases where both messages should be displayed.

How about this:

```Python
    # Check unpacked version of Torizon OS
    img_major, img_minor, _ = get_tezi_image_version(input_dir)
    
    if img_major is None or img_minor is None:
        log.warning("Warning: Unable to determine image version in input directory. "
                    "Proceeding anyway.")
    elif online_data:
        if hibernate and (img_major < 6 or (img_major == 6 and img_minor <= 7)):
            log.warning("Warning: Hibernated auto-provisioning is not supported...")
        if fleets and (img_major < 7 or (img_major == 7 and img_minor <= 6)):
            log.warning("Warning: Fleet auto-provisioning is not supported...")
```

That way the warnings don't exclude each other.

jsrc27: Yeah let's go with that approach. Implemented in the next revision just now.

jsrc27:
The related PR on the OS side has been merged, so moving this out of draft.

tcbuilder/backend/images.py
```diff
@@ -489,7 +496,7 @@ def prov_add_provdata_tarball(output_dir):
 
 
 def provision(input_dir, output_dir, shared_data, online_data, *,
-              hibernated=False, force=False):
+              hibernated=False, fleets, force=False):
```
rborn-tx: Should we set a default here? `fleets=None`

tcbuilder/backend/images.py
```diff
@@ -415,7 +416,7 @@ def prov_check_provdata_presence(input_dir):
     return config.search_filelist(src=PROV_DATA_FILENAME) is not None
 
 
-def prov_gen_provdata_tarball(output_dir, shared_data, online_data, hibernated):
+def prov_gen_provdata_tarball(output_dir, shared_data, online_data, hibernated, fleets):
```
rborn-tx: With the increasing number of parameters, I'd suggest changing the function definition to make `hibernated` and `fleets` keyword-only parameters just like `provision()`:

```python
def prov_gen_provdata_tarball(output_dir, shared_data, online_data, *, hibernated=False, fleets=None):
```

tcbuilder/backend/images.py
```diff
@@ -519,18 +527,35 @@ def provision(input_dir, output_dir, shared_data, online_data, *,
...
+            for uuids in fleets:
+                try:
+                    # Torizon Cloud uses version 4 UUIDs
+                    uuid.UUID(uuids, version=4)
+                except ValueError:
+                    bad_uuids.append(uuids)
+
+            if bad_uuids:
+                raise InvalidArgumentError(
+                    f"The following UUID(s) do not appear to be a UUID: {bad_uuids}")
```
rborn-tx: How about moving this code into a small `_validate_fleets(fleets)`?

rborn-tx (commented):
Overall, looks good to me. I've added a couple of suggestions mainly for code readability.

tcbuilder/backend/images.py
jsrc27: Sounds sensible. Done in the next revision.

tcbuilder/backend/images.py
jsrc27: Sure thing. Done in the next revision.

tcbuilder/backend/images.py
jsrc27: Yeah that's fair. Done in the next revision.

tcbuilder/backend/images.py
```diff
@@ -519,18 +528,25 @@ def provision(input_dir, output_dir, shared_data, online_data, *,
...
             # Torizon OS 6.7 or below doesn't support hibernated auto-provisioning
             # without changes to the auto-provisioning script.
             log.warning("Warning: Hibernated auto-provisioning is not supported on Torizon "
                         "OS 6.7 or older. Proceeding anyway.")
+        if fleets and ((img_major == 7 and img_minor <= 6) or img_major < 7):
+            # Torizon OS 7.6 or below doesn't support fleet auto-provisioning
+            log.warning("Warning: Fleet auto-provisioning is not supported on Torizon "
+                        "OS 7.6 or older. Proceeding anyway.")
+
+            _validate_fleets(fleets)
```
lucas-akira: Looking again at this part, if the fleet validation is called here it would only be done for versions up to (and including) 7.6.0, excluding any later versions. One way to prevent this is to make a nested `if` for the version check:

```Python
        if fleets:
            if (img_major == 7 and img_minor <= 6) or img_major < 7:
                # Torizon OS 7.6 or below doesn't support fleet auto-provisioning
                log.warning("Warning: Fleet auto-provisioning is not supported on Torizon "
                            "OS 7.6 or older. Proceeding anyway.")
            _validate_fleets(fleets)
```

jsrc27: Nice catch, you're right this check should be run if fleets are defined regardless of version. I must have missed this since while I was working on this 7.7.0 did not exist yet.

Fixed this in the next revision now.

rborn-tx (approved):

lucas-akira (approved):

PR #146, opened 2026-06-02
Add completion script to TorizonCore Builder container image

rborn-tx:
This is a first step towards having a completion script that is always in sync with the version of TorizonCore Builder. Basically this copies the completion script from https://github.com/toradex/tcb-env-setup and stashes it into the container image. A newer version of the setup script will then source it from the image when available.

The completion script here has been modified to align it with the CLI changes/additions in TorizonCore Builder since December 2024 (when the setup script was last updated).


lucas-akira (approved):
LGTM

jsrc27 (approved):

PR #149, opened 2026-06-15
backend: Suppress progress bar output when stdout is not a TTY

rborn-tx:
When testing the new `tcb-env-setup` script in CI, I found that TCB was producing lots of output when downloading the toolchain. The reason was the output of a progress bar that send out characters to stdout with each status update from the `urlretrieve` function.

This PR solves the situation by creating a context manager to handle the progress information and using it in all invocations of `urlretrieve`. The manager basically only displays progress when the stdout is connected to a terminal.

Please see the commit message for further details.

### DIFF
--- a/tcbuilder/backend/build.py
+++ b/tcbuilder/backend/build.py
@@ -6,7 +6,6 @@ import os
 import copy
 import logging
 import re
-import sys
 import shutil
 import tempfile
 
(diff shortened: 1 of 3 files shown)


lucas-akira (approved):
LGTM

jsrc27 (approved):

PR #150, opened 2026-06-16
images provision, build: Add support for raw disk images

lucas-akira:
Related-to: TCB-866

### DIFF
--- a/tcbuilder/backend/images.py
+++ b/tcbuilder/backend/images.py
@@ -25,7 +25,8 @@ import fabric
 from tcbuilder.backend.common import (get_rootfs_tarball, get_tar_compress_program_options,
                                       set_output_ownership, run_with_loading_animation,
                                       get_tezi_image_version, open_disk_image, RAW_PROP_TO_ARGNAME,
-                                      DEFAULT_RAW_ROOTFS_LABEL, SECBOOT_ARTIFACTS_DIR)
+                                      DEFAULT_RAW_ROOTFS_LABEL, SECBOOT_ARTIFACTS_DIR,
+                                      TAR_EXT_TO_PROGRAM, OSTREE_SOTA_DIR_PATH)
 from tcbuilder.backend import ostree
 from tcbuilder.backend.secboot import DEFAULT_TCB_SIGNING_FILES_TARNAME, BOOTLOADER_CONTAINER_NAME
 from tcbuilder.errors import (TorizonCoreBuilderError, InvalidArgumentError, InvalidStateError)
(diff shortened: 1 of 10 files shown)


tcbuilder/backend/images.py
```diff
@@ -408,14 +409,86 @@ def import_local_image(image_dir_or_file, tezi_dir, src_sysroot_dir, src_ostree_
...
+        return False
+
+    find_out = gfs.find("/ostree/deploy/torizon/var/sota/import/")
+
+    # Check contents of import directory
+    if "director" not in find_out or "repo" not in find_out:
+        return False
+
+    if "director/root.json" not in find_out or "repo/root.json" not in find_out:
+        return False
```
jsrc27: I assume there wasn't a more elegant way to check? Rather than checking each directory level.

tcbuilder/backend/images.py
```diff
@@ -547,45 +589,146 @@ def provision(input_dir, output_dir, shared_data, online_data, *,
...
+            shared_data=prov_data['shared'],
+            online_data=prov_data['online'],
+            hibernated=hibernated,
+            fleets=fleets)
+        ext = os.path.splitext(PROV_DATA_FILENAME)[1]
+        log.info("Copying provisioning data to '%s'.", rootfs_label)
+        gfs.mkdir_p("/ostree/deploy/torizon/var/sota/")
+        gfs.tar_in(os.path.join(tmpdir, PROV_DATA_FILENAME),
+                   "/ostree/deploy/torizon/var/sota/",
+                   compress=TAR_EXT_TO_PROGRAM[ext])
```
jsrc27: Could be better to have this `/ostree/deploy/torizon/var/sota` path as a variable. I think we use this path in quite a couple of spots throughout the codebase, even before this PR.

jsrc27 (commented):
Looks overall good to me. Just left some minor non-critical comments.

tcbuilder/backend/images.py
lucas-akira: I agree, I'll create a new constant to this path in `backend/common.py` since it's used in the combine command as well.

tcbuilder/backend/images.py
lucas-akira: Good point, I looked at the guestfs documentation and found out that there are functions that check the existence of a directory or a file in a given mounted path. Thanks, I'll replace the current check with those.

lucas-akira: OK, I've updated `prov_check_mounted_disk_data_presence()` with a new logic using the guestfs `is_file()` function.

jsrc27 (approved):
LGTM

tcbuilder/backend/images.py
```diff
@@ -408,14 +409,75 @@ def import_local_image(image_dir_or_file, tezi_dir, src_sysroot_dir, src_ostree_
             log.warning("Warning: Unsupported image version detected. Proceed at your own risk.")
 
 
-def prov_check_provdata_presence(input_dir):
+def prov_check_tezi_data_presence(input_dir):
```
rborn-tx: This new name looks a bit odd to me.

How about `prov_check_provdata_presence_in_tezi` (or even `check_provdata_presence_in_tezi`)?

tcbuilder/backend/images.py
```diff
@@ -408,14 +409,75 @@ def import_local_image(image_dir_or_file, tezi_dir, src_sysroot_dir, src_ostree_
...
     """Determine if input TEZI image already has provisioning data"""
 
     config_fname = os.path.join(input_dir, DEFAULT_IMAGE_JSON_FILENAME)
     config = ImageConfig(config_fname)
     return config.search_filelist(src=PROV_DATA_FILENAME) is not None
 
 
+def prov_check_mounted_disk_data_presence(gfs):
```
rborn-tx: Again related to naming. Most of the code dealing with disk images so far used the expression "raw image" but here we are starting to use "disk image". Is there a reason for the new naming? If not, I'd keep using the former for consistency (here and in all other functions being introduced).

E.g. `prov_check_provdata_presence_in_raw_image` (or even `check_provdata_presence_in_raw_image`)

tcbuilder/backend/images.py
```diff
@@ -408,14 +409,75 @@ def import_local_image(image_dir_or_file, tezi_dir, src_sysroot_dir, src_ostree_
...
+    sota_import_dir = OSTREE_SOTA_DIR_PATH + "/import/"
+    import_dirs = ["director/", "repo/"]
+    for import_dir in import_dirs:
+        if not gfs.is_file(sota_import_dir + import_dir + "root.json"):
+            return False
+
+    return True
+
+
+def prov_check_disk_image(disk_img, rootfs_label):
```
rborn-tx: What this function does seems very specific to me, as if it was just a helper for implementing `prov_check_input()`. Do you foresee it being used in other modules? If not, I'd suggest keeping it local to this module and renaming it to something like `_prov_get_image_version_raw_image`.

tcbuilder/cli/tcbuild.template.yaml
```diff
@@ -201,6 +216,9 @@ output:
       # registry: hub.docker.com
       # ca-certificate: cacert.pem
       # keep-double-dollar-sign: false
+      # dind-params:
+        # - '--param1'
+        # - '--param2'
```
rborn-tx, line 221: Since you've added this missing property here, I decided to check and noticed we are also missing `dind-env` 😩.

tests/integration/testcases/wic/images.bats
rborn-tx, line 1: Nice additions to the tests!

tcbuilder/backend/tcbuild.schema.yaml
rborn-tx, line 1: Nice work reusing `bundle` and `provisioning` definitions for raw and tezi images!

rborn-tx (commented):
Added a couple of suggestions and comments. Overall the implementation looks good to me.

My main concern is with the general approach we are following where we have different implementations for raw and tezi images inside TCB which is increasing maintenance effort more and more over time.

I wish we unify the two at some point...

tcbuilder/cli/tcbuild.template.yaml
lucas-akira: Nice catch, I'll add it as well then.

tcbuilder/backend/images.py
lucas-akira: I started to use "disk image" because I think it's a bit clearer than the more generic term "raw image".

In retrospect I should've probably used "disk image" from the beginning, but since "raw image" is already used in various places throughout the code, I guess I can keep using it for consistency as you said.

I originally wanted to keep the `prov` prefix while avoiding duplicating the `prov` term and also keeping the function name short (or not very long). That being said, from your two options I would choose to use `check_provdata_presence_in_raw_image`.

rborn-tx: I'd lean more towards consistency than to becoming clearer. If you believe "disk image" is clearer, you could also make a separate commit (maybe at the end of this work or even on a separate PR) changing the names everywhere.

lucas-akira: Right, I'll keep using the "raw image" term for now, as it is clear enough for everyone currently involved in the code. Changing the term everywhere could be done later on, I don't think it's something that needs to be done right now.

tcbuilder/backend/images.py
lucas-akira: Yeah I don't see it being used outside the module. Thanks, I'll rename it then.

rborn-tx (approved):

PR #151, opened 2026-06-17
cli: dto: mark dto deploy command as deprecated

GabrielSSAlmeida:
The dto deploy command is now marked as deprecated.

### Changes:
- Updated help and description messages to indicate deprecation and point users to the build command
- Added a runtime log.warning() when the command is invoked, ensuring the deprecation notice is visible even if the user doesn't read the help

Related-to: TCB-389

### DIFF
--- a/tcbuilder/cli/dto.py
+++ b/tcbuilder/cli/dto.py
@@ -24,6 +24,12 @@ log = logging.getLogger("torizon." + __name__)
 
 MAX_DTBO_FILE_SIZE = 1*1024*1024
 
+BUILD_COMMAND_URL = (
+    "https://developer.toradex.com/"
+    "torizon/os-customization/torizoncore-builder-tool-commands-manual"
+    "#the-build-command"
+)
+
 
 def _test_apply_overlay_nonfit(*, dtob_path, dtob_name, dtb_name, overlay_names):
     """Try to apply an overlay on top of a device-tree (non-FIT case).
@@ -697,6 +703,9 @@ def do_dto_deploy(args):
     Run just one command to deploy an overlay in the device, so it is easier
     and less error-prone to the user.
     """
+
+    log.warning("\n\n[DEPRECATED] Use the 'build' command instead.\n"
+                f"See: {BUILD_COMMAND_URL}\n\n")
 
     # Download TEZI image and checkout Device Tree files.
     args.remove_storage = True
--- a/tcbuilder/cli/dto.py
+++ b/tcbuilder/cli/dto.py
@@ -811,8 +811,15 @@ def init_parser(subparsers):
     # dto deploy
     subparser = subparsers.add_parser(
         "deploy",
-        description="Deploy a device tree overlay in the device",
-        help="Deploy a device tree overlay in the device",
+        description=(
+            "[DEPRECATED] Deploy a device tree overlay in the device. "
+            "This command is deprecated. Use the 'build' command instead. "
+            f"For more information: {BUILD_COMMAND_URL}"
+        ),
+        help=(
+            "[DEPRECATED] Deploy a device tree overlay in the device. "
+            "Use the 'build' command instead."
+        ),
         allow_abbrev=False)
     subparser.add_argument(
         "--remote-host",


tcbuilder/cli/dto.py
```diff
@@ -802,8 +809,13 @@ def init_parser(subparsers):
     # dto deploy
     subparser = subparsers.add_parser(
         "deploy",
-        description="Deploy a device tree overlay in the device",
-        help="Deploy a device tree overlay in the device",
+        description=(
+        "[DEPRECATED] Deploy a device tree overlay in the device. "
+        "This command is deprecated. Use the 'build' command instead. "
+        "For more information: "
+        "https://developer.toradex.com/torizon/os-customization/torizoncore-builder-tool-commands-manual#the-build-command"
```
jsrc27: Shouldn't we use the `build_command_url` variable here as well?

tcbuilder/cli/dto.py
```diff
@@ -698,6 +698,13 @@ def do_dto_deploy(args):
     and less error-prone to the user.
     """
 
+    build_command_url="https://developer.toradex.com/torizon/os-customization/"\
+        "torizoncore-builder-tool-commands-manual#the-build-command"
+    
+    log.warning("\n\n[DEPRECATED] Use the 'build' command instead.\n"
+        f"See: {build_command_url}\n\n"
```
jsrc27: After printing the warning here, should we just stop execution of this command as well?

Opinions?

jsrc27 (commented):
Left some minor comments.

tcbuilder/cli/dto.py
GabrielSSAlmeida: Yeah, I think so. This change will resolve the pylint issue as well. I'll change it.

tcbuilder/cli/dto.py
lucas-akira: When we deprecated the `dt checkout` command for Torizon OS 6 and newer we stopped the command execution by raising a `TorizonCoreBuilderError` exception: 

https://github.com/torizon/torizoncore-builder/blob/ae6e0cd1803c126c3cc0ac7f1a8bb10f36e54dc4/tcbuilder/backend/common.py#L509

Since `dto deploy` eventually calls `do_dt_checkout()`, I think we should also raise a warning here and stop the command entirely.

tcbuilder/cli/dto.py
```diff
@@ -802,8 +808,13 @@ def init_parser(subparsers):
...
         "deploy",
-        description="Deploy a device tree overlay in the device",
-        help="Deploy a device tree overlay in the device",
+        description=(
+        "[DEPRECATED] Deploy a device tree overlay in the device. "
+        "This command is deprecated. Use the 'build' command instead. "
+        f"For more information: {BUILD_COMMAND_URL}"
+        ),
+        help="[DEPRECATED] Deploy a device tree overlay in the device. " \
+        "Use the 'build' command instead.",
```
rborn-tx: Suggestion: follow our common indentation. See good examples here (close to the line linked below):

https://github.com/torizon/torizoncore-builder/blob/ae6e0cd1803c126c3cc0ac7f1a8bb10f36e54dc4/tcbuilder/cli/secboot.py#L236

tcbuilder/cli/dto.py
rborn-tx: Since the goal was deprecating the command, for me it seems we should just show the warning.

In practice, the command would likely fail anyways (right?) because users would probably be using TOS version 6 or newer where the checkout would fail. Unless they have the `device-trees` directory present in the working directory, in which case, I don't see a reason for preemptively stopping this (rather awkward) command.

Of course I'd prefer to stop the command or even fully drop it, but that doesn't seem aligned with my understanding of "deprecation".

tcbuilder/cli/dto.py
```diff
@@ -698,6 +700,10 @@ def do_dto_deploy(args):
     and less error-prone to the user.
     """
 
+    log.warning("\n\n[DEPRECATED] Use the 'build' command instead.\n"
```
rborn-tx, line 707: @GabrielSSAlmeida Please look for other places in the code to see how we usually indent the strings in the call to log functions taking multiple lines.

Hint: The starting quotes in consecutive lines are aligned and the final parenthesis is in the same line as the final quote character.

tcbuilder/cli/dto.py
```diff
@@ -24,6 +24,8 @@
 
 MAX_DTBO_FILE_SIZE = 1*1024*1024
 
+BUILD_COMMAND_URL="https://developer.toradex.com/torizon/os-customization/"\
+        "torizoncore-builder-tool-commands-manual#the-build-command"
```
rborn-tx: - We generally avoid continuing strings with backslashes.
- There should be a space character surrounding the equals sign (in most places but not all (e.g. not in function calls)).

Ideas (semantically breaking the URL):

```
BUILD_COMMAND_URL = \
    ("https://developer.toradex.com/"
     "torizon/os-customization/torizoncore-builder-tool-commands-manual"
      "#the-build-command")

BUILD_COMMAND_URL = (
    "https://developer.toradex.com/"
    "torizon/os-customization/torizoncore-builder-tool-commands-manual"
    "#the-build-command"
)
```

rborn-tx (changes requested):

tcbuilder/cli/dto.py
lucas-akira: > Since the goal was deprecating the command, for me it seems we should just show the warning.

The only concern I have with keeping the execution is that the `dt checkout` message showing up in the `dto deploy` command could potentially confuse users.

But if there's an explicit deprecation warning beforehand (as this PR does) I guess that would work too.

tcbuilder/cli/dto.py
GabrielSSAlmeida: Ok, @rborn-tx. Huge thanks for all your support. I will update all the points you commented on.

tcbuilder/cli/dto.py
GabrielSSAlmeida: Hey guys, I understand all the points and trade-offs involved, thank you all for the explanation. So, which path should I take? Do we have a final resolution?

lucas-akira: You can keep the program running and just log a warning (as you already did), it should be clear enough for the user.

tcbuilder/cli/dto.py
GabrielSSAlmeida: @rborn-tx, one question about these changes. For the TCB, do you prefer to rewrite the git history (it would result in a cleaner commit history) or just add new commits that address the requested changes? Let me know the best approach before I send the refactor.

tcbuilder/cli/dto.py
GabrielSSAlmeida: Thanks! @lucas-akira

tcbuilder/cli/dto.py
rborn-tx: @GabrielSSAlmeida While the PR is under review, we (and I should say pretty much all projects) rewrite Git history until things are settled for the merge.

GabrielSSAlmeida: Thanks!

rborn-tx:
Hey @GabrielSSAlmeida,

Notice that the commit message title is still above the limit (see the `CONTRIBUTING.md` guide):

```
cli: dto: add warning message when it uses the deprecated dto deploy command
<============================== 72 ====================================>
```

Also, in this repository, we generally capitalize the first word after the scope. Both of your commit messages are using lowercase characters for that first word.

GabrielSSAlmeida:
> Hey @GabrielSSAlmeida,
> 
> Notice that the commit message title is still above the limit (see the `CONTRIBUTING.md` guide):
> 
> ```
> cli: dto: add warning message when it uses the deprecated dto deploy command
> <============================== 72 ====================================>
> ```
> 
> Also, in this repository, we generally capitalize the first word after the scope. Both of your commit messages are using lowercase characters for that first word.

@rborn-tx  I've updated it. Sorry for the inconvenience

jsrc27 (approved):

rborn-tx (approved):

GabrielSSAlmeida:
I just did a rebase in the Trixie branch

lucas-akira (approved):
LGTM

PR #152, opened 2026-06-29
Add ARM64 host support

GabrielSSAlmeida:
## What changed
- Dockerfile: Added IMAGE_ARCH usage comment and fixed Go binary download in skopeo-builder to use the correct architecture.
- Kernel backend: Added ARM64 host support for building kernel modules. For ARM64→ARM64 builds, the arm-gnu-toolchain-14.2 variant is always used, disregarding Torizon Image major versions. For ARM64→ARM32, the 13.3 variant is used for Torizon 7 and the 11.3 variant is used for Torizon 6. In cases where the toolchain download from Toradex sources fails, a new code logic was added so the download falls back to official Arm sources.
- GitLab CI/CD: Extended pipeline to build, deploy, and publish the ARM64 variant.

## What changed from the previous version of this PR
- Correct toolchains versions for each (host_arch, target_arch, torizon_major_version) combination
- A new download fallback to Arm sources
- A new logic to decide the toolchain version using the CROSS_COMPILERS map, in order to make the code more declarative and avoid long if/elif statements.
- All linters and test jobs that just mirrored the amd64 jobs and didn't add much value were removed.

## Testing
All the new toolchains were tested in a verdin-imx8mp as a host machine (in a docker dind container). The tests focus mainly on Kernel Build functionality, using two kernel modules as examples: https://github.com/toradex/hello-mod and https://github.com/vi/virtual_touchscreen.

## Summary by CodeRabbit

* **New Features**
  * Added arm64 support across image building, testing, and deployment pipelines.
  * Multi-architecture images now include both amd64 and arm64 variants.
  * Added automated arm64 testing and deployment options for early-access and internal releases.
  * Kernel module builds now support aarch64 hosts and cross-compilation.
  * Improved toolchain downloads with a fallback source and clearer download errors.
  * Updated container tooling to support arm64 environments.
* **Bug Fixes**
  * Official arm64 deployments now fail clearly when unsuccessful.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -482,6 +482,36 @@ test-torizoncore:
     - TCB_REPORT=1 TCB_UNDER_CI=1 TCB_MACHINE=${BUILD_MACHINE} TCB_TESTCASE="$TCB_HOST_TESTS" ${T} ./run.sh
     - echo -e "\e[0Ksection_end:$(date +%s):test_section\r\e[0K"
 
+.test-base-arm64:
+  extends: .test-base
+  tags:
+    - arm64
+  variables:
+    IMAGE_NAME: torizoncore-builder-arm64
+    # List of all tests suites specific for arm64 test jobs
+    TCB_ARCH_SENSITIVE_TESTS: "kernel"
+  needs: ["build-arm64"]
+
+test-host-arm64-tc6-64bit-release:
+  extends: .test-base-arm64
+  variables:
+    YOCTO_BRANCH: "kirkstone-6.x.y"
+    TARGET_BUILD_TYPE: "release"
+    TCB_MACHINE: "apalis-imx8"
+  script:
+    - ./run_all.sh --machine $TCB_MACHINE --report --testcase "$TCB_ARCH_SENSITIVE_TESTS"
+      --tcb-custom-image "${TCB_IMAGE_UNDER_TEST}"
+
+test-host-arm64-tos7-64bit-nightly:
+  extends: .test-base-arm64
+  variables:
+    YOCTO_BRANCH: "scarthgap-7.x.y"
+    TARGET_BUILD_TYPE: "nightly"
+    TCB_MACHINE: "verdin-imx8mp"
+  script:
+    - ./run_all.sh --machine $TCB_MACHINE --report --testcase "$TCB_ARCH_SENSITIVE_TESTS"
+      --tcb-custom-image "${TCB_IMAGE_UNDER_TEST}"
+
 .aggregate-reports:
   stage: report
   image:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -169,11 +169,11 @@ build-amd64:
     IMAGE_NAME: torizoncore-builder-amd64
     DOCKERFILE_NAME: torizoncore-builder.Dockerfile
 
-# TODO: Consider removing arm64 completely.
 build-arm64:
   extends: .build-base
+  tags:
+    - arm64
   timeout: 5h
-  when: manual
   variables:
     IMAGE_ARCH: linux/arm64/v8
     IMAGE_NAME: torizoncore-builder-arm64
@@ -570,8 +570,14 @@ aggregate-reports-manual:
   script:
     - ${D} docker login -u "${CI_REGISTRY_USER}" -p "${CI_REGISTRY_PASSWORD}" "${CI_REGISTRY}"
     - ${D} docker pull "${CI_REGISTRY_IMAGE}/${IMAGE_NAME_AMD64}:${GITLAB_DOCKERREGISTRY_SUFFIX}"
-    - ${D} docker manifest create "${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}" "${CI_REGISTRY_IMAGE}/${IMAGE_NAME_AMD64}:${GITLAB_DOCKERREGISTRY_SUFFIX}"
-    - ${D} docker manifest annotate "${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}" "${CI_REGISTRY_IMAGE}/${IMAGE_NAME_AMD64}:${GITLAB_DOCKERREGISTRY_SUFFIX}" --os linux --arch amd64
+    - ${D} docker pull "${CI_REGISTRY_IMAGE}/${IMAGE_NAME_ARM64}:${GITLAB_DOCKERREGISTRY_SUFFIX}"
+    - ${D} docker manifest create "${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}"
+                                   "${CI_REGISTRY_IMAGE}/${IMAGE_NAME_AMD64}:${GITLAB_DOCKERREGISTRY_SUFFIX}"
+                                   "${CI_REGISTRY_IMAGE}/${IMAGE_NAME_ARM64}:${GITLAB_DOCKERREGISTRY_SUFFIX}"
+    - ${D} docker manifest annotate "${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}"
+                                    "${CI_REGISTRY_IMAGE}/${IMAGE_NAME_AMD64}:${GITLAB_DOCKERREGISTRY_SUFFIX}" --os linux --arch amd64
+    - ${D} docker manifest annotate "${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}"
+                                    "${CI_REGISTRY_IMAGE}/${IMAGE_NAME_ARM64}:${GITLAB_DOCKERREGISTRY_SUFFIX}" --os linux --arch arm64 --variant v8
     - ${D} docker manifest inspect -v "${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}"
     - ${D} docker manifest push "${CI_REGISTRY_IMAGE}/${IMAGE_NAME}:${GITLAB_DOCKERREGISTRY_SUFFIX_LATEST}"
 
@@ -582,6 +588,7 @@ aggregate-reports-manual:
   variables:
     IMAGE_NAME: torizoncore-builder
     IMAGE_NAME_AMD64: torizoncore-builder-amd64
+    IMAGE_NAME_ARM64: torizoncore-builder-arm64
   stage: build-multiarch
 
 build-multiarch-auto:
@@ -599,7 +606,7 @@ build-multiarch-auto:
 
 build-multiarch-manual:
   extends: .build-multiarch
-  needs: [build-amd64]
+  needs: [build-amd64, build-arm64]
   rules:
     # Following condition is needed to avoid the execution of merge-request pipelines.
     - if: '$CI_PIPELINE_SOURCE == "merge_request_event"'
@@ -657,8 +664,13 @@ build-multiarch-manual:
         done
     - for tag in ${DOCKER_TAGS} ${DOCKER_TAGS_CHECK};
         do
-          ${D} docker manifest create "${DOCKER_HUB_REPOSITORY}/${IMAGE_NAME}:${tag}" "${DOCKER_HUB_REPOSITORY}/${IMAGE_NAME_AMD64}:${tag}";
-          ${D} docker manifest annotate "${DOCKER_HUB_REPOSITORY}/${IMAGE_NAME}:${tag}" "${DOCKER_HUB_REPOSITORY}/${IMAGE_NAME_AMD64}:${tag}" --os linux --arch amd64;
+          ${D} docker manifest create "${DOCKER_HUB_REPOSITORY}/${IMAGE_NAME}:${tag}"
+                                      "${DOCKER_HUB_REPOSITORY}/${IMAGE_NAME_AMD64}:${tag}"
+                                      "${DOCKER_HUB_REPOSITORY}/${IMAGE_NAME_ARM64}:${tag}";
+          ${D} docker manifest annotate "${DOCKER_HUB_REPOSITORY}/${IMAGE_NAME}:${tag}"
+                                        "${DOCKER_HUB_REPOSITORY}/${IMAGE_NAME_AMD64}:${tag}" --os linux --arch amd64;
+          ${D} docker manifest annotate "${DOCKER_HUB_REPOSITORY}/${IMAGE_NAME}:${tag}"
+                                        "${DOCKER_HUB_REPOSITORY}/${IMAGE_NAME_ARM64}:${tag}" --os linux --arch arm64 --variant v8;
           ${D} docker manifest inspect -v "${DOCKER_HUB_REPOSITORY}/${IMAGE_NAME}:${tag}";
           ${D} docker manifest push "${DOCKER_HUB_REPOSITORY}/${IMAGE_NAME}:${tag}";
         done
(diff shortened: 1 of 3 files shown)


tcbuilder/backend/kernel.py
```diff
@@ -24,6 +24,12 @@
     7: "arm-gnu-toolchain-13.3.rel1"
 }
 
+# Toolchain versions used when the host is aarch64 and the target is also aarch64.
+# These are built with crosstool-ng and have a different naming convention.
+IMAGE_MAJOR_TO_GCC_MAP_AARCH64_NATIVE = {
```
jsrc27: At the time of writing Torizon 6 is still supported. Should we provide a toolchain in this case as well? Or was it already decided that we only support Torizon 7?

tcbuilder/backend/kernel.py
```diff
@@ -24,6 +24,12 @@
...
+    7: "crosstool-ng-13.3"
```
jsrc27: Where is this toolchain coming from exactly? Is it part of the TCB container somehow? I don't see in the download sources with the other toolchains: https://sources.toradex.com/tcb/toolchains/

tcbuilder/backend/kernel.py
```diff
@@ -103,14 +119,23 @@ def _get_toolchain(image_major_version, linux_src):
...
         c_c = "arm-none-linux-gnueabihf-"
         toolchain = os.path.join(
-            toolchain_path, f"{version_gcc}-x86_64-arm-none-linux-gnueabihf/bin")
+            toolchain_path, f"{version_gcc}-{host}-arm-none-linux-gnueabihf/bin")
     elif arch == "arm64":
-        c_c = "aarch64-none-linux-gnu-"
-        toolchain = os.path.join(
-            toolchain_path, f"{version_gcc}-x86_64-aarch64-none-linux-gnu/bin")
+        if host == "aarch64":
+            # On aarch64 hosts, use a crosstool-ng-built native cross-compiler.
```
jsrc27: Small nitpick. It's a bit of a confusing, contradicting statement to say "native cross-compiler".

tcbuilder/backend/kernel.py
GabrielSSAlmeida: This is a point that I forgot to explain in my PR description. I haven't put the toolchain on https://sources.toradex.com/tcb/toolchains/ yet, because I didn't know if you guys would accept this method. So, just for my local test, I put the toolchain locally in the Docker's /storage. Do you want me to upload this toolchain before continuing this review?

tcbuilder/backend/kernel.py
GabrielSSAlmeida: Ok, I'll update to 'native toolchain'. Thanks!

tcbuilder/backend/kernel.py
jsrc27: As long as the toolchain reliably works I don't have any objections.

But I'll let @lucas-akira and @rborn-tx chime in here as well with their thoughts.

tcbuilder/backend/kernel.py
lucas-akira: > Or was it already decided that we only support Torizon 7?

As far as I know nothing was decided yet. I think we should have the arm64 host toolchains available for Torizon OS 6 as well given that it is currently supported.

tcbuilder/backend/kernel.py
```diff
@@ -239,10 +264,13 @@ def download_toolchain(toolchain, toolchain_path, version_gcc):
     """Download toolchain from online if it doesn't already exist"""
 
     url_prefix = "https://sources.toradex.com/tcb/toolchains/"
```
lucas-akira: It isn't the main purpose of this PR, but I think it would be nice to have URL fallback options when downloading the toolchains. 

I imagined something like this: try to download the toolchain from our servers, and if that's not possible for whatever reason e.g. outage, then try to download the toolchain from the official ARM GNU toolchain URL. If that fails, then raise an exception.

This can be implemented for currently supported images (Torizon OS 6 and 7).

.github/workflows/pr-tests.yml
```diff
@@ -87,8 +92,33 @@ jobs:
...
+        run: docker run --rm -v ${GITHUB_WORKSPACE}:/work -w /work tcbuilder-dev-arm64
+                  pylint --output-format=colorized
+                         --disable=missing-module-docstring
+                         --disable=missing-class-docstring
+                         --disable=missing-function-docstring
+                         --disable=fixme
+                         $(find tcbuilder/ -type f -name "*.py") *.py || echo "exit_code=$?" >> ${GITHUB_ENV}
+      - name: Pylint result
+        run: exit ${exit_code}
+
```
lucas-akira: Do we need to run the Python linter on an arm64 host? Its results should be independent from the host architecture, so I don't see the need for this job.

.gitlab-ci.yml
```diff
@@ -128,6 +128,22 @@ lint-python:
...
+    DEV_CACHE_TAG: tcbuilder-dev-latest
+
+lint-python-arm64:
+  extends: .lint-python-base
+  tags:
+    - arm64
+  variables:
+    ARCH_SUFFIX: arm64
+    IMAGE_ARCH: linux/arm64/v8
+    DEV_CACHE_TAG: tcbuilder-dev-arm64-latest
```
lucas-akira: Same point I made on the corresponding GitHub Actions entry: I don't think we need to run the linters on a different host architecture.

tcbuilder/backend/kernel.py
rborn-tx: IMO it's a bit concerning to publish and use a toolchain that was built on someone's machine with an unknown setup. Since the toolchain was built by Toradex, it means we'd be responsible for it.

Say a user complains about some setting used for building the toolchain. We would have trouble to generate a new version only changing that setting since we wouldn't know the original settings.

Given the above, if the only option is to build the toolchain ourselves, then I think we should set up a new CI project for building and possibly publishing a toolchain with a well-tracked setup.

This is of course not a small topic we should discuss.

Cc @griloHBG

tcbuilder/backend/kernel.py
```diff
@@ -103,14 +119,23 @@ def _get_toolchain(image_major_version, linux_src):
...
+            assert version_gcc, "Unable to determine native aarch64 GCC toolchain version"
+            c_c = "aarch64-unknown-linux-gnu-"
+            toolchain = os.path.join(
+                toolchain_path, f"{version_gcc}-{host}-aarch64-unknown-linux-gnu/bin")
+        else:
+            c_c = "aarch64-none-linux-gnu-"
+            toolchain = os.path.join(
+                toolchain_path, f"{version_gcc}-{host}-aarch64-none-linux-gnu/bin")
     else:
         assert False, "build_module: Unhandled architecture"
```
rborn-tx: Idea for consideration:

Now that we have two dimensions being used to select the toolchain, maybe we could refactor this code to be more declarative. Instead of having logic to define the cross-compiler prefix and toolchain names we could have a 2-dimensional variable in this form:

```python
CROSS_COMPILERS = {
    ("<host-arch>", "<target-arch>"): ("<c_c>", "<toolchain>")
}
```

where `<host-arch>` comes from `host` and `<target-arch>` comes from `arch` (I'd actually rename that variable to `target`.

I think this would be helpful for readability, but before settling on that let me ask @jsrc27 and @lucas-akira opinions.

.gitlab-ci.yml
```diff
@@ -482,6 +499,98 @@ test-torizoncore:
     - TCB_REPORT=1 TCB_UNDER_CI=1 TCB_MACHINE=${BUILD_MACHINE} TCB_TESTCASE="$TCB_HOST_TESTS" ${T} ./run.sh
     - echo -e "\e[0Ksection_end:$(date +%s):test_section\r\e[0K"
 
+# ---
+# ARM64 host test variants
+# Each job mirrors its amd64 counterpart but pulls torizoncore-builder-arm64.
+# ---
```
rborn-tx: Given that the majority of TCB's code is independent of the host architecture, I'd guess we are not adding much value by mirroring these various tests from amd64.

Maybe we could consider something more focused such as running the "host tests" for a single image and then only the `kernel build_module` command (and other commands that we know could be impacted by the host arch) with various image combinations? This should save energy, time and money 😉...

tcbuilder/backend/kernel.py
GabrielSSAlmeida: Just an extra piece of information regarding your concern: I used Claude Code to build a crosstool-ng setup using Docker to standardize this toolchain build. When I started this project, I wasn't expecting to improve or maintain it, but after some improvements, it could be a good starting point for the CI project mentioned, at least to see all the versions I used.

.gitlab-ci.yml
GabrielSSAlmeida: ok, I'll remove it.

.gitlab-ci.yml
GabrielSSAlmeida: Make sense. I'll update it.

tcbuilder/backend/kernel.py
griloHBG: I agree that if the custom-built toolchain is needed, we would need to somehow maintain it.

But what could we do to no have this burden? We can discuss in a meeting.

But one idea: whatif we use a newer toolchain version, which IIRC has both ARM64->ARM32 and ARM64->ARM64 (the missing one) available as pre-built?

Which issues could happen?

GabrielSSAlmeida: Hey @rborn-tx, one question about this `CROSS_COMPILERS`. I started to develop this and it resulted in something like this:

```python
ARM_CC_PREFIX = "arm-none-linux-gnueabihf-"
ARM64_CC_PREFIX = "aarch64-none-linux-gnu-"
# ("<host-arch>", "<target-arch>", "<image-major-version") : (<c_c>, <toolchain>)
CROSS_COMPILERS = {
    ("x86_64", "arm", 5): (ARM_CC_PREFIX, "gcc-arm-9.2-2019.12"),
    ("x86_64", "arm64", 5): (ARM64_CC_PREFIX, "gcc-arm-9.2-2019.12"),
    ("x86_64", "arm", 6): (ARM_CC_PREFIX, "arm-gnu-toolchain-11.3.rel1"),
    ("x86_64", "arm64", 6): (ARM64_CC_PREFIX, "arm-gnu-toolchain-11.3.rel1"),
    ("x86_64", "arm", 7): (ARM_CC_PREFIX, "arm-gnu-toolchain-13.3.rel1"),
    ("x86_64", "arm64", 7): (ARM64_CC_PREFIX, "arm-gnu-toolchain-13.3.rel1"),
    
    ("aarch64", "arm", 6): (ARM_CC_PREFIX, "arm-gnu-toolchain-11.3.rel1"),
    ("aarch64", "arm64", 6): (ARM64_CC_PREFIX, "arm-gnu-toolchain-14.2.rel1"),
    ("aarch64", "arm", 7): (ARM_CC_PREFIX, "arm-gnu-toolchain-13.3.rel1"),
    ("aarch64", "arm64", 7): (ARM64_CC_PREFIX, "arm-gnu-toolchain-14.2.rel1"),

}
``` 

One thing that was necessary to add is the Image Version number, because this version influences the toolchain choice when the host machine is aarch64.This solution seems very verbose and I would like to know if this is your intention. 

I also tried another format for this `CROSS_COMPILER` map and it results in something like this:

```python
TARGET_TRIPLES = {
    "arm": "arm-none-linux-gnueabihf",
    "arm64": "aarch64-none-linux-gnu",
}

CROSS_COMPILERS = {
    ("x86_64", "arm", 5): "gcc-arm-9.2-2019.12",
    ("x86_64", "arm64", 5): "gcc-arm-9.2-2019.12",
    ("x86_64", "arm", 6): "arm-gnu-toolchain-11.3.rel1",
    ("x86_64", "arm64", 6): "arm-gnu-toolchain-11.3.rel1",
    ("x86_64", "arm", 7): "arm-gnu-toolchain-13.3.rel1",
    ("x86_64", "arm64", 7): "arm-gnu-toolchain-13.3.rel1",
    
    ("aarch64", "arm", 6): "arm-gnu-toolchain-11.3.rel1",
    ("aarch64", "arm64", 6): "arm-gnu-toolchain-14.2.rel1",
    ("aarch64", "arm", 7): "arm-gnu-toolchain-13.3.rel1",
    ("aarch64", "arm64", 7): "arm-gnu-toolchain-14.2.rel1",

}
``` 

This solution leverages the `TARGET_TRIPLES` map to build the final toolchain path and the `c_c` prefix. We can cut out a whole if statement entirely:
```python
target_triple = TARGET_TRIPLES.get(target_arch)
c_c = f"{target_triple}-"
toolchain = os.path.join(
                   toolchain_path, f"{version_gcc}-{host_arch}-{target_arch}/bin")
``` 

Is that declarative enough for you? Do you think that remove the if statement a good idea? Let me know your thoughts.

rborn-tx: This last approach looks good to me.

tcbuilder/backend/kernel.py
```diff
@@ -44,6 +59,16 @@
 MOD_DIR_COPY_EXCLUDE_SET = {"dtb"}
 
 
+def _host_arch():
+    """Return the Arm Ltd. host-arch prefix for cross-compiler toolchain names."""
```
jsrc27: Not sure why the comments here specifies "Arm Ltd."

tcbuilder/backend/kernel.py
```diff
@@ -99,28 +124,26 @@ def _amend_makefile(linux_src):
...
-        c_c = "aarch64-none-linux-gnu-"
-        toolchain = os.path.join(
-            toolchain_path, f"{version_gcc}-x86_64-aarch64-none-linux-gnu/bin")
-    else:
-        assert False, "build_module: Unhandled architecture"
+    target_arch = _kernel_arch_from_source(linux_src)
+    host_arch = _host_arch()
+
+    version_gcc = CROSS_COMPILERS.get((host_arch, target_arch, image_major_version))
+    assert version_gcc, "Unable to determine the GCC toolchain version"
```
jsrc27: For debugging purposes we could print the `host_arch`, `target_arch` and `image_major_version` in case this does fail.

tcbuilder/backend/kernel.py
```diff
@@ -99,28 +124,26 @@ def _amend_makefile(linux_src):
...
+
+    target_triple = TARGET_TRIPLES.get(target_arch)
+    assert target_triple, "build_module: Unhandled architecture"
```
jsrc27: Similarly we could print the `target_arch` to help debug the failure. Though it would be unlikely to occur I guess given this comes from `CROSS_COMPILERS` which has it's own check.

jsrc27 (commented):
Just some minor comments on my side. Nothing major though, looks overall good.

tcbuilder/backend/kernel.py
GabrielSSAlmeida: Sorry, this comment was generated incorrectly by AI and I didn't notice it before. I'll correct it.

tcbuilder/backend/kernel.py
```diff
@@ -237,28 +260,61 @@ def autoload_module(module, kernel_changes_dir):
...
     log.info(f"Downloading toolchain from {url}.")
     log.info("Please wait this could take a while...")
-
     try:
         with download_progress() as reporthook:
-            urllib.request.urlretrieve(url, filename=tarball, reporthook=reporthook)
+            urllib.request.urlretrieve(url, filename=dest, reporthook=reporthook)
+        return True
     except Exception as exc:
+        log.debug("Download from %s failed: %s", url, exc)
```
lucas-akira, line 283: Since the Exception is being handled right after exiting this function, you can add a `# pylint: disable-next=broad-exception` right before the Exception line to remove the Pylint warning found in the GitHub Actions pipeline.

lucas-akira (commented):
I only have one comment about a linter warning. The new version of the code looks good to me.

tcbuilder/backend/kernel.py
GabrielSSAlmeida: I've updated this line with your tip. Thanks.

tcbuilder/backend/kernel.py
GabrielSSAlmeida: Updated.

tcbuilder/backend/kernel.py
GabrielSSAlmeida: Updated.

.gitlab-ci.yml
rborn-tx: @GabrielSSAlmeida This is ongoing, right?

GabrielSSAlmeida: Oops, totally forgot to add this to the new PR. Already fixed it!

jsrc27 (approved):

lucas-akira (approved):

.gitlab-ci.yml
rborn-tx: With arm64 becoming officially supported, it seems we should drop this `allow_failure: true` here?

rborn-tx:
@GabrielSSAlmeida Checking https://sources.toradex.com/tcb/toolchains/ I see we are not mirroring the aarch64 compilers from ARM yet. Do we have intentions to mirror them?

.gitlab-ci.yml
GabrielSSAlmeida: Makes sense, I'll correct this part.

GabrielSSAlmeida:
Hi @rborn-tx 
> @GabrielSSAlmeida Checking https://sources.toradex.com/tcb/toolchains/ I see we are not mirroring the aarch64 compilers from ARM yet. Do we have intentions to mirror them?

I talked to the automation team to mirror the toolchains in that link. However, they asked me to create a ticket with a due date to bring more organization on their side. To avoid rework on their end, I decided to wait until this PR was approved, or at least until we reached more maturity in the approach we'll follow. I think we've reached that maturity now, so I will create the ticket for the automation team with a due date of 1 week. Is this date ok for you?

rborn-tx:
> ... I think we've reached that maturity now, so I will create the ticket for the automation team with a due date of 1 week. Is this date ok for you?

@GabrielSSAlmeida Sure. Thanks!

griloHBG:
Hi @GabrielSSAlmeida !

Is this PR really blocked? I know that we are waiting on the toolchain availability, but couldn't we merge this already?

Or maybe something else more blocking is missing and I am not aware?

GabrielSSAlmeida:
> Hi @GabrielSSAlmeida !
> 
> Is this PR really blocked? I know that we are waiting on the toolchain availability, but couldn't we merge this already?
> 
> Or maybe something else more blocking is missing and I am not aware?

Hi, @griloHBG . I was waiting for the toolchain availability before asking for a new review. Maybe that was not the best approach :sweat_smile: , but the toolchain is available, tested and I'm requesting the review now. Thanks.

.gitlab-ci.yml
GabrielSSAlmeida: Updated.

GabrielSSAlmeida:
> > ... I think we've reached that maturity now, so I will create the ticket for the automation team with a due date of 1 week. Is this date ok for you?
> 
> @GabrielSSAlmeida Sure. Thanks!

The toolchains are now available at https://sources.toradex.com/tcb/toolchains/.

rborn-tx (approved):

rborn-tx:
@GabrielSSAlmeida We merged CI fixes yesterday. If you rebase this PR, tests should now pass.

PR #153, opened 2026-07-01
Add Device Tree Overlay support for U-Boot raw disk images, plus other features for Common Torizon

lucas-akira:
Add support to apply, remove and check applied device tree overlays for raw disk images using U-Boot , either using the `dto` command or using the corresponding `tcbuild.yaml` option for the `build` command.

Add support to download supported Common Torizon images in raw disk format from the Toradex servers with `images download` and with the input option of the `build` command.

Also implement input image (TEZI and raw disk) checksum verification for `images download` and `build`.

See the commit messages for more details.

Related-to: TCB-865


lucas-akira:
I see that I forgot to add the tests for the new features. I'll put this PR in draft until I add them.

lucas-akira:
OK, I've added the tests. It's ready for review

tcbuilder/backend/tcbuild.schema.yaml
```diff
@@ -5,8 +5,87 @@ properties:
...
+      raw-img-machines:
+        type: string
+        enum:
+          - am62lxx-evm
+          - am62pxx-evm
+          - am62xx-evm
+          - beagley-ai
+          - intel-corei7-64
+          - luna-sl1680
+          - sl1680
```
jsrc27, line 42: I'm kind of torn on this. On one hand it feels "weird" to list this here when trying to download the image for either of these machines will cause an error. So it's not really "supported" in that sense

But it's also a good way to guide/advertise the use of the conversion script, since they will see the log messages in this case.

Thoughts?

tcbuilder/cli/build.py
```diff
@@ -148,20 +160,55 @@ def handle_easy_installer_input(props, download_dir=None):
...
+
+        else:
+            # Parse remote which may contain integrity checking information.
+            remote_url, remote_fname, cksum = bb.parse_remote(props["remote"])
+            log.debug(f"Remote URL: {remote_url}, name: {remote_fname}, "
+                      f"expected sha256: {cksum}")
+
+        # Next call will download the file if necessary.
+        local_file, _ = common.fetch_remote(
+            remote_url, fname=remote_fname, cksum=cksum, download_dir=download_dir)
```
jsrc27, line 198: Do we need to set the output ownership on the download file here?

jsrc27 (commented):
Overall looks okay, no major issues. I did comment on some minor stuff though.

tcbuilder/cli/build.py
lucas-akira: Nice catch, I'll set the proper ownership for the downloaded file and for the download directory path if TCB needs to create it.

lucas-akira: OK, that should be fixed now.

tcbuilder/backend/tcbuild.schema.yaml
lucas-akira: I had the same question, but ultimately I decided to add these to the schema and throw an exception later about the SYNAIMG format in the error message. I thought that would be clearer than not adding the machines to the schema and users getting a generic schema error.

tcbuilder/backend/common.py
```diff
@@ -1038,3 +1086,246 @@ def open_disk_image(image_path, *, delete_on_error=False, readonly=False, loadin
...
+def sanitize_fname(fname, repl="_"):
+    """Replace disallowed characters in file names"""
+    return re.sub(r"[^\w\.\-\+]", repl, fname)
+
+
+# From https://stackoverflow.com/questions/37060344/
+# how-to-determine-the-filename-of-content-downloaded-with-http-in-python
+#
+def parse_disposition_header(header):
+    """Simplified parser of Content-Disposition header (RF6266)"""
```
rborn-tx: @lucas-akira Could we take the opportunity to fix this typo (present before your refactoring):

"RF6266" -> "RFC 6266"

tcbuilder/cli/build.py
```diff
@@ -95,28 +95,41 @@ def handle_input_section(props, **kwargs):
...
 
+    download_dir = None
+    if "image-download-policy" in props and props["image-download-policy"] == "use-cache":
+        if "image-download-dir" in props:
+            if not os.path.isdir(props["image-download-dir"]):
+                os.makedirs(props["image-download-dir"])
+                common.set_output_ownership(props["image-download-dir"], set_parents=True)
+            download_dir = props["image-download-dir"]
+        else:
+            download_dir = os.getcwd()
```
rborn-tx: I'm glad to see you're clearing an important TODO of adding support for caching input images.

Since this is a whole new feature, I think it's worth discussing its behavior.

If I get it correctly from this stretch of code, the downloads directory would be used only when the user specified the download-policy "use-cache". From this I understand that the download directory is seen as a synonym to the cache directory, right?

Some questions come to mind:

- If the user specified a download directory, shouldn't we always use it, regardless of a download-policy?
- What do we lose if we don't have a download-policy at all?

I raise these questions to see if we could simplify the UX even further.

Say we didn't have the policy I'd expect a behavior like this:

- If the user specified the directory, we would always store the file in that directory.
- If the file already exists and it matches the **user-specified** hash (when the `remote` property was used), then do not download the file.
- If the file already exists and the above condition was not true then check if the existing file hash matches the one on the server; if so, do not download it again.
- If none of the previous two conditions was true, download the file.

tcbuilder/backend/tcbuild.schema.yaml
rborn-tx: I'm leaning towards keeping the behavior proposed in the present PR where an error is thrown and the user learns that they need to manually download it. Despite being a bit odd, it seems unlikely to confuse users given that the error message is quite clear.

tcbuilder/cli/dt.py
```diff
@@ -30,6 +30,13 @@ def do_dt_status(_args):
     """Perform the 'dt status' command."""
 
     images_unpack_executed()
+    img_arch = get_arch_from_ostree()
```
rborn-tx: Have you considered some way to determine if U-Boot is being used as bootloader? If we had a reliable way to do so, we could make the conditions independent of the machine and dependent on the bootloader, which I believe is what matters here, right?

rborn-tx (commented):
Unfortunately I didn't have time to look into the tests, but the implementation looks great.

I've raised a few questions for discussion.

rborn-tx:
I forgot to mention in the review: the commit message title:

```
dto: Add support for U-Boot images in raw disk format
```

seems a bit odd to me (particularly the terms `U-Boot images`); how about any of these:

```
dto: Add support for raw disk images using U-Boot as bootloader
dto: Add support for raw disk images using U-Boot
dto: Add support for U-Boot-based images in raw disk format
```

tcbuilder/backend/common.py
lucas-akira: Sure, no problem. I'll update it.

tcbuilder/cli/dt.py
lucas-akira: I considered checking for U-Boot here, but I imagined that using the target architecture of the OS image would be a more generic way to filter out device tree/device tree overlay support, as x86_64 machines don't usually support them.

Thinking now that would probably be a valid check to make if DT/DTO was not supported at all for x86_64, which is not necessarily the case.

So it makes more sense to have a whitelist i.e. a list of supported bootloaders rather than trying to blacklist unsupported bootloaders, target architectures, etc., which is basically what you suggested.

The `deploy` command uses a backend/common.py function called `get_image_bootloader()` to determine the bootloader of the unpacked OS image. It's a function that I made more or less based on the method used by libostree to determine the bootloader (U-Boot and GRUB2). I'm thinking on using that here as well.

lucas-akira:
Sure, `dto: Add support for raw disk images using U-Boot` sounds better to me. I'll change it.

tcbuilder/cli/build.py
lucas-akira: > - If the user specified a download directory, shouldn't we always use it, regardless of a download-policy?
> - What do we lose if we don't have a download-policy at all?
>
> I raise these questions to see if we could simplify the UX even further.

Hmm right, that is more or less the approach I did for `images download`: the image file would only be locally saved if the user explicitly passes the new `--download-dir` argument.

I agree that just having `image-download-dir` would be simpler. Thanks, I'll apply your suggestion.

jsrc27 (approved):
@lucas-akira From my side I don't have any further comments. I think all of @rborn-tx comments have been addressed as well. If you're fine with it we can merge now since he's out for the next couple of weeks. Or do you want to wait?

lucas-akira:
@jsrc27 In the `dto` commands I still need to replace the target architecture check of the OS image with a bootloader check. After that and a review on your side I think we can merge it if everything is OK.

lucas-akira:
OK, I've replaced the arch check with the bootloader one and it's ready for review.

I also had to do a small change to the `find` command of `get_dtb_kernel_subdir()` in `backend/dt.py` because for some Common Torizon images for TI machines it was returning an unexpected `dtb` sysroot path (`boot/dtb` instead of the expected `usr/lib/modules/<kernel-version>/dtb` one).

PR #154, opened 2026-07-08
Support kernel build_module for arm32/64 raw disk images

jsrc27:
This allows the `kernel build_module` command to be used on WIC/raw disk images as long as the architecture of the image is either arm32 or arm64. Other architectures are still unsupported here.

After some testing it appears that we can just reuse the same arm32 and arm64 cross-compile toolchains that we have been using for Toradex devices so far. This was tested on a Luna SL1680. I was able to build and load the simple "hello world" kernel module that we use for tests.

Finally, this also adds the corresponding tests on the WIC side.

### DIFF
--- a/tests/integration/testcases/wic/kernel.bats
+++ b/tests/integration/testcases/wic/kernel.bats
@@ -1,0 +1,161 @@
+bats_load_library 'bats/bats-support/load.bash'
+bats_load_library 'bats/bats-assert/load.bash'
+bats_load_library 'bats/bats-file/load.bash'
+load '../lib/common.bash'
+
+setup_file() {
+    local UNSUPPORTED_MACHINES="intel-corei7-64 qemux86-64"
+    local ARCHIVE='/storage/ostree-archive/'
+    local DEFAULT_REF='base'
+    torizoncore-builder images --remove-storage unpack "${DEFAULT_WIC_IMAGE}"
+    IMG_MACHINE=$(torizoncore-builder-shell "ostree --repo=${ARCHIVE} show \
+                  --print-metadata-key='oe.machine' ${DEFAULT_REF}" | tr -d "'")
+
+    case "${UNSUPPORTED_MACHINES}" in
+        *${IMG_MACHINE}* )
+            UNSUPPORTED_MACHINE="1"
+            ;;
+        *)
+            UNSUPPORTED_MACHINE="0"
+            ;;
+    esac
+
+    export UNSUPPORTED_MACHINE
+    export IMG_MACHINE
+}
+
+setup() {
+  if [ "${UNSUPPORTED_MACHINE}" = "1" ]; then
+    skip "DT/DTO customization is not supported for ${IMG_MACHINE}"
+  fi
+}
+
+@test "kernel: run without parameters" {
+    run torizoncore-builder kernel
+    assert_failure 2
+    assert_output --partial "{build_module,set_custom_args,get_custom_args,clear_custom_args}"
+    assert_output --partial "error: the following arguments are required: cmd"
+}
+
+@test "kernel: check help output" {
+    run torizoncore-builder kernel --help
+    assert_success
+    assert_output --partial "usage: torizoncore-builder kernel [-h]"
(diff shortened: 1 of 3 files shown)


tests/integration/testcases/wic/kernel.bats
```diff
@@ -0,0 +1,159 @@
+bats_load_library 'bats/bats-support/load.bash'
+bats_load_library 'bats/bats-assert/load.bash'
+bats_load_library 'bats/bats-file/load.bash'
+load '../lib/common.bash'
+
+setup_file() {
+    local UNSUPPORTED_MACHINES="intel-corei7-64 qemux86-64"
+    torizoncore-builder images --remove-storage unpack "${DEFAULT_WIC_IMAGE}"
+    IMG_MACHINE=$(torizoncore-builder-shell "ostree --repo=${ARCHIVE} show \
+                  --print-metadata-key='oe.machine' ${DEFAULT_REF}" | tr -d "'")
```
lucas-akira, line 12: I think you forgot to declare `ARCHIVE` and `DEFAULT_REF` before referencing them here.

tests/integration/testcases/wic/kernel.bats
```diff
@@ -0,0 +1,159 @@
...
+    run torizoncore-builder kernel build_module "${SRC_DIR}"
+    assert_failure
+    assert_output --partial "Error: could not find an Easy Installer or WIC image in the storage."
+    assert_output --partial "Please use the 'images' command to unpack an image before running this command."
+
+    torizoncore-builder-shell "rm -rf ${SRC_DIR}"
+}
+
+@test "kernel build_module: check success cases" {
+    torizoncore-builder images --remove-storage unpack "${DEFAULT_TEZI_IMAGE}"
```
lucas-akira: It should be `DEFAULT_WIC_IMAGE` here.

lucas-akira (commented):
I tried running the new kernel tests for WIC images locally, but the Common Torizon images that I've used failed during `make` with the following message:

```
make[2]: *** No rule to make target 'arch/arm64/include/generated/uapi/asm/unistd_64.h', needed by 'all'.  Stop.
```

I've tested the latest nightly images for supported TI boards: `am62xx-evm`, `am62pxx-evm`, `am62lxx-evm` and `beagley-ai`. All of them had the same error message.

I saw that you noticed a similar issue for the Luna SL1680 image as well, related to kernel versions equal or newer than 6.12: https://github.com/torizon/meta-toradex-torizon/pull/568

All of the TI images mentioned above use kernel 6.12, so it's probably the same issue.

tests/integration/testcases/wic/kernel.bats
jsrc27: Huh weird I have the declarations on my local branch. I must pushed something old. Thanks for catching it.

tests/integration/testcases/wic/kernel.bats
jsrc27: Nice catch.

jsrc27:
> All of the TI images mentioned above use kernel 6.12, so it's probably the same issue.

Thanks for checking, I wasn't sure what kernel version every common Torizon device was one. I'll just change the PR on the OS side to apply to all devices then.

tests/integration/testcases/wic/kernel.bats
jsrc27: Fixed in V2

tests/integration/testcases/wic/kernel.bats
jsrc27: Fixed in V2

lucas-akira (approved):
Latest revision looks good to me. Should we merge this PR right now or after merging the required changes of the kernel source on the OS side?

jsrc27:
> Should we merge this PR right now or after merging the required changes of the kernel source on the OS side?

Hmm I guess it doesn't matter if we merge now right? This goes into early-access first anyways, so it's not going to be available right away. If you wanna wait though for the other PR I'm okay with that too.

lucas-akira:
Right, there's the 3.16.0 release process to do soon, but I guess we can merge this now.

PR #155, opened 2026-07-15
Add support to add/remove custom kernel arguments for raw disk images using U-Boot

lucas-akira:
Add support to add and remove custom kernel arguments for raw disk images that use U-Boot as bootloader. Adding custom kernel arguments is also supported using the `build` command via its respective option in the `tcbuild.yaml` file.

Test cases were also added for this new feature.

Related-to: TCB-868

### DIFF
--- a/tcbuilder/backend/kernel.py
+++ b/tcbuilder/backend/kernel.py
@@ -28,6 +28,8 @@ OSTREE_KERNEL_SUBDIR_PATH = "usr/lib/modules/{kver}/"
 OSTREE_KERNEL_FILENAME = "vmlinuz"
 KERNEL_FIT_FILENAME = OSTREE_KERNEL_FILENAME
 
+KARGS_SUPPORTED_BOOTLOADERS = ["U-Boot"]
+
 # Name of the "function" in uEnv.txt responsible for handling boot arguments
 # passed by means of an overlay.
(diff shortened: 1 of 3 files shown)


tcbuilder/backend/kernel.py
```diff
@@ -28,6 +28,8 @@
 OSTREE_KERNEL_FILENAME = "vmlinuz"
 KERNEL_FIT_FILENAME = OSTREE_KERNEL_FILENAME
 
+KARGS_SUPPORTED_BOOTLOADERS = ["U-Boot"]
```
jsrc27, line 31: Previously we added a similar `DT_SUPPORTED_BOOTLOADERS = ["U-Boot"]`. Should we just have a common "SUPPORTED_BOOTLOADERS" for everything? Or do you think it make sense to have a per-command list?

jsrc27 (commented):
Looks good overall, just one minor comment.

tcbuilder/backend/kernel.py
lucas-akira: I think it makes sense to have a per-command list, as each command requires different work to be done in order to support another bootloader.

I imagine that the work for each command is independent from one another.

jsrc27 (approved):

PR #156, opened 2026-07-15
Support configurable sector size for raw/WIC images

vdskvortsov:
### Problem
The raw/WIC image path assumes a 512-byte logical sector throughout. On 4Kn media (such as UFS on the MediaTek Genio target) the WIC has a 4096-byte sector, so libguestfs mis-reads the partition table and `deploy` / `combine` / `images unpack` / `images provision` fail.

### Change
Add `--raw-sector-size {512,4096}` (and tcbuild `sector-size:` / `base-sector-size:` keys), threaded to every `open_disk_image()` → `add_drive_opts(blocksize=N)`. `virt-resize` can't open a 4Kn disk, so for non-512 sectors the image is grown with libguestfs directly: enlarge the file, `part_expand_gpt`, and extend the last partition — preserving its GPT identity so it still boots — then reformat. The rootfs must be the last partition (checked), which holds for WIC layouts. Default stays 512, so existing images are unaffected.

### Context
Prompted by torizon/meta-toradex-torizon#570, where 4Kn handling was done by patching the container externally; this option removes that need.

### Testing
Validated on a MediaTek Genio 1200 with 4Kn UFS: a customization exceeding the base rootfs partition exercised the grow path, and the target booted with the rootfs grown (1.6 → 5.0 GiB) and the payload present. The 512-byte path is unchanged — verified on an STM32MP target.


tcbuilder/backend/common.py
```diff
@@ -305,6 +310,12 @@ def add_common_raw_image_arguments(subparser):
                            default=None) # Default is in RAW_PROP_DEFAULTS. Default should only be
                                          # set if arg is used in a raw image.
                                          # Arg value should remain None if a tezi image is used.
+    subparser.add_argument(RAW_PROP_TO_ARGNAME["raw_sector_size"], dest="raw_sector_size",
+                           metavar="BYTES", type=int, choices=SUPPORTED_RAW_SECTOR_SIZES,
+                           help="(raw images only) logical sector size of the source WIC/raw "
+                                "image, in bytes. Use 4096 for 4Kn media (e.g. some UFS "
+                                f"targets). (default: {DEFAULT_RAW_SECTOR_SIZE})",
+                           default=None) # Default is in RAW_PROP_DEFAULTS; see raw_rootfs_label.
```
jsrc27, line 318: Wouldn't it make sense to set the default of this argument to `DEFAULT_RAW_SECTOR_SIZE`? I already see that you set this as the default throughout the various function calls and definitions that use sector size.

jsrc27 (commented):
No major comments on my end, just a minor one.

jsrc27:
Oh and also pylint is failing. Looks like it's due to many arguments/variables in some functions, due to the addition of the sector size variable everywhere in this PR. Typically for too many arguments/variables we just decide to suppress the pylint error in the sections of affected code.

Not sure if there is a way to condense this in the PR, but probably not with sector size being needed everywhere.

tcbuilder/backend/common.py
vdskvortsov: The default is set to `None` for the same reason the default is `None` for the `--raw-rootfs-label` arg above. The tezi deploy path (`do_deploy_tezi_image` in `cli/deploy.py`) warns "…is specific to raw images. Ignoring." for any non-`None` raw-only prop, so leaving the CLI default `None` keeps it quiet on Easy Installer images; the real `512` is applied from `RAW_PROP_DEFAULTS` on the raw path, the same way `raw_rootfs_label` gets `DEFAULT_RAW_ROOTFS_LABEL`. So I'd lean to keeping it `None`.

vdskvortsov:
Fixed — suppressed `too-many-positional-arguments` (and `too-many-locals` where it also tripped) on the four functions that gained the `sector_size` parameter, matching the existing convention (e.g. in `deploy.py`). Ran the CI `lint-python` pylint locally over the whole tree and it is back to 10.00/10.

tcbuilder/backend/common.py
```diff
@@ -1058,12 +1069,19 @@ def is_file_type_fit(file_path):
...
+    """Create a context manager to open and manipulate raw disk images.
+
+    :param sector_size: Logical sector size of the image, in bytes. Must be
+                        passed for 4Kn media so libguestfs can read the GPT.
+    """
 
     try:
         gfs = guestfs.GuestFS(python_return_dict=True)
-        gfs.add_drive_opts(image_path, format="raw", readonly=readonly)
+        # blocksize optarg needs libguestfs >= 1.44; omit it on the default path.
```
lucas-akira: From what I found the `blocksize` parameter was added to `add_drive_opts` in libguestfs 1.42, not 1.44:

https://libguestfs.org/guestfs-release-notes-1.42.1.html

lucas-akira (commented):
Very nice addition overall, code changes look good to me. I have one minor note on a comment.

tcbuilder/cli/build.py
rborn-tx: It looks like the forwarding of the `sector_size` is missing here:

`handle_provisioning()` calls `images_be.provision()` which now has an extra `sector_size` parameter; but sector_size parameter is not being sent to `handle_provisioning()` and forwarded to `images_be.provision()`.

tcbuilder/backend/combine.py
rborn-tx: Will this expansion work for 4Kn images?

rborn-tx (commented):
@vdskvortsov Nice addition to TorizonCore Builder. Thanks for your contribution!

I've raised a question and found a **possible** bug for your evaluation.

rborn-tx:
Reminder: since this PR is adding new switches to commands (particularly the `--raw-sector-size` switch to `deploy`, `images provision` and `images unpack`), the completion script (`torizoncore-builder-completion.bash`) would need to be updated too.

dmitrys-emcraft:
> Reminder: since this PR is adding new switches to commands (particularly the `--raw-sector-size` switch to `deploy`, `images provision` and `images unpack`), the completion script (`torizoncore-builder-completion.bash`) would need to be updated too.

Updated the completion script, added --raw-sector-size to the completion lists for `deploy`, `images provision`, and `images unpack`.

tcbuilder/backend/common.py
dmitrys-emcraft: Fixed the version number.

tcbuilder/cli/build.py
dmitrys-emcraft: Fixed - `handle_raw_image_output()` now passes `base_sector_size` to `handle_provisioning()`, which forwards it to `images_be.provision()`, `handle_provisioning()` call in `handle_easy_installer_output()` left unchanged.

tcbuilder/backend/combine.py
dmitrys-emcraft: No, it didn't - `expand_disk_partition()` is using `virt-resize`, which can't open a 4Kn disk. Fixed by using `grow_last_partition()` for 4Kn images instead.

tests/integration/testcases/wic/raw-sector-size.bats
jsrc27, line 1: Minor nitpick. Our tests typically are organized by command name. It would be nice to keep this organization if possible.

jsrc27 (commented):
Latest revision looks good to me, just one very minor comment.

tests/integration/testcases/wic/raw-sector-size.bats
dmitrys-emcraft: Fixed - moved each test into the appropriate per-command file (`images-unpack.bats`, `deploy.bats`, `combine.bats`), and the shared synthesis helpers into `tests/integration/testcases/lib/raw-sector-size.bash`.
The tests are unchanged and perform the same way as before the split.

rborn-tx (approved):
Nice! Thanks for the contribution.

jsrc27 (approved):

lucas-akira (approved):
LGTM

PR #159, opened 2026-07-21
deploy: Expand the root fs partition while keeping its disk position

lucas-akira:
Currently when deploying a raw image and the root filesystem needs to be expanded, the output disk is expanded first and its corresponding root partition is created as the last one on the disk to guarantee that it has the extra space.

This logic of creating the partition at the end of the disk isn't really necessary since virt-resize supports expanding partitions directly via the --expand option. This assures that the partition ordering of the output disk is kept the same as the input one, which can be relevant when flashing to certain machines, like the Synaptics SL1680 ones.

The change should not affect the functionality of the SYNAIMG/raw disk conversion programs included in recent Torizon OS images for the SL1680 machines, nor any current TCB features for raw disk images in general.

Related-to: TCB-842

### DIFF
--- a/tcbuilder/backend/deploy.py
+++ b/tcbuilder/backend/deploy.py
@@ -355,13 +355,13 @@ def create_output_raw_image(base_raw_img, output_raw_img, base_rootfs_partition,
     # except base_rootfs_partition:
     log.info("Starting virt-resize...")
     print("------------------------------------------------------------")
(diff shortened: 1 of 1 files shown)


rborn-tx (approved):
LGTM

jsrc27 (approved):

PR #164, opened 2026-08-06
build: Add support for inserting fusing data into image

jsrc27:
This adds functionality similar to the 'ubootenv fuses' command. However instead of ingesting a yaml file with the fuse information, this requires re-signing to be done in the same 'build' invocation. The output from the re-signing is then used to determine the fuse information.

There is also an option to output the equivalent yaml file.

This also adds related checks in the testing.

This also fixes a bug where `ubootenv fuses` was resulting in the wrong fuse order for ahab devices.

Related-to: TCB-897
Related-to: TCB-903

## Summary by CodeRabbit

- **New Features**
  - Added Easy Installer support for automatic device fusing.
  - Fusing can program fuses, close the device, and/or generate a fuse configuration file.
  - Added validation for required settings, supported hardware, secure-boot prerequisites, fuse values, and output file locations.
  - Generated images include fuse programming information and device-close status when applicable.

- **Tests**
  - Added integration coverage for fuse generation, programming values, device-close settings, missing prerequisites, and invalid fuse file locations.

### DIFF
--- a/tcbuilder/backend/platform.py
+++ b/tcbuilder/backend/platform.py
@@ -1881,6 +1881,23 @@ def push_fuse(credentials, target, version, fuse_file, hardwareids, *,
         update_description(description, target, version, credentials)
 
 
+def fuse_count_for_tech(tech):
+    """Determine the number of fuses based on the secure-boot technology
+
+    :param tech: The secure-boot technology being used.
+    :returns: An int with the number of fuses
+    """
+
+    if tech == "hab":
+        fuse_num = 8
+    elif tech == "ahab":
+        fuse_num = 16
+    else:
+        assert False, f"Unhandled Secure Boot technology '{tech}'."
+
+    return fuse_num
+
+
 def fuse_hwid_to_machine(hwid):
     """Translate a fuse hwid to a machine name."""
     if not hwid.endswith(FUSE_HWID_SUFFIX):
@@ -1915,12 +1932,7 @@ def validate_fuse_file(fuse_file, hardwareids):
             f"Provided hardware ids: {hardwareids} are associated with different"
             " Secure Boot technologies which is not possible in one fuse type package.")
 
-    if "hab" in secboot_techs:
-        fuse_num = 8
-    elif "ahab" in secboot_techs:
-        fuse_num = 16
-    else:
-        assert False, f"Unhandled Secure Boot technology '{secboot_techs}'."
+    fuse_num = fuse_count_for_tech(next(iter(secboot_techs)))
 
     # Next check overall layout against schema
     parsed_file = parse_config_file(fuse_file, schema_path=FUSE_SCHEMA_FILE)
--- a/tcbuilder/backend/tcbuild.schema.yaml
+++ b/tcbuilder/backend/tcbuild.schema.yaml
@@ -629,6 +629,24 @@ properties:
                 $ref: '#/properties/output/definitions/provisioning'
               bundle:
                 $ref: '#/properties/output/definitions/bundle'
+              fusing:
+                type: object
+                description: "Automatic fusing configuration"
+                properties:
+                  auto-programming:
+                    type: boolean
+                    description: "Whether to enable auto-fusing or not"
+                  close-device:
+                    type: boolean
+                    description: "Whether to set closing fuse during auto-fusing"
+                  fuse-file:
+                    type: string
+                    description: "Name of file to create containing fusing information for Torizon Cloud"
+                    pattern: '\.ya?ml$'
+                required:
+                  - auto-programming
+                  - close-device
+                additionalProperties: false
             required: [local]
             # No extra props inside 'easy-installer'
             additionalProperties: false
--- a/tcbuilder/backend/ubootenv.py
+++ b/tcbuilder/backend/ubootenv.py
@@ -6,11 +6,13 @@ import os
 import logging
 import re
 import shutil
+import tempfile
 
 from tcbuilder.backend.common import \
-    (set_output_ownership, get_tezi_image_version)
+    (set_output_ownership, get_tezi_image_version, FUSE_CMD_TXT_NAME)
 from tcbuilder.backend.platform import \
-    (validate_fuse_file, FUSE_HARDWAREIDS, restore_hex)
+    (validate_fuse_file, FUSE_HARDWAREIDS, restore_hex,
+     SECBOOT_TECH_PER_MACHINE, fuse_count_for_tech)
 from tcbuilder.errors import \
     (FileContentMissing, TorizonCoreBuilderError, InvalidStateError,
      PathNotExistError)
@@ -40,22 +42,7 @@ def env_fuses(input_dir, output_dir, fuse_file, force=False):
                     "OS before 7.2.0. Proceeding anyways, but this will most likely not "
                     "do anything.")
 
-    initial_env_filename = get_env_filename(input_dir)
-    initial_env = os.path.join(input_dir, initial_env_filename)
-
-    with open(initial_env, "r", encoding="utf-8") as file:
-        env = file.read()
-
-    # Check to see if initial u-boot env already contains fuse variables
-    if any(fuse_var in env for fuse_var in FUSE_VARIABLES):
-        raise InvalidStateError(
-            "Input image already contains fuse related u-boot variables")
-
-    board = find_board(env)
-    hwid = board + "-fuses"
-    if hwid not in FUSE_HARDWAREIDS:
-        raise InvalidStateError(
-            f"Hardware type \"{board}\" is not supported by this command.")
+    hwid, _ = _detect_fuse_hwid(input_dir)
 
     # Validate and check fuse file before proceeding
     fuse_file_data = validate_fuse_file(fuse_file, [hwid])
@@ -63,10 +50,11 @@ def env_fuses(input_dir, output_dir, fuse_file, force=False):
     var_list = get_fuse_vars(fuse_file_data)
 
     inplace = False
+    initial_env_filename = get_env_filename(input_dir)
     if os.path.normpath(output_dir) == os.path.normpath(input_dir):
         log.debug("Updating Torizon OS image in place.")
         output_dir = input_dir
-        output_env = initial_env
+        output_env = os.path.join(input_dir, initial_env_filename)
         inplace = True
     else:
         if os.path.exists(output_dir) and force:
@@ -80,6 +68,9 @@ def env_fuses(input_dir, output_dir, fuse_file, force=False):
         write_uboot_vars(output_env, var_list)
         set_output_ownership(output_dir)
         log.info("Fuse U-Boot variables successfully added to image.")
+        log.info("WARNING: Any device flashed with this image will perform "
+                 "automatic fusing for secure-boot. This is an irreversible "
+                 "operation. Take care to check before flashing this image.")
     except:
         if not inplace:
             log.debug("Removing output directory due to error.")
@@ -173,3 +164,81 @@ def write_uboot_vars(env_file, var_list):
 
     with open(env_file, "a", encoding="utf-8") as file:
         file.writelines(var + '\n' for var in var_list)
+
+
+def _detect_fuse_hwid(image_dir):
+    """Determine the hardware-id based on the input image
+
+    :param image_dir: Path to directory containing input image
+    :returns: A tuple containing the detected hardware-id and board
+    """
+
+    initial_env_filename = get_env_filename(image_dir)
+    initial_env = os.path.join(image_dir, initial_env_filename)
+
+    with open(initial_env, "r", encoding="utf-8") as file:
+        env = file.read()
+
+    # Check to see if initial u-boot env already contains fuse variables
+    if any(fuse_var in env for fuse_var in FUSE_VARIABLES):
+        raise InvalidStateError(
+            "Input image already contains fuse related u-boot variables")
+
+    board = find_board(env)
+    hwid = board + "-fuses"
+    if hwid not in FUSE_HARDWAREIDS:
+        raise InvalidStateError(
+            f"Hardware type \"{board}\" is not supported by this command.")
+
+    return hwid, board
+
+
+# pylint: disable-next=too-many-locals
+def generate_fuse_file(close, image_dir):
+    """Generate valid fuse.yaml file based on re-signing information
+
+    :params close: Boolean on whether closing should be set in the yaml
+    :params image_dir: Path to image directory containing the signed image
+    :returns: Path to generated yaml file
+    """
+
+    fuse_instructions = os.path.join(image_dir, FUSE_CMD_TXT_NAME)
+    if not os.path.isfile(fuse_instructions):
+        raise PathNotExistError("Unable to find fusing instructions file: "
+                                f"{FUSE_CMD_TXT_NAME}. Is the image signed?")
+
+    # Figure out what machine this is for
+    hwid, board = _detect_fuse_hwid(image_dir)
+
+    _tech = SECBOOT_TECH_PER_MACHINE.get(board)
+    fuse_num = fuse_count_for_tech(_tech)
+
+    # Parse fusing values from fusing instruction file created by re-signing
+    hex_regex = re.compile(r"\b(?:0[xX])[0-9a-fA-F]+\b")
+
+    fuse_values = []
+    with open(fuse_instructions, 'r', encoding="utf-8") as file:
+        for line in file:
+            fuse_value = hex_regex.findall(line)
+            if fuse_value:
+                fuse_values.append(fuse_value[0])
+
+            if len(fuse_values) == fuse_num:
+                break
+
+    # Create temporary fuse yaml file to be used later
+    with tempfile.NamedTemporaryFile(mode='w+', delete=False) as tmp_fuse_file:
+        tmp_fuse_file.write("fuses:\n")
+        _index = 1
+        for value in fuse_values:
+            tmp_fuse_file.write(f"  fuse-val{_index}: {value}\n")
+            _index += 1
+
+        tmp_fuse_file.write(f"  fuse-close: {close}\n")
+
+    tmp_fuse_path = tmp_fuse_file.name
+
+    # Quick sanity check on generated file
+    validate_fuse_file(tmp_fuse_path, [hwid])
+
+    return tmp_fuse_path
--- a/tcbuilder/backend/ubootenv.py
+++ b/tcbuilder/backend/ubootenv.py
@@ -135,13 +135,16 @@ def get_fuse_vars(fuse_data):
     restore_hex(fuse_data)
 
     val_list = ""
-    for key in sorted(fuse_data['fuses'].keys()):
-        if "fuse-val" in key:
-            num = fuse_data["fuses"].get(key)
-            if not val_list:
-                val_list = num
-            else:
-                val_list = val_list + " " + num
+    # Subtract one to account for 'fuse-close'
+    fuse_num = len(fuse_data['fuses']) - 1
+    # Parse fuse values in order
+    for index in range(fuse_num):
+        # Add 1 cause range starts at 0
+        num = fuse_data["fuses"].get(f"fuse-val{index + 1}")
+        if not val_list:
+            val_list = num
+        else:
+            val_list = val_list + " " + num
 
     fuse_close = fuse_data["fuses"].get("fuse-close")
     if fuse_close:
--- a/tcbuilder/cli/build.py
+++ b/tcbuilder/cli/build.py
@@ -26,6 +26,7 @@ from tcbuilder.backend import images as images_be
 from tcbuilder.backend import kernel as kernel_be
 from tcbuilder.backend import ostree as ostree_be
 from tcbuilder.backend import secboot as secboot_be
+from tcbuilder.backend import ubootenv as ubootenv_be
 from tcbuilder.cli import deploy as deploy_cli
 from tcbuilder.cli import dt as dt_cli
 from tcbuilder.cli import dto as dto_cli
@@ -554,7 +555,8 @@ def _union_parse_ostree_key(ostree_key_dir, ostree_key_props, *,
     return ostree_key_obj
 
 
-def handle_output_section(props, changes_dirs=None, default_base_raw_image=None):
+def handle_output_section(props, changes_dirs=None, default_base_raw_image=None,
+                          sign_type=None):
     """Handle the output section of the configuration file
 
     :param props: Dictionary holding the data of the section.
@@ -562,6 +564,7 @@ def handle_output_section(props, changes_dirs=None, default_base_raw_image=None)
     :param default_base_raw_image: Default base raw image. Should always be the
                                    input raw image. If dealing with tezi images,
                                    this value is equal to 'None'.
+    :param sign_type: Secure-boot signing technology being used.
     """
 
     if props:
@@ -616,7 +619,7 @@ def handle_output_section(props, changes_dirs=None, default_base_raw_image=None)
 
     if common.unpacked_image_type() == "tezi":
         tezi_props = props.get("easy-installer", {})
-        handle_easy_installer_output(tezi_props, union_params)
+        handle_easy_installer_output(tezi_props, union_params, sign_type)
     else:
         raw_props = props.get("raw-image", {})
         handle_raw_image_output(raw_props, union_params, default_base_raw_image)
@@ -667,11 +670,12 @@ def handle_raw_image_output(props, union_params, default_base_raw_image):
                              base_sector_size)
 
 
-def handle_easy_installer_output(props, union_params):
+def handle_easy_installer_output(props, union_params, sign_type=None):
     """Handle the output/easy-installer section of the configuration file
 
     :param props: Dictionary holding the data of the section.
     :param union_params: Parameters related to union(). This is a required arg.
+    :param sign_type: Secure-boot technology being used.
     """
 
     # Note that the following test should never fail (due to schema validation).
@@ -696,6 +700,9 @@ def handle_easy_installer_output(props, union_params):
 
     if "provisioning" in props:
         handle_provisioning(output_dir, props.get("provisioning"))
+
+    if "fusing" in props:
+        handle_fusing(output_dir, props.get("fusing"), sign_type)
 
 
 def handle_bundle_common(bundle_props, compress_tar=True):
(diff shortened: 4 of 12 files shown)


tcbuilder/cli/build.py
```diff
@@ -878,6 +882,41 @@ def handle_provisioning(output_dir, prov_props, rootfs_label=None, sector_size=N
     images_be.provision(**prov_params)
 
 
+def handle_fusing(output_dir, fuse_props):
+    """Handle the fusing step of the output generation."""
+
+    # Check to make sure re-signing of bootloader was done earlier
+    if fuse_props.get("re-sign") is not True:
```
lucas-akira: [PEP 8](https://peps.python.org/pep-0008/#programming-recommendations) recommends doing comparisons to None with `is None`/`is not None`:
```
if fuse_props.get("re-sign") is None:
```

tcbuilder/backend/ubootenv.py
```diff
@@ -173,3 +176,59 @@ def write_uboot_vars(env_file, var_list):
...
+                                f"{secboot.FUSE_CMD_TXT_NAME}. Is the image signed?")
+
+    # Parse fusing values from fusing instruction file created by re-signing
+    hex_regex = re.compile(r"\b(?:0[xX])[0-9a-fA-F]+\b")
+
+    fuse_values = []
+    with open(fuse_instructions, 'r', encoding="utf-8") as file:
+        for line in file:
+            if not line.strip():
+                # There is an empty line after the 8/16 fuse values
```
lucas-akira: I think it would be more robust to break the loop once `fuse_values` has the expected length i.e. 8 for HAB, 16 for AHAB.

Breaking the loop at the presence of an empty line can cause problems if `fuse-cmds.txt` has any extra empty lines before the fuse values or between them, which may happen if any additional comments are added for example.

lucas-akira (commented):
I made some comments, but things look good overall.

tcbuilder/cli/build.py
jsrc27: Done in V2

tcbuilder/backend/ubootenv.py
jsrc27: Done in V2

tcbuilder/cli/build.py
```diff
@@ -878,6 +882,49 @@ def handle_provisioning(output_dir, prov_props, rootfs_label=None, sector_size=N
...
+            output_root = os.path.realpath(output_dir)
+            fuse_output = os.path.realpath(os.path.join(output_root, _do_file))
+            if (fuse_output == output_root or
+                    os.path.commonpath((output_root, fuse_output)) != output_root):
+                raise InvalidDataError("'fuse-file' must be inside the output directory.")
+            shutil.copy(fuse_file, fuse_output)
+            common.set_output_ownership(fuse_output)
+    finally:
+        if os.path.exists(fuse_file):
+            os.remove(fuse_file)
```
rborn-tx, line 930: ### `handle_fusing()` cleanup masks `generate_file()` failures

`fuse_file` is bound by the first statement in the `try`. If `generate_file()` raises, the `finally` clause runs with `fuse_file` unbound and replaces the useful original error with `UnboundLocalError`.

Initialize `fuse_file = None` before the `try`, then guard cleanup with `if fuse_file and os.path.exists(fuse_file):`.

Related: `generate_file()` creates its temporary file before validation but returns its path only after validation succeeds. If validation raises, the file leaks in `/tmp`; clean it up in `generate_file()` before re-raising.

tcbuilder/backend/ubootenv.py
```diff
@@ -6,11 +6,15 @@
...
+import tempfile
 
 from tcbuilder.backend.common import \
     (set_output_ownership, get_tezi_image_version)
 from tcbuilder.backend.platform import \
-    (validate_fuse_file, FUSE_HARDWAREIDS, restore_hex)
+    (validate_fuse_file, FUSE_HARDWAREIDS, restore_hex,
+     SECBOOT_TECH_PER_MACHINE)
+# Import whole file to avoid circular import
+from tcbuilder.backend import secboot
```
rborn-tx: ### The new `ubootenv` → `secboot` import is a genuine cycle

This module-level import does not avoid the cycle: `secboot.py` imports `get_env_filename` and `find_board` from `ubootenv`, but those names are not yet defined when an import enters via `ubootenv`. Directly importing `tcbuilder.backend.ubootenv` therefore regresses to an import-from-partially-initialized-module error.

Move `FUSE_CMD_TXT_NAME` to a dependency-neutral module, or use a function-local import in `generate_file()` if it cannot be moved. The current comment should also be corrected.

tcbuilder/cli/build.py
```diff
@@ -878,6 +882,49 @@ def handle_provisioning(output_dir, prov_props, rootfs_label=None, sector_size=N
     images_be.provision(**prov_params)
 
 
+def handle_fusing(output_dir, fuse_props):
+    """Handle the fusing step of the output generation."""
```
rborn-tx, line 889: ### All `fusing` validation happens after the whole build has run

`handle_fusing()` runs after input unpacking, module building, re-signing, OSTree work, Tezi deployment, bundling, and provisioning. Both the re-signing prerequisite and the `fuse-file` containment rule are static configuration checks, so a typo can waste the entire build before the enclosing error handler deletes the output.

Validate these conditions in `_build_setup()` (or a dedicated pre-flight helper). This also removes the need for the injected internal `re-sign` key.

tcbuilder/cli/tcbuild.template.yaml
```diff
@@ -263,3 +263,11 @@ output:
       # >> Replace example UUID with your actual fleet UUID.
       # fleets:
         # - 0dfab76b-5b71-4435-ab6f-c6cdc62690c6
+    # fusing:
```
rborn-tx, line 268: ### Automatic OTP fuse programming has no safety warning

The manual-fusing path prominently warns that fuse programming is irreversible. This new automatic path only says that automatic fuse programming is being added, even though `close-device: true` permanently closes every device that installs the image and a PKI error can brick units.

Please add a prominent irreversible-operation warning to the template and a build-time `log.warning()` when auto-programming is enabled, with stronger wording when `close-device` is true.

tcbuilder/cli/build.py
```diff
@@ -955,6 +1002,15 @@ def build(config_fname, *, substs=None, enable_subst=True, force=False):
     except KeyError as _exc:
         pass
 
+    # Track whether bootloader re-signing is being done for auto-fusing
+    try:
+        if config.get("customization", {}).get("secboot", {}).get("sign-bootloader-hab"):
+            config["output"]["easy-installer"]["fusing"]["re-sign"] = True
```
rborn-tx: ### The injected `re-sign` config key is fragile

This writes an undeclared member into a schema-validated user object and suppresses every `KeyError` while finding that object. It only transports one boolean, unlike the adjacent `sign-ostree` relocation.

Pass an explicit `resigned` boolean into `handle_easy_installer_output()` / `handle_fusing()` instead. That is clearer and permits the prerequisite check to run during pre-flight validation. `InvalidDataError` also better describes an invalid configuration combination than `InvalidStateError`.

P.S. (from @rborn-tx): I imagine this direct change of the parsed config was done to mimic the sign-ostree relocation a few lines above. The reason for that relocation is that OSTree signing is fully processed during output processing, but from the user's perspective, it's considered a customization, which is what made the relocation legitimate. The goal wasn't to use the parsed config to pass data around, as the "re-sign" field seems to be used here.

tcbuilder/cli/build.py
```diff
@@ -878,6 +882,49 @@ def handle_provisioning(output_dir, prov_props, rootfs_label=None, sector_size=N
...
+def handle_fusing(output_dir, fuse_props):
+    """Handle the fusing step of the output generation."""
+
+    # Check to make sure re-signing of bootloader was done earlier
+    if fuse_props.get("re-sign") is None:
+        raise InvalidStateError("To do 'fusing', 'sign-bootloader-hab' must be defined "
+                                "at the same time.")
+
+    _do_fuse = fuse_props.get("auto-programming", False)
+    _do_file = fuse_props.get("fuse-file", [])
```
rborn-tx: ### `fuse-file` defaults to a list despite being string-typed

`fuse_props.get("fuse-file", [])` works only because both a list and a string can be falsy, but `[]` is confusing for a string-typed schema property. Use `None` or an empty string as the sentinel.

tcbuilder/backend/ubootenv.py
```diff
@@ -173,3 +177,69 @@ def write_uboot_vars(env_file, var_list):
...
+    :params image_dir: Path to image directory containing the signed image
+    :returns: Path to generated yaml file
+    """
+
+    fuse_instructions = os.path.join(image_dir, secboot.FUSE_CMD_TXT_NAME)
+    if not os.path.isfile(fuse_instructions):
+        raise PathNotExistError("Unable to find fusing instructions file: "
+                                f"{secboot.FUSE_CMD_TXT_NAME}. Is the image signed?")
+
+    # Figure out what machine this is for
```
rborn-tx: ### Board detection and HAB/AHAB fuse-count logic are duplicated

`generate_file()` repeats the environment-file → board → HWID detection in `env_fuses()`, and repeats the HAB=8 / AHAB=16 mapping in `validate_fuse_file()`. These paths will drift as secure-boot support expands.

Extract `_detect_fuse_hwid(image_dir)` and a shared `fuse_count_for_tech(tech)` helper, then use them from all three sites. Renaming `generate_file()` to `generate_fuse_file()` would also make the public call site clearer.

tcbuilder/backend/ubootenv.py
```diff
@@ -173,3 +177,69 @@ def write_uboot_vars(env_file, var_list):
...
+
+    # Parse fusing values from fusing instruction file created by re-signing
+    hex_regex = re.compile(r"\b(?:0[xX])[0-9a-fA-F]+\b")
+
+    fuse_values = []
+    with open(fuse_instructions, 'r', encoding="utf-8") as file:
+        for line in file:
+            fuse_value = hex_regex.findall(line)
+            if fuse_value:
+                fuse_values.append(fuse_value[0].strip("'"))
```
rborn-tx: ### Dead quote stripping in fuse-value parsing

`fuse_value[0].strip("'")` is dead: the regex `\b(?:0[xX])[0-9a-fA-F]+\b` cannot capture a quote.

tcbuilder/cli/build.py
```diff
@@ -878,6 +882,49 @@ def handle_provisioning(output_dir, prov_props, rootfs_label=None, sector_size=N
...
+            ubootenv_be.env_fuses(**ubootenv_params)
+
+        if _do_file:
+            log.info(l2_pref(f"Creating fuse-file '{fuse_props['fuse-file']}'."))
+            output_root = os.path.realpath(output_dir)
+            fuse_output = os.path.realpath(os.path.join(output_root, _do_file))
+            if (fuse_output == output_root or
+                    os.path.commonpath((output_root, fuse_output)) != output_root):
+                raise InvalidDataError("'fuse-file' must be inside the output directory.")
+            shutil.copy(fuse_file, fuse_output)
```
rborn-tx, line 925: ### Generated `fuse.yaml` is installed with restrictive permissions

`NamedTemporaryFile` creates mode `0600`, and `shutil.copy` preserves permission bits, so the delivered `fuse.yaml` remains `0600` while the other Tezi output files are `0644`. `set_output_ownership()` changes ownership only.

Set mode `0644` after the copy, or use `shutil.copyfile`, to match the rest of the output.

tcbuilder/backend/tcbuild.schema.yaml
```diff
@@ -629,6 +629,27 @@ properties:
...
+              fusing:
+                type: object
+                description: "Automatic fusing configuration"
+                properties:
+                  auto-programming:
+                    type: boolean
+                    description: "Whether to enable auto-fusing or not"
+                  close-device:
+                    type: boolean
+                    description: "Whether to set closing fuse during auto-fusing"
```
rborn-tx, line 641: ### `fuse-file` schema validation is deferred and incomplete

The schema accepts `fuse-file: fuse.txt`, which only fails later in `canonicalize_fuse_file()`, and accepts nested output paths without ensuring their parent directory exists before `shutil.copy`.

Add a YAML filename constraint such as `pattern: '\.ya?ml$'` and validate the output parent path early so users receive a direct configuration error.

tcbuilder/backend/tcbuild.schema.yaml
```diff
@@ -629,6 +629,27 @@ properties:
                 $ref: '#/properties/output/definitions/provisioning'
               bundle:
                 $ref: '#/properties/output/definitions/bundle'
+              fusing:
```
rborn-tx, line 632: ### Raw-image output omits the fusing feature

Consider stating that limitation in the template (`tcbuilder/cli/tcbuild.template.yaml`).

tcbuilder/cli/tcbuild.template.yaml
```diff
@@ -263,3 +263,11 @@ output:
       # >> Replace example UUID with your actual fleet UUID.
       # fleets:
         # - 0dfab76b-5b71-4435-ab6f-c6cdc62690c6
```
rborn-tx, line 265: ### The fusing template obscures enabled state and the required field

`>> Enabling automatic fusing` immediately followed by `auto-programming: false` reads as contradictory. The schema also requires `close-device` whenever the `fusing` object is present, but the template presents all three fields as equally optional.

Clarify how the section is enabled and identify `close-device` as required.

tests/integration/testcases/build.bats
```diff
@@ -273,6 +277,17 @@ teardown_file() {
     run test "${FOUND_KEY_NAME}" == "${KERNEL_KEY_NAME}"
     assert_success
 
+    # Check fusing output
```
rborn-tx, line 280: ### The new coverage checks only fuse-value presence, not correctness

The test verifies only that `fuse_status=pending` exists and that a `fuse-close: False` line was written. It does not assert that the eight generated values and `fuse_prog_list` match `fuse-cmds.txt` in the correct order.

Please compare those values directly. Also cover `close-device: true`, fusing without re-signing, `fuse-file` path escape, and fuse-file-only mode. The current `False` assertion also bakes non-canonical boolean spelling into the suite.

tcbuilder/backend/tcbuild.schema.yaml
```diff
@@ -629,6 +629,27 @@ properties:
...
+                    type: boolean
+                    description: "Whether to set closing fuse during auto-fusing"
+                  fuse-file:
+                    type: string
+                    description: "Name of file to create containing fusing information for Torizon Cloud"
+                anyOf:
+                  - required:
+                      - auto-programming
+                  - required:
+                      - fuse-file
```
rborn-tx: Wouldn't it be preferable to make `auto-programming` a required property just like `close-device`?

rborn-tx (changes requested):
@jsrc27 I've asked a review to Claude, did some filtering and added some personal comments. The result is posted here.

Besides these, Claude found an apparent major bug which would probably require a separate ticket to solve. I'll post it separately for completeness.

rborn-tx:
As mentioned in the review, Claude also found this issue. It was found while reviewing `generate_file()`'s AHAB branch.

---

**`get_fuse_vars()` orders fuse values lexicographically** — `tcbuilder/backend/ubootenv.py:151`:

```python
    for key in sorted(fuse_data['fuses'].keys()):
```

For HAB (8 values) `sorted()` happens to be correct. For AHAB (16 values) it is not. Reproduced with a valid 16-value `apalis-imx8-fuses` file whose values are `0x01010101 … 0x10101010` in order:

```
fuse_prog_list=0x1010101 0xa0a0a0a 0xb0b0b0b 0xc0c0c0c 0xd0d0d0d 0xe0e0e0e
               0xf0f0f0f 0x10101010 0x2020202 0x3030303 0x4040404 0x5050505
               0x6060606 0x7070707 0x8080808 0x9090909
```

`fuse-val10` through `fuse-val16` sort before `fuse-val2`. The resulting `fuse_prog_list` writes the AHAB SRK hash into the OTP fuses in the wrong order — which for a one-time-programmable region means an unrecoverable device.

This affects the shipped `ubootenv fuses` command today for `apalis-imx8` and `colibri-imx8x`.

---

I've created a ticket for this (TCB-903).

tcbuilder/cli/build.py
jsrc27: Sounds sensible. Though thinking further, pragmatically speaking I guess the topic of temporary file cleanup is a bit unneeded. These are temporary files in a container. Cleanup is basically "free" in this context.

But it's minor to fix so whatever I guess.

tcbuilder/backend/ubootenv.py
jsrc27: Are we okay with moving `FUSE_CMD_TXT_NAME` to `common.py`? Or is it better to do a function-local import?

Before submitting the PR I did think about moving `FUSE_CMD_TXT_NAME` but I wasn't sure how we felt about that. Currently we have no precedence for function-local imports I believe, so I'm not sure if we're okay with that as well.

tcbuilder/cli/build.py
jsrc27: Hmm I can see the point, but this would be inconsistent with how the other `handle_*` sections do their own local checks/validation. Do we prefer doing specific section validation before any actual work in `build` is performed? Or, do we stick with the current precedent.

tcbuilder/cli/tcbuild.template.yaml
jsrc27: Okay sure, we can add additional warnings. Is `log.warning()` right here? Or is `log.info` sufficient? Cause warning has a negative connotation.

tcbuilder/cli/build.py
jsrc27: So my original idea here was to pass a parameter all the way through the various functions down to `handle_fusing()`, but this seemed "off" to me. In the sense that it would be adding an additional parameter to pass through other functions where it's not used to only be used in the final `handle_fusing()`.

Is adding additional function parameters really preference here?

tcbuilder/cli/build.py
jsrc27: Okay sure.

tcbuilder/backend/ubootenv.py
jsrc27: Sure this would need some refactoring, but if it is preferable then okay.

tcbuilder/backend/ubootenv.py
jsrc27: This seems to be correct. I was parsing the fuse vales in a slightly different way at first, that was leaving quotes around the parsed hex values. But, the current method in the PR seems to no longer do so.

tcbuilder/cli/build.py
jsrc27: Sure I suppose.

tcbuilder/backend/tcbuild.schema.yaml
jsrc27: Well I didn't bother checking for the existence of `output_dir` since it's essentially guaranteed to have been created by the processes running prior to `handle_fusing`. If those processes failed which would mean `output_dir` does not exist, then the `build` command would have failed before `handle_fusing` is reached. We can add the check but it seems unnecessary in this context.

Are we okay with adding the `yaml/yml` constraint to the filename? We don't currently constrain the compose file, in the schema so I didn't bother to do it here.

tcbuilder/backend/tcbuild.schema.yaml
jsrc27: Sure. I mean the template already only shows `fusing:` under `easy-installer`. But if we want to be more explicit about it.

tcbuilder/cli/tcbuild.template.yaml
jsrc27: I'm not sure I even understand what this feedback is suggesting.

tests/integration/testcases/build.bats
jsrc27: Sure.

tcbuilder/backend/tcbuild.schema.yaml
jsrc27: If you want. In the original planning document for this I interpreted `auto-programming` or `fuse-file` to be valid.

jsrc27:
> As mentioned in the review, Claude also found this issue. It was found while reviewing generate_file()'s AHAB branch.

As discussed internally, I will try to address this issue in this PR.

tcbuilder/cli/build.py
rborn-tx: Yeah, I don't see the need for fixing the "Related" issue found by Claude (which would go inside `generate_file()`). I'd only solve the main issue of the unbound variable by following the Claude's suggestion.

tcbuilder/backend/ubootenv.py
rborn-tx: > Are we okay with moving FUSE_CMD_TXT_NAME to `common.py`?

For the short-term I'd say yes. In the future we'll need a full refactor to solve the issue properly (e.g. making a `lib` package with independent modules (no cross-dependencies)).

tcbuilder/cli/build.py
rborn-tx: I just posted what Claude found, but I'd be okay with keeping our current approach with it's non-ideal behavior given that the same issue is present with most other types of customization done by `build`.

I've checked Jira and found no tickets about this topic, so I've created one: TCB-912

tcbuilder/cli/tcbuild.template.yaml
rborn-tx: I'd add `# WARNING: ...` to the template, and `log.info("WARNING: ...")` to the code.

tcbuilder/cli/build.py
rborn-tx: Well, I wouldn't use the parsed configuration as a way to pass *information* across functions. Passing parameters is the standard/cleaner way of doing that.

I imagine you say it looks off because the intermediate functions wouldn't use the information directly. But that's usual, functions may take parameters that they only forward until the real interested party gets it.

---

My proposal would be very close to what Claude suggested, but with an eye in our future implementations for AHAB. The call-chain would look like this:

```
build()
  sign_type = ...
  -> handle_output_section(..., sign_type)
     -> handle_easy_installer_output(..., sign_type)
        -> handle_fusing(..., sign_type)
```

So, `sign_type` would be forwarded three levels down, which doesn't look so bad.

Notice I wrote `sign_type` to be a bit more generic and allow the addition of `ahab` later. I'd set `sign_type` like this:

```
    # Track whether bootloader re-signing is being done for auto-fusing
    _sign_type = None
    try:
        if "sign-bootloader-hab" in config["customization"]["secboot"]:
            _sign_type = "hab"
        # in the future:
        #elif "sign-bootloader-ahab" in config["customization"]["secboot"]:
        #    _sign_type = "ahab"
    except KeyError as _exc:
        pass
...
    handle_output_section(..., sign_type=_sign_type)
```

That's basically the first idea.

Another idea would be to not pass the information to `handle_fusing` at all (via the call-chain), and instead use information from state (storage). We could save some information in storage to indicate the bootloader was re-signed and then use it in `handle_fusing`.

tcbuilder/backend/tcbuild.schema.yaml
rborn-tx: I'm in favor of constraining the filename (even of the compose file if you want too).

Regarding the output directory, Claude is saying that the user can pass "nested" directories which are not guaranteed to exist (e.g. `fuse-file: a/b/fuse.yml`). We don't handle this situation in other cases, so it would be fine to not do it here.

tcbuilder/cli/tcbuild.template.yaml
rborn-tx: Yeah, it's definitely not very clear. I'd translate it to this:

```
      # >> Enable/disable automatic fusing (REQUIRED)
           ^^^^^^^^^^^^^^                   ^^^^^^^^
        (avoid the contradiction)      (by my other suggestion)
      # auto-programming: true
      # >> Automatically close device during fusing (REQUIRED)
      # close-device: false
      # >> When set, output a file containing the fusing information that can be used
      # >> with Torizon Cloud.
      # fuse-file: fuse.yaml
```

Question that came to mind: is the combination `auto-programming: false` and `close-device: true` accepted?

tcbuilder/backend/tcbuild.schema.yaml
rborn-tx: Hmm, I think I missed there the information of whether the presence of the properties would be optional or not.

I believe forcing users to set both `auto-programming` and `close-device` (whenever the fusing section is present) would be less error prone. What do you think?

I'd keep `fuse-file` optional since the effect of generating/not-generating the file has low potential for causing trouble to users.

tcbuilder/cli/build.py
jsrc27: Addressed in the latest revision.

tcbuilder/backend/ubootenv.py
jsrc27: Addressed in the latest revision. Went ahead and moved `FUSE_CMD_TXT_NAME` to `common.py`

tcbuilder/cli/build.py
jsrc27: As discussed won't fix.

tcbuilder/cli/tcbuild.template.yaml
jsrc27: Addressed in the latest revision.

tcbuilder/cli/build.py
jsrc27: Addressed in the latest revision. Went ahead and just added a new parameter to pass all the way down to `handle_fusing`.

tcbuilder/cli/build.py
jsrc27: Addressed in the latest revision.

tcbuilder/backend/ubootenv.py
jsrc27: Addressed in the latest revision.

tcbuilder/backend/ubootenv.py
jsrc27: Addressed in the latest revision.

tcbuilder/cli/build.py
jsrc27: Addressed in the latest revision.

tcbuilder/backend/tcbuild.schema.yaml
jsrc27: Addressed in the latest revision. Now the schema enforced `yaml/yml` extension on the passed filename.

tcbuilder/backend/tcbuild.schema.yaml
jsrc27: Addressed in the latest revision.

tcbuilder/cli/tcbuild.template.yaml
jsrc27: Addressed in the latest revision.

> Question that came to mind: is the combination auto-programming: false and close-device: true accepted?

Yes there is a check for that case where we print a message that nothing will be done regarding the `fusing` section then we just return from `handle_fusing` early without doing any work.

tests/integration/testcases/build.bats
jsrc27: Addressed in the latest revision.

tcbuilder/backend/tcbuild.schema.yaml
jsrc27: Addressed in the latest revision.

I went and made `auto-programming` and `close-device` required parameters while `fuse-file` is optional.

jsrc27:
> As discussed internally, I will try to address this issue in this PR.

Latest revision has an additional commit to try and address this bug.

tcbuilder/backend/ubootenv.py
rborn-tx: **Residual nit (by Claude):** `backend/deploy.py:26` still does `from tcbuilder.backend.secboot import FUSE_CMD_TXT_NAME`, i.e. it reaches the constant through `secboot`'s re-export rather than from `common`. It works, but it keeps a `deploy → secboot` edge alive for nothing and undoes part of the point of moving the constant. One-line change.

tcbuilder/cli/tcbuild.template.yaml
rborn-tx: Shouldn't we have the `(REQUIRED)` text in the `close-device` header?

```
      # >> Automatically close device during fusing (REQUIRED)
      # close-device: false
```

rborn-tx:
There's some incident going on with GitHub so I couldn't access the files to post these residual nits (by Claude):

- Typo in the new commented-out AHAB branch, `build.py:1015`: "Once ahab **devics** are supported".
- Stray double blank line at `ubootenv.py:148-149`.

tcbuilder/backend/platform.py
```diff
@@ -1881,6 +1881,23 @@ def push_fuse(credentials, target, version, fuse_file, hardwareids, *,
...
+    """
+
+    if tech == "hab":
+        fuse_num = 8
+    elif tech == "ahab":
+        fuse_num = 16
+    else:
+        assert False, f"Unhandled Secure Boot technology '{tech}'."
+
+    return fuse_num
```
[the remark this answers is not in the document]
rborn-tx: We're accepting this.

tcbuilder/backend/ubootenv.py
jsrc27: Fixed in latest revision.

tcbuilder/cli/tcbuild.template.yaml
jsrc27: Added in latest revision.

jsrc27:
> Typo in the new commented-out AHAB branch, build.py:1015: "Once ahab devics are supported".

Fixed in latest revision.

> Stray double blank line at ubootenv.py:148-149.

Removed in latest revision.

rborn-tx (approved):

lucas-akira (approved):
Latest revision looks good to me.

PR #165, opened 2026-08-07
Fix circular import and grow_last_partition() cleanup

dmitrys-emcraft:
Two follow-up fixes for #156.

## common: break the dt/deploy/secboot/kernel circular import

A pre-existing bug encountered while working on #156.

`dt.py` importing `get_kernel_changes_dir` from `kernel.py` and
`get_image_bootloader` from `deploy.py` closes a cycle:
`dt -> deploy -> secboot -> kernel -> dt`. Move them to
`common.py`, which has no dependency back into the ring; 
and make `kernel.py` and `deploy.py` re-export the relocated 
functions so existing call sites are unaffected.

## deploy: clean up `grow_last_partition()` on a mid-way failure

`grow_last_partition()` left a corrupt image on disk on a mid-way failure:

- the rootfs-must-be-last-partition and growth-size preconditions ran
  *after* the partition table had already been mutated;
- `delete_on_error` was accepted but never forwarded to the function's
  own `open_disk_image()` call;
- the failure path relied on `TorizonCoreBuilderError` being a
  `RuntimeError` subclass to trigger `open_disk_image()`'s own rollback -
  it isn't, so that rollback never ran.

Both preconditions now run before any partition-table mutation, and the
whole operation is wrapped so a failure removes the fresh output
(`delete_on_error=True`) or restores the original file size
(`delete_on_error=False`, the in-place case), catching
`TorizonCoreBuilderError` directly. A cleanup failure is logged and the
original error is still raised, not omitted.

## Testing

Verified with `pylint` (clean, 10/10), the upstream `bats` wic test
group (`torizoncore-builder.bats` + `wic/*.bats`, 99 tests / 0 failures),
and a hardware run on a real LEC-MTK-i1200 board covering unpack,
deploy, combine, provisioning, and the `grow_last_partition()`
mid-failure cleanup path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved raw image resizing reliability and cleanup after failures.
  * Added safeguards against unsafe partition-table changes.
  * Improved detection of U-Boot and GRUB2 bootloaders, including EFI configurations.
  * Added handling for images without a recognized bootloader.
  * Improved support for 4 KiB-sector images while preserving GPT metadata.

* **New Features**
  * Added centralized management of `uEnv.txt` settings.
  * Improved device-tree and kernel customization using configured change locations.
  * Preserved multiline settings when updating `uEnv.txt` values.

### DIFF
--- a/tcbuilder/backend/combine.py
+++ b/tcbuilder/backend/combine.py
@@ -16,6 +16,7 @@ from tcbuilder.backend.common import \
     (set_output_ownership, check_licence_acceptance, run_with_loading_animation, open_disk_image,
      get_tar_compress_program_options, DOCKER_BUNDLE_TARNAME, TAR_EXT_TO_PROGRAM,
      OSTREE_SOTA_DIR_PATH, DEFAULT_RAW_SECTOR_SIZE)
+from tcbuilder.backend.deploy import grow_last_partition
 from tcbuilder.errors import InvalidStateError, InvalidDataError, TorizonCoreBuilderError
 
 log = logging.getLogger("torizon." + __name__)
@@ -382,10 +383,6 @@ def combine_raw_image(image_path, bundle_dir, output_path, rootfs_label, force,
                 os.remove(tmp_image)
     else:
         # virt-resize can't open 4Kn disks; grow directly via libguestfs instead.
-        # Local import: a module-level one re-enters kernel.py's import cycle
-        # before cli/build.py's own import order finishes resolving it.
-        # pylint: disable-next=import-outside-toplevel
-        from tcbuilder.backend.deploy import grow_last_partition
         grow_last_partition(output_path, extra_disk_size_kb, sector_size, root_partition)
 
     with open_disk_image(output_path, delete_on_error=delete_on_error,
--- a/tcbuilder/backend/combine.py
+++ b/tcbuilder/backend/combine.py
@@ -383,7 +383,8 @@ def combine_raw_image(image_path, bundle_dir, output_path, rootfs_label, force,
                 os.remove(tmp_image)
     else:
         # virt-resize can't open 4Kn disks; grow directly via libguestfs instead.
-        grow_last_partition(output_path, extra_disk_size_kb, sector_size, root_partition)
+        grow_last_partition(output_path, extra_disk_size_kb, sector_size, root_partition,
+                            delete_on_error=delete_on_error)
 
     with open_disk_image(output_path, delete_on_error=delete_on_error,
                          sector_size=sector_size) as gfs:
--- a/tcbuilder/backend/deploy.py
+++ b/tcbuilder/backend/deploy.py
@@ -6,7 +6,9 @@ import json
 import logging
 import os
 import shutil
+import struct
 import subprocess
+import tempfile
 import threading
 import shlex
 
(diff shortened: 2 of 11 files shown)


dmitrys-emcraft:
Addressed both review comments:

- `partnum` is now derived from the actual last GPT partition
  (`gfs.list_partitions()[-1]` -> `part_to_partnum()`) instead of a
  partition count, so an image with a numbering gap no longer
  misidentifies the last partition.
- The in-place case (`delete_on_error=False`) now mutates a temporary
  sibling copy and swaps it into place only once the whole grow
  succeeds, so a failure partway through (after the partition table has
  already been rewritten) leaves the original image completely
  untouched, not merely restored to its original length.

Also added a related fix found while reproducing the first item:
after `part_del()` + `part_add()`, the regrown partition isn't
guaranteed to come back with the same number - `part_add()` picks the
lowest free slot, which can differ if another gap exists lower in the
table. The partition number is now re-derived from its start sector
before the GPT identity (name/type/GUID/attributes) is restored and the
filesystem is resized.

Verified with `pylint` (clean), the upstream `bats` wic test group, and
a hardware run on a real LEC-MTK-i1200 board.

tcbuilder/backend/kernel.py
```diff
@@ -10,9 +10,12 @@
 import urllib.request
 
 from tcbuilder.backend import ostree, dt
+# get_kernel_changes_dir is unused here; imported to re-export as kernel.get_kernel_changes_dir.
+# pylint: disable-next=unused-import
 from tcbuilder.backend.common import \
     (download_progress, get_tar_compress_program_options, get_storage_dir,
-     set_output_ownership, OSTREE_ROOT_DEPLOY_PATH)
+     set_output_ownership, OSTREE_ROOT_DEPLOY_PATH,
+     get_kernel_changes_dir)
```
lucas-akira: I don't like moving `get_kernel_changes_dir()` to `backend/common.py` and then re-exporting it here because I think it makes more sense to keep it in `kernel.py`.

I'd try to remove the circular import issue by removing the `kernel.py` dependency to `dt`: `backend/kernel.py` only uses `get_current_uenv_txt_path()` from `dt`, and that even has a TODO comment on top of it to consider moving it to a separate module other than `common.py`.

I think it's worth following the TODO and create a new module for uEnv.txt related functions currently in `dt` like `uenv.py`  or something similar.

lucas-akira (changes requested):
I agree on moving `get_image_bootloader()` to `common.py`, but I think `get_kernel_changes_dir()` should stay in `backend/kernel.py`.

tcbuilder/backend/deploy.py
```diff
@@ -325,46 +293,90 @@ def deploy_tezi_image(tezi_dir, src_sysroot_dir, src_ostree_archive_dir,
...
+            gfs.part_set_gpt_guid(dev, new_partnum, gpt_guid)
+            gfs.part_set_gpt_attributes(dev, new_partnum, gpt_attributes)
+
+            # Growing only the partition would leave the fs at its old size; the
+            # original rootfs_partition may no longer exist if renumbered above.
+            gfs.resize2fs(new_rootfs_partition)
+
+        if tmp_img:
+            os.replace(tmp_img, raw_img)
+    except (TorizonCoreBuilderError, OSError, subprocess.CalledProcessError):
```
rborn-tx: Wouldn't it be better to catch `BaseException` here?

See https://docs.python.org/3/library/exceptions.html#exception-hierarchy).

`BaseException` would cover `KeyboardInterrupt` and `RuntimeError` exceptions which are not covered by any of the three classes currently listed.

tcbuilder/backend/deploy.py
```diff
@@ -325,46 +293,90 @@ def deploy_tezi_image(tezi_dir, src_sysroot_dir, src_ostree_archive_dir,
         copy_signed_artifacts(commit_dir, output_dir)
 
 
-def grow_last_partition(raw_img, added_size_kb, sector_size, rootfs_partition):
+# pylint: disable-next=too-many-locals,too-many-positional-arguments
```
rborn-tx: It seems we can drop `too-many-positional-arguments` from here since our setup allows up to 5 positional arguments.

tcbuilder/backend/combine.py
```diff
@@ -382,11 +383,8 @@ def combine_raw_image(image_path, bundle_dir, output_path, rootfs_label, force,
                 os.remove(tmp_image)
     else:
         # virt-resize can't open 4Kn disks; grow directly via libguestfs instead.
-        # Local import: a module-level one re-enters kernel.py's import cycle
-        # before cli/build.py's own import order finishes resolving it.
-        # pylint: disable-next=import-outside-toplevel
-        from tcbuilder.backend.deploy import grow_last_partition
-        grow_last_partition(output_path, extra_disk_size_kb, sector_size, root_partition)
+        grow_last_partition(output_path, extra_disk_size_kb, sector_size, root_partition,
+                           delete_on_error=delete_on_error)
```
rborn-tx: The alignment with the above line here looks wrong.

dmitrys-emcraft:
@coderabbitai review

dmitrys-emcraft:
@lucas-akira @rborn-tx

- Returned `get_kernel_changes_dir()` to `kernel.py` and moved uEnv.txt functions from `dt.py`  to `uenv.py` as requested. The import cycle should not exist now.
- The cleanup except clause now catches `BaseException`.
- Dropped the redundant pylint disable and fixed the alignment.

lucas-akira (approved):
Latest version looks good to me.

dmitrys-emcraft:
@jsrc27 @rborn-tx — following up now that the earlier review comments have been addressed and the PR has an approval. Please let us know if there's anything further needed from your side, or if you're able to review when convenient.

jsrc27 (approved):

rborn-tx:
@dmitrys-emcraft Could you please do a rebase before another review round? TIA!

dmitrys-emcraft:
> @dmitrys-emcraft Could you please do a rebase before another review round? TIA!

Rebased the branch

rborn-tx (approved):
Did a final pass -- everything looks good. Thanks again for the contribution!

PR #166, opened 2026-08-18
bundle: wait for DinD API readiness before creating the docker client

dmitrys-emcraft:
`DindManager.get_client()` calls `docker.DockerClient()` immediately after `_wait_certs()` returned, with no check that the daemon's TCP listener was actually bound yet - certificate presence on disk does not imply that. A fresh DinD instance could still refuse the first connection attempt, producing an intermittent `Error trying to bundle Docker containers` failure with no useful detail (the caller's bare `except:` discards the real exception).

Add a bounded wait loop for the raw TCP port to accept a connection before calling `docker.DockerClient`, similar to the wait loop already implemented in `_wait_certs()`. It only checks whether anything is listening, so it should not misclassify a TLS failure (e.g. a malformed certificate) as the daemon still starting; that keeps surfacing on the unmodified `DockerClient()` call, on the first attempt, exactly as before.

This looks like a similar issue reported on the Toradex community forum in ["TorizonCore Builder on Azure VM results in ConnectionRefusedError"](https://community.toradex.com/t/torizoncore-builder-on-azure-vm-results-in-connectionrefusederror/25468) - same `ConnectionRefusedError` connecting to the daemon's `/version` endpoint during `bundle`, on unrelated infrastructure.

## Testing

Used a small script that drives `DindManager` directly (start a DinD instance, call `get_client()`) instead of going through the full `torizoncore-builder build` path, to get a real `ConnectionRefusedError` traceback out of the race instead of the generic wrapped error. 

Ran the upstream `bats` wic test group (0 failures) and a full `torizoncore-builder build` against a config with a `bundle:` block after the fix, which went through cleanly on the first try.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Docker connection reliability by validating the configured host address and waiting for the Docker API to become available.
  * Added retries during startup to ensure generated certificates and daemon readiness are complete before connecting.

### DIFF
--- a/tcbuilder/backend/bundle.py
+++ b/tcbuilder/backend/bundle.py
@@ -10,6 +10,7 @@ import socket
 import subprocess
 import sys
 import time
+import urllib.parse
 
 from contextlib import closing
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


dmitrys-emcraft:
Switched to `parsed.hostname`/`parsed.port` as suggested, requiring the `tcp` scheme and a defined port, and catching `TypeError`/`ValueError`.

`docker_host` is already built without IPv6 bracketing in `DindManager.start()` (`self.docker_host = f"tcp://{dind_ip}:{port}"` when `network_name != "host"`), so `_wait_api_ready()`'s `netloc.rsplit(":", 1)` is just parsing the address `start()` already produced, not introducing a new limitation. The `fetch-dind-network` bridge only ever generates IPv4 addresses, and IPv6 doesn't seem to be used anywhere in torizoncore-builder - left this part as is.

jsrc27 (approved):
Looks like a fine improvement to me. Out of curiosity were you, or your customer running into this issue where dind wasn't quite "ready"?

dmitrys-emcraft:
> Looks like a fine improvement to me. Out of curiosity were you, or your customer running into this issue where dind wasn't quite "ready"?

Yes - encountered it during our own QA/CI runs, not from a customer report, when we were running `torizoncore-builder build`  multiple times consecutively with a `bundle:` section in the config.

rborn-tx (approved):
Nice addition!

lucas-akira (approved):
Looks good to me.

PR #168, opened 2026-08-27
torizoncore-builder: Take the libfdt bindings from Debian

rborn-tx:
This is expected to solve the build failure happening in CI lately.

pylibfdt has no wheels, so pip built it from source and the image needed GCC and the Python development headers just for that. Debian trixie packages the same 1.7.2 bindings as python3-libfdt, so we use it now.

Since pylibfdt was the only Python requirement that needed to be built from source, the use of the Debian package also allows us to avoid temporarily installing build-essential and python3-dev in the image; this is also done here.

Related-to: TCB-914

## Summary by CodeRabbit

* **Chores**
  * Updated the Debian runtime environment to include the required device-tree tooling.
  * Simplified dependency installation by removing unnecessary compiler and Python development packages.
  * Removed a redundant dependency declaration from the Debian requirements.

### DIFF
--- a/requirements_debian.txt
+++ b/requirements_debian.txt
@@ -10,6 +10,3 @@ python-dateutil==2.9.0.post0
 pyyaml==6.0.3
 debugpy==1.8.19
 fabric==3.2.2
-
-# pylibfdt requires GCC and python3-dev.
-pylibfdt==1.7.2.post1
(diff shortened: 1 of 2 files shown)


jsrc27 (approved):
LGTM

lucas-akira (approved):

PR #170, opened 2026-09-04
backend/kernel: Before updating System.map, check if it is a symlink

lucas-akira:
In the `kernel build_module` command, the updated `System.map-{kernel_version}` is always copied to `System.map`.

Building any kernel modules for Wrynose Torizon OS images for the Nvidia Jetson Orin Nano (using JetPack 7) fails because `System.map` already exists as a symlink to `System.map-{kernel_version}`, so `shutil.copyfile()` throws a `SameFileError` exception.

Add a check that only copies `System.map-{kernel_version}`  to `System.map` if both don't refer to the same file or if `System.map` doesn't exist.

Related-to: TCB-910

## Bug Fixes

* Improved Linux kernel module installation handling for existing `System.map` links and files.
* Preserved valid symbolic or hard links that reference the matching versioned map.
* Safely handled conflicting entries and avoided overwriting the map when the destination already exists.


jsrc27 (approved):

tcbuilder/backend/kernel.py
```diff
@@ -159,9 +159,11 @@ def _prep_linux_src_for_modules_install(src_ostree_archive_dir, linux_src):
...
     repo = ostree.open_ostree(src_ostree_archive_dir)
     kernel_version = ostree.get_kernel_version(repo, ostree.OSTREE_BASE_REF)
-    shutil.copyfile(
-        os.path.join(linux_src, f"System.map-{kernel_version}"),
-        os.path.join(linux_src, "System.map"))
+    system_map_src = os.path.join(linux_src, f"System.map-{kernel_version}")
+    system_map_dst = os.path.join(linux_src, "System.map")
+    # Only replace if System.map isn't a symlink to System.map-{kernel_version}
+    if not os.path.isfile(system_map_dst) or not os.path.samefile(system_map_src, system_map_dst):
+        shutil.copyfile(system_map_src, system_map_dst)
```
rborn-tx, line 170: Have you tested the case where the symlink exists but pointing to a different `System-<version>.map` file? I wouldn't expect this to work correctly because checking the docs I see that `shutil.copyfile` copies contents of a file and it follows destination symlinks (following source symlinks is optional).

So, if you have a link `a -> b` and you write to `a` it ends up overwriting `b` which wouldn't be correct.

rborn-tx (changes requested):
Found likely problem.

tcbuilder/backend/kernel.py
lucas-akira: I haven't considered this case, good point. With the help of Claude AI I've updated the checks to replace `System.map` if it's a symlink/hardlink to a different file or a dangling symlink, so the copy procedure now won't follow links and write to unintended paths.

rborn-tx (approved):

PR #171, opened 2026-09-17
secboot: Re-sign the bootloader of Verdin AM62 images (TCB-898)

rborn-tx:
TorizonCore Builder can re-sign the bootloader of a `verdin-imx8mp` or `verdin-imx8mm` image with the customer's own keys, but not of a `verdin-am62`, whose boot chain is TI's K3 scheme rather than NXP's HAB. Toradex publishes AM62 pre-signed images built with engineering keys and tells customers to re-sign before production, so until now that instruction had no tool behind it and the only route to a production image was building Torizon OS in Yocto.

This adds the host side. It consumes `tcb_signing_files.tar.gz`, the export TOR-4543 added to signing-enabled AM62 images, and reproduces what the Yocto build does: assemble and sign the three `tiboot3` variants, `tispl.bin` and `u-boot.img`, and embed the public key the bootloader verifies the kernel with, so the chain of trust from ROM to kernel is unbroken in one operation.

### Important note

CI tests for the `verdin-am62` have not been integrated because `tdx-signed` builds are currently failing for that machine. The idea is to add those tests in a future PR that will introduce the re-signing for the `verdin-am62p`. Hopefully the build failures will be solved by then.

### Summary

- New `secboot sign-bootloader-k3` command: `--k3-key`, `--k3-degenerate-key`, `--kernel-key`, `--kernel-key-dir`, `--target-device`. Named for the signing scheme, not the chain-of-trust level.
- New backend runs binman over the two roots the image exports, R5 and A53, with the customer key staged where the binman descriptor resolves it from.
- Everything machine-specific is read from the signing files and `image.json`: device tree, firmware blobs, roots, output names. The machine table only says which machines are supported.
- `--kernel-key` is optional: without it the public key already in the image is recovered from the bootloader FIT and carried over, which is a convenience for the development loop.
- `--target-device` retargets an image between HS-FS and HS-SE, so a customer who fuses their module is not left with an image that boots only on unfused parts.
- All three `tiboot3` variants are signed and shipped, not only the installed ones, so a retargeted image can never point at a container somebody else signed.
- `build` gains the same feature under `customization.secboot`, with the schema section, the generated template and shell completion.
- `images unpack` stops warning that secboot is unsupported on K3 machines and tracks their signing files with the commit; `sign-kernel` now accepts `verdin-am62`.
- The container carries the binman patch that pins certificate serials and validity, without which re-signing produces different bytes every run and cannot be compared against the original.
- TorizonCore Builder ships TI's published degenerate key, so the user is not asked for a file that protects nothing and is the same for everyone.
- Thirteen integration cases in `secboot.bats` and one in `build.bats` covering both signing paths.  No new key material is committed.

Related-to: TCB-898

<details>

### Main changes

**`tcbuilder/backend/secboot_k3.py`** (new) is the whole signing path. The image exports two directories, `k3r5` and `a53`, each a binman working tree; the backend stages the customer key into each, runs binman over it, and collects what comes out.

binman runs with the signing directory as both its working directory and its output directory, the way the U-Boot build runs it in the build directory. Both halves are load-bearing: the descriptor reaches its inputs through relative paths, and it stages the signing keys into the output directory, which is where the openssl calls that sign the containers then look for them.

Nothing machine-specific is tabulated. The device tree to build for, the firmware blobs to embed, the roots to build and the names of the binaries to produce all come out of the signing files and `image.json`. `K3_SIGNING_SUPPORTED_MACHINES` only answers whether a machine is supported.

The `a53` descriptor ships without a `/signature` node, so the kernel public key is embedded into a copy of it, remade on every run — signing twice cannot stack one key on top of another. The GP binaries are signed with TI's degenerate key, as the OS build does.

**Carrying the kernel key over.** With no `--kernel-key`, the key already in the image is recovered from the bootloader FIT: resolve `/configurations` to the device tree it names, skip the certificate the payload is prefixed with by reading its DER length and checking for the device tree magic where that lands, and read the RSA numbers out of `/signature`. The key is wrapped in a certificate and embedded with U-Boot's own `fdt_add_pubkey`, which calls the same code `mkimage` calls, so the node is written by U-Boot rather than assembled here. The certificate is signed by a key generated and discarded on the spot, which is not a trust shortcut: the tool reads the certificate only to take its `SubjectPublicKeyInfo` and never verifies the signature. Nothing is guessed anywhere in that path, and every step fails visibly, so an image this cannot read refuses and names `--kernel-key` instead of signing something wrong. The commit is self-contained, so reverting it leaves the command working with `--kernel-key` required.

**Retargeting.** A K3 image installs the boot container matching the kind of module it is for. The request travels with the signed artifacts as a file that `deploy` consumes, and the edit lands in `deploy` rather than in the signing command because `image.json` in the output has just been updated with the size of the root filesystem packed a few lines above; overwriting the file from the artifact overlay would restore the size the image had before deploying, which is wrong in exactly the images this feature exists for. The rewrite is derived from the filenames `image.json` already carries, so no list of product ids is needed and none can go stale. Signing refuses a request whose container it did not produce, and one on an image where no installed binary carries a variant in its name at all — the second used to be a silent no-op from end to end.

**`images unpack`** decided whether an image could be signed by looking the machine up in a table of the single container name each HAB machine installs. A K3 image has four bootloader binaries and is not in that table, so every unpack ended with a warning and its signing files went untracked. The check now accepts either scheme; the container-tracking half stays HAB-only, there being no single container to track.

**The container.** binman's openssl bintool creates every certificate without a serial number or a validity period, so OpenSSL invents a random serial and a 30-day window on each run, and a re-signed artifact could never be compared against the original. The patch that pins both — the serial derived from the data being signed, the validity from `SOURCE_DATE_EPOCH` — is carried here and kept byte-identical to the copy in `meta-toradex-security`, where it is maintained until it lands upstream. `patches/u-boot/README.md` records where it came from and when it can be dropped. Three runtime dependencies are declared while there: `openssl`, which binman shells out to for every certificate and which was reaching the image only through libguestfs and a JRE, plus `yamllint` and `pyelftools`, which binman's `ti-board-config` entry type and ELF symbol writing need.

**TI's degenerate key** is shipped in `tcbuilder/secure_boot_files/`. It signs `tiboot3-am62x-gp` and the `tifsstub-gp` subimage inside `tispl.bin`, and it is a published BSD-3-Clause constant from TI's `core-secdev-k3` rather than key material in the sense that matters. `--k3-degenerate-key` exists for a build that replaced it, which nothing in the image lets us detect.

### Verification

Container built from this branch, signing the 2026-09-10 reference image with the keys that image was built with:

| Check | Result |
|---|---|
| Five artifacts against the build's own | all five `cmp`-identical |
| After `union` and `deploy tezi` | all five present and identical; `image.json` differs only in the `uncompressed_size` deploy computes |
| With no `--kernel-key` | recovers the key and reproduces all five byte-identically |
| `--target-device hs-se` | artifacts identical, all eight HS entries name the `hs` container, `gp` untouched, all three variants present |

The repacked `tcb_signing_files.tar.gz` is byte-identical to the one the image shipped, gzip stream included, so the comparison reaches the whole Tezi image rather than the five binaries in it. That holds only for an image built with pinned certificates and only when the reference key is available.

**A re-signed image has been booted**, on an unfused (HS-FS) Verdin AM62, with no issues. That is what the byte comparisons cannot show: the ROM, the R5 firmware and the bootloader each accepted a container assembled and signed here, and the bootloader verified a kernel against the public key embedded here.

### Tests

Thirteen cases in `secboot.bats` beside the HAB ones: argument errors, the refusal on an unsupported machine, signing, carrying the image's key over, retargeting, and deploying a retargeted image. Two concern reproducibility and only one ever runs, chosen from the manifest inside the image's own signing files rather than from a CI variable — with pinned certificates every installed binary must come back byte for byte, and without them the comparison is of the payload under the certificate.  What holds either way is checked unconditionally: signing twice gives the same bytes, and a generated certificate carries the fixed 2049 end date, which is what fails if the reproducibility patch ever falls out of the container. `build.bats` covers the declarative flow for both the bootloader and the kernel FIT, asking for the device kind the image does not already target so the rewrite is exercised, and comparing the binaries in the output image against the ones the run produced rather than trusting the log. No new key material is committed: the kernel key comes from the existing samples, the K3 signing key is generated when the tests start, and the degenerate key from the container.

### Limitations and what is not here

- **Retargeting to HS-SE has a documented limitation.** The Tezi recovery scripts hardcode the HS-FS filename and live in the Easy Installer recovery bundle rather than in the OS image, so TCB cannot touch them: a fused module recovered over DFU still needs a matching bundle.
- **A retargeted image has not been booted on a fused module.** The HS-FS boot above does not exercise the one thing retargeting changes: a fused module's ROM checks the container signature against the customer MPK hash in its fuses, where an unfused one does not. What is shown is that a retargeted image is internally consistent — the right container in the right `image.json` entries, signed with the right key.
- **Documentation** — the commands manual entry, the K3 section of the signing use case and the limitation above — is a separate `developer-website` pull request.
- **Running the secure boot suite per machine in CI** is a separate pull request. The suite exists but CI has never run it against an image that can be signed, so today it passes by skipping.

</details>

## Summary by CodeRabbit

- **New Features**
  - Added TI K3 bootloader signing through `secboot sign-bootloader-k3`.
  - Added support for customer and optional degenerate keys, kernel verification keys, and `hs-fs`/`hs-se` target-device selection.
  - Added K3 re-signing for compatible images, including bootloader and kernel FIT artifacts.
  - Added K3 signing support for Verdin AM62.

- **Improvements**
  - Re-signed K3 images preserve requested target-device settings.
  - Certificate generation is reproducible when a source timestamp is provided.
  - Added validation and reproducibility checks for K3 signing workflows.


tcbuilder/backend/secboot_k3.py
```diff
@@ -0,0 +1,1147 @@
...
+    carries one, which is all that is needed: only its public half is ever embedded.
+
+    :param tezi_dir: Path to the unpacked Toradex Easy Installer image
+    :param image_config: `ImageConfig` object of that image
+    :param keys_dir: Path to the directory to write the recovered key to
+    :returns: Dictionary describing the key, with the certificate written to keys_dir
+    """
+
+    key = None
+    for rawfile in _iter_image_rawfiles(image_config):
```
rborn-tx, line 603: **Kernel-key recovery takes the first `/signature` key in the first raw file that has one**

```python
    for rawfile in _iter_image_rawfiles(image_config):
        ...
        key = _read_verification_key(device_tree)
        if key:
            log.debug(f"Found a kernel verification key in {rawfile['filename']}.")
            break
```

Nothing ties the recovered key to the artifact that actually verifies the kernel. The scan takes whichever raw file first yields a device tree with a `/signature` whose first subnode carries the five properties, in `image.json` order, and within that node `_read_verification_key()` (line 489) never looks at siblings.

On today's images this is correct, and the commit message documents the measurement that shows it: `tiboot3*` are not FIT images, `tispl.bin` carries a device tree with no `/signature`, so `u-boot.img` is the only candidate and its recovered key reproduces the deployed device tree byte for byte. So this is not reachable now. It becomes reachable the day an earlier artifact grows a `/signature` node — enabling FIT signature verification in the A53 SPL is the obvious way, and the key in *that* node is the one that verifies U-Boot proper, not the kernel. The failure mode is the bad one: signing succeeds, the artifacts are byte-stable and internally consistent, and the mismatch only shows up as a kernel that will not verify on the device. Likewise a `/signature` with two key nodes silently loses the second; TCB itself never produces such an image (`_handle_secboot_sign_bootloader_k3()` rejects more than one `kernel-key`), so that half is theoretical.

Suggestion: make the recovery say what it assumes instead of taking the first hit — either scan only the artifact that carries U-Boot proper, or collect every candidate and fail when more than one raw file yields a key or when `/signature` has more than one subnode, naming `--kernel-key` as the way out, the way the "found nothing" path already does. The existing `log.debug` of the filename could be raised into the warning that already announces the carried-over key, so the user sees *which* artifact it came from and not just the key name.

tcbuilder/backend/secboot_k3.py
```diff
@@ -0,0 +1,1147 @@
...
+# the name happens to contain.
+K3_CONTAINER_EXT = ".bin"
+
+# Matches the variant infix of a boot container filename, the '-hs-fs-' in
+# tiboot3-am62x-hs-fs-verdin.bin for instance. Names without it, the GP container and the
+# binaries common to every variant, are the same whatever the image targets.
+K3_HS_VARIANT_RE = re.compile(r"-(hs(?:-fs)?)-")
+
+
+def _detect_unpacked_image_board():
```
rborn-tx, line 91: **A third copy of the board-detection block**

```python
def _detect_unpacked_image_board():
    """Read the machine name out of the U-Boot environment of the unpacked image."""
```

This is the same eight lines as in `check_unpacked_tezi_hab_signing_support()` (`tcbuilder/backend/secboot.py:476`) and `check_unpacked_tezi_kernel_signing_support()` (`tcbuilder/backend/secboot.py:641`) — read the env filename, read the file, `find_board()`, log "Detected image for machine".

This one is the nicest of the three and the natural place to converge on; hoisting it into `secboot.py` and having the two older callers use it would remove the duplication rather than add to it.

tcbuilder/cli/secboot.py
```diff
@@ -105,6 +109,62 @@ def do_sign_bootloader_hab(args):
...
+
+    if not os.path.isfile(k3_key):
+        raise InvalidArgumentError(f"Key file \"{k3_key}\" does not exist. Aborting.")
+
+    k3_degenerate_key = k3_degenerate_key or secboot_k3.K3_DEGENERATE_KEY
+    if not os.path.isfile(k3_degenerate_key):
+        raise InvalidArgumentError(
+            f"Key file \"{k3_degenerate_key}\" does not exist. Aborting.")
+
+    if kernel_key_dir and not kernel_key:
```
rborn-tx, line 129: **The `build` path reports a missing `kernel-key` with CLI switch names**

```python
    if kernel_key_dir and not kernel_key:
        raise InvalidArgumentError(
            "Error: --kernel-key-dir was passed but --kernel-key was not provided. Aborting.")
```

Copied verbatim from `sign_bootloader_hab()` (line 50), so the new command behaves exactly like the one beside it — which is a defensible reason to leave it alone. Raising it only because this PR doubles the reach of the wording: `_handle_secboot_sign_bootloader_k3()` (`tcbuilder/cli/build.py:432`) enters this same function from `tcbuild.yaml`, so a user who wrote `kernel-key-dir:` and forgot `kernel-key:` in the YAML is told about two command-line switches they never typed.

The tree does distinguish the two vocabularies elsewhere: `_sign_kernel_parse_ostree_key()` (`tcbuilder/cli/build.py:391`) words the identical situation as "Property 'sign-kernel.ostree-key-dir' was set, but 'sign-kernel.ostree-key' was not provided".

If it is worth fixing at all, the fix belongs to both commands at once — validate the pair in the `build` handlers with property wording — not to this one alone.

tests/integration/functions.sh
```diff
@@ -273,6 +273,48 @@ requires-supported-hab-signing-machine() {
...
+    torizoncore-builder-shell "
+        tar -xOzf /storage/tezi/tcb_signing_files.tar.gz tcb-signing/manifest.json 2>/dev/null \
+            | jq -e '.signing.certificates_pinned == true' >/dev/null"
+}
+export -f unpacked-image-has-pinned-certificates
+
+# The key that signed the image under test, when whoever runs the tests holds it. Only then
+# can a re-signed binary be compared with the original byte for byte: same sources and same
+# binman give the same bytes, but only when signed by the same key.
+requires-k3-reference-key() {
```
rborn-tx, line 308: **`TCB_K3_REFERENCE_KEY` is undocumented, and this helper reads a variable only one suite exports**

Two small things.

First, `TCB_K3_REFERENCE_KEY` gates the one test that compares re-signed binaries against the image's own, and it is not listed in `tests/integration/README` beside `TCB_MACHINE`, `TCB_TAGS` and the rest — including the thing a reader will get wrong, that the value is resolved inside the container (the runner mounts `$(pwd)` at `/workdir`), so a host absolute path will not work.

Second, the helper lives in the shared `functions.sh` but reads `K3_REFERENCE_KEY`, which only `secboot.bats:49` sets and exports; called from any other suite it would skip unconditionally and look like "no key configured". Reading `TCB_K3_REFERENCE_KEY` directly here would make the helper self-contained.

rborn-tx (commented):
Reviewed by Claude.

tcbuilder/backend/secboot_k3.py
```diff
@@ -0,0 +1,1147 @@
...
+            f"image is for \"{board}\": they do not belong to this image. Aborting.")
+
+    roots = manifest.get("roots")
+    if not roots:
+        raise FileContentMissing(
+            f"The signing files declare no directories to build from in \"{K3_MANIFEST_NAME}\". "
+            "Aborting.")
+
+    for root in roots:
+        if not os.path.isdir(os.path.join(signing_dir, root)):
```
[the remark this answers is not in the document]
rborn-tx: Fixed in `59cbc1dd`. `read_k3_signing_manifest()` now resolves each root and refuses anything that does not land inside the extracted signing files. Resolving before comparing means an absolute path, a `..`, and a root that is itself a symlink out are all caught by the same check.

Reproduced before and after with a hand-written manifest: roots of `/tmp` and `../escape` were both accepted and would have staged the key there; both are now refused, and the real roots still pass.

tcbuilder/backend/secboot_k3.py
```diff
@@ -0,0 +1,1147 @@
...
+    :param degenerate_key: Path to TI's degenerate key, which signs the GP artifacts
+    :param kernel_key_dir: Path to the directory holding the kernel key
+    :param kernel_key_name: Name of the kernel key, or None to carry over the key already in
+                            the image being signed
+    :param kernel_key_algo: Pair of hashing and crypto algorithms used to sign the kernel
+    :param target_device: Device state the deployed image should target, or None to leave the
+                          image targeting what it already targets
+    """
+
+    board = check_unpacked_tezi_k3_signing_support()
```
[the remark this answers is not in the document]
rborn-tx: Fixed, folded into `secboot: Add bootloader signing for TI K3 based machines` (`33567af0`). `completed` is initialized first and the machine check moved inside the `try`, as proposed.

One consequence worth recording: the work directory the `finally` now also removes on this path is shared with the HAB and kernel signing paths. Each of those wipes and recreates it when it starts, and nothing reads it between commands, so widening the cleanup costs nothing.

tests/integration/functions.sh
```diff
@@ -273,6 +273,48 @@ requires-supported-hab-signing-machine() {
...
+
+requires-pinned-certificates() {
+    if ! unpacked-image-has-pinned-certificates; then
+        skip "image was not built with reproducible certificates"
+    fi
+}
+export -f requires-pinned-certificates
+
+requires-unpinned-certificates() {
+    if unpacked-image-has-pinned-certificates; then
```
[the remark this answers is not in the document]
rborn-tx: Fixed, folded into `tests: Cover K3 bootloader signing` (`ad7961c9`).

The helper now returns the value rather than answering a yes-or-no question, so the caller can tell the three cases apart: `requires-unpinned-certificates` runs only on an explicit `false` and skips when there is no readable manifest to go on. `requires-pinned-certificates` keeps the behaviour it had, since "not explicitly true" is the right answer there.

tcbuilder/backend/secboot_k3.py
```diff
@@ -0,0 +1,1147 @@
...
+    :param root_dir: Path to the signing root
+    :param k3_key: Path to the customer key every boot container is signed with
+    :param degenerate_key: Path to TI's degenerate key, which signs the GP artifacts
+    """
+
+    keys_dir = os.path.join(root_dir, K3_KEYS_SUBDIR)
+    os.makedirs(keys_dir, exist_ok=True)
+
+    shutil.copy2(k3_key, os.path.join(keys_dir, K3_CUSTMPK_NAME))
+    shutil.copy2(degenerate_key, os.path.join(keys_dir, K3_DEGENERATE_KEY_NAME))
```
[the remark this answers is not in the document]
rborn-tx: Diagnosis confirmed. Both halves reproduce on the real signing tarball, extracted: a keys directory that already exists as a symlink out of the work directory receives the key through `shutil.copy2()`, and a link planted among the original entries is handed to the repack, which dereferences it.

We are declining it as an accepted risk. It needs the user to sign an image they did not build, which is not the workflow the feature targets — the images are Toradex's own or the customer's Yocto output. A fix was written and measured, then dropped as costing more in the reading of the module than the exposure justifies.

One note for whoever revisits this, since the obvious fix is the wrong one: it must not be to drop `--dereference` and `--hard-dereference`. Those are the options used by the recipe that produces the tarball (`tcb-signing-files.bb` in `meta-toradex-torizon`), and matching them is what makes the repacked tarball byte-identical to the one the image shipped, which is how a re-signed image is compared with its original. Refusing links is the route that keeps both — and since the recipe dereferences as well, a well-formed tarball has no links to refuse: the reference one has 69 members and not one of them is a link.

tcbuilder/backend/secboot_k3.py
rborn-tx: Declining this one as an accepted risk, having written and measured the fix first.

The reachability reads as stated: `u-boot.img` is the only candidate today, and it stops being so only if an earlier stage grows a `/signature` — a change on the OS side, made by someone already working in this area. The carried-over key is a development convenience in the first place; anyone going to production passes `--kernel-key` and never enters this path.

Recorded rather than forgotten, because the failure mode is the silent one: signing succeeds, the artifacts are byte-stable and internally consistent, and the mismatch only shows up as a kernel the device will not verify. If the boot chain ever grows a second signature node, this is the first thing to revisit.

tcbuilder/cli/secboot.py
rborn-tx: Deferred, on this comment's own reasoning: fixing it for the K3 command alone would leave `sign-bootloader-hab` next to it wording the identical situation differently, for no reason a reader could see. It belongs in the `build` handlers with property wording, covering both commands, as a change of its own.

tests/integration/functions.sh
rborn-tx: Both halves fixed in `68c55cbe`.

`requires-k3-reference-key` now reads `TCB_K3_REFERENCE_KEY` directly, as `functions.sh` already does for `TCB_UNDER_CI`, and the local alias in `secboot.bats` is gone so the key has one name rather than two. The variable is documented in `tests/integration/README` beside the other `TCB_` ones, including the part that is easy to get wrong: the path is resolved inside the container, so it has to be given relative to the test working directory rather than as a host path.

tcbuilder/backend/secboot_k3.py
rborn-tx: Declining it for this pull request, with a `TODO` left in the code so the next person to touch the function meets it.

The end state you describe is the right one — one helper in `secboot.py` with all three callers on it. What keeps it out of here is that the two older copies are inlined in `check_unpacked_tezi_hab_signing_support()` and `check_unpacked_tezi_kernel_signing_support()`, so converging means editing two functions the K3 signing work does not otherwise touch. That belongs in a change of its own rather than widening this diff.

The `TODO` names both call sites and says why it was left, so the note is actionable rather than a marker.

jsrc27:
I didn't notice any major issues. I have some questions though just to help my understanding here.

* I don't think we have pre-signed images publicly published yet for the AM62, correct? I assume this will be available "soon"
* Regarding the degenerate key. Is this something TI never/very rarely changes to your knowledge?
  * Related to this, the option for a user to supply their own degenerate key. They would have to tweak this all the way in the Yocto recipe that provides TI's key I assume? Is there a reason a user would ever want to change this?

rborn-tx:
@jsrc27 Thanks for the review.

Answering your questions:

> I didn't notice any major issues. I have some questions though just to help my understanding here.
> 
>     * I don't think we have pre-signed images publicly published yet for the AM62, correct? I assume this will be available "soon"

That's right. We'll need to publish them for the re-signing to become truly useful to most users; the same is true for the `verdin-imx8mm` whose re-signing capability is already present in TCB but no images are publicly available yet.

>     * Regarding the degenerate key. Is this something TI never/very rarely changes to your knowledge?

The degenerate key is needed by the tooling to produce the bootloader container but the corresponding signature is not checked by the ROM code, only the container format is. Due to that, any degenerate key would work fine for signing. Therefore, in theory TI would have no reasons to change it; checking https://github.com/TexasInstruments/core-secdev-k3/blob/ed6951fd3877c6cac7f1237311f7278ac21634f3/keys/ti-degenerate-key.pem I see a single commit made three years ago.

>       * Related to this, the option for a user to supply their own degenerate key. They would have to tweak this all the way in the Yocto recipe that provides TI's key I assume? Is there a reason a user would ever want to change this?

In Yocto we require users to create and pass a degenerate key; that's probably not great but that's what we have today. In TCB, I decided to allow the user not to pass it - in which case the tool would use an internal fixed key taken from TI. But they can pass one if they want and the only reason I imagine they would pass one is to reproduce an Yocto build.

In fact, part of the work on the OS side prior to this work on TCB (https://github.com/toradex/meta-toradex-security/pull/191) was exactly to make builds reproducible on the Yocto side. Considering that use, I'd say that we are probably the main users of this option. The reproducibility allows us to tell with certainty that the signing is working correctly (correctly meaning the same as in Yocto) by just comparing TCB's output with Yocto's output byte-for-byte. Without it, we'd need to run the image in the hardware to get that confirmation every time we changed something related to signing, which is of course a complication in terms of maintenance.

Btw, one test that I run is worth mentioning: I produced two images in Yocto with different key sets: A and B; then I took image A and re-signed it with the key set B producing bootloader binaries matching byte-for-byte those of image B.

P.S.: You question here made me realize the help text for the degenerate-key switch wasn't really accurate; fixed in the new version.

jsrc27 (approved):
No further comments.

lucas-akira (approved):
I'm not very familiar on how secure boot is done for K3 devices at the moment, so I had to do some research to get more context on the implementation details and that took a bit longer than expected.

With that being said, overall the code looks good on my side.

